PR #3, opened 2026-01-18
AI: Refactor `ast/base.rs`. Extract shared metadata (span, type) into a packed struct. Implement the `AstEntity` marker trait for the new unified ThinNode system.

mohsen1:
{"id":"3_base_refactor","description":"Refactor `ast/base.rs`. Extract shared metadata (span, type) into a packed struct. Implement the `AstEntity` marker trait for the new unified ThinNode system.","files":["src/parser/ast/base.rs"],"dependencies":["2_core_thin_node_def"]}


mohsen1:
AgentName: M5Manager

PR #14, opened 2026-01-18
AI: Correct logic for 'use strict' directive detection which is causing false negatives.

mohsen1:
{"id":"fix_strict_directive_parsing","description":"Correct logic for 'use strict' directive detection which is causing false negatives.","files":["test-strict-directive.mjs"]}


mohsen1:
AgentName: M5Manager

PR #35, opened 2026-01-18
AI: Refactor subtyping checks in `subtype.rs` to an iterative algorithm using the shared `Worklist` infrastructure to prevent stack overflows.

mohsen1:
{"id":"T9","description":"Refactor subtyping checks in `subtype.rs` to an iterative algorithm using the shared `Worklist` infrastructure to prevent stack overflows.","files":["src/solver/subtype.rs","src/solver/subtype_tests.rs"]}


mohsen1:

PR #513, opened 2026-04-19
perf(tsz-server): persistent node worker for native TypeScript operations

mohsen1:
## Summary

Rename/classification/etc. delegates to real `tsc` via a Node.js subprocess. Previously every call spawned a fresh `node` process that had to `require(\"typescript\")` from scratch — roughly 1–2 s of module-load overhead per call. Tests that make many native-LS calls in series (rename across node_modules, autoImportFileExcludePatterns, etc.) kept hitting the 15 s watchdog timeout.

## What changed

- Extracted the inline Node.js worker script into a sibling file (`native_ts_worker.js`) so both the legacy single-shot path and the new persistent path share the same source of truth.
- Added a `TSZ_NATIVE_TS_PERSISTENT=1` mode to the script: newline-delimited JSON requests on stdin, newline-delimited JSON responses on stdout, TypeScript runtime loaded once and reused for every request.
- `Server` grows a `native_ts_worker: Option<Mutex<NativeTsWorker>>` field, populated at startup. `try_native_typescript_operation` now tries the worker first and falls back to the original per-call `node -e SCRIPT` pathway if the worker isn't healthy.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo nextest run -p tsz-cli` (966/966)
- [x] Architecture guardrail (`scripts/arch/check-checker-boundaries.sh`)
- [x] Fourslash: 6562/6562 on a fresh local run
- [x] Individual slow-path sanity:
  - `renameImport.ts` 30 s → 5.7 s
  - `renameTemplateLiteralsComputedProperties` 22 s → 2.4 s

### DIFF
--- a/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_tests.rs
+++ b/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_tests.rs
@@ -39,6 +39,7 @@ fn make_server() -> Server {
         include_completions_with_class_member_snippets: false,
         new_line_character: None,
         plugin_configs: FxHashMap::default(),
+        native_ts_worker: None,
     }
 }
 
--- a/crates/tsz-cli/src/bin/tsz_server/handlers_info_alias.rs
+++ b/crates/tsz-cli/src/bin/tsz_server/handlers_info_alias.rs
@@ -37,668 +37,7 @@ impl Server {
         &self,
         mut payload: serde_json::Value,
     ) -> Option<serde_json::Value> {
-        const SCRIPT: &str = r#"
-const fs = require("fs");
-const path = require("path");
-
-function loadTypescript() {
-  const cwd = process.cwd();
-  const candidates = [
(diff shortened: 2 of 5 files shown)


mohsen1:
Architectural direction reversed. The tsz-server should not require TypeScript at all — the right move is to remove the native-TypeScript delegation entirely, not optimize its startup cost. Opening a follow-up PR that implements the currently-delegated operations natively in Rust (rename, navtree/navbar/navto, encodedSemanticClassifications, format/formatOnKey).

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76919, additions=2104295, deletions=8872713, head=perf/native-ts-worker, mergeStateStatus=DIRTY. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #519, opened 2026-04-19
fix(checker): TS7027 for 'var x = N' after unreachable point

mohsen1:
## Summary
- `is_var_without_initializer` was iterating `VARIABLE_STATEMENT.declarations` as if it contained `VARIABLE_DECLARATION` nodes directly, but it actually contains a single `VARIABLE_DECLARATION_LIST` that wraps the individual declarations.
- As a consequence, every `var x = N` after a terminator was silently suppressed (treated as "without initializer"), while `let`/`const` still reported TS7027 because the list flag check was independent of the data access.
- Descend into the inner list before checking each declaration's initializer, matching the pattern used elsewhere (e.g. `module_augmentation.rs`, `cross_file_conflicts.rs`).

## Impact
- Fixes `jsFileCompilationBindReachabilityErrors.ts`.
- Compiler lane: +1 test (6257 → 6258), no regressions.
- `var t;` (no initializer) after a terminator remains correctly suppressed.
- `unreachableDeclarations.ts` — pre-existing fingerprint-only extras are unchanged by this fix.

## Test plan
- [x] Target test passes: `./scripts/conformance/conformance.sh run --filter "jsFileCompilationBindReachabilityErrors"`
- [x] Compiler lane: `./scripts/conformance/conformance.sh run --filter "compiler"` — 6258/6519 (+1 vs baseline 6257)
- [x] Reachability neighborhood: `--filter "controlFlow"`, `--filter "unreach"`, `--filter "Reachab"` — no new regressions
- [x] Manual parity with tsc for `var x = N`, `var t`, and `let` after a `return`

### DIFF
--- a/crates/tsz-checker/src/flow/reachability_checker.rs
+++ b/crates/tsz-checker/src/flow/reachability_checker.rs
@@ -834,17 +834,22 @@ impl<'a> CheckerState<'a> {
         let Some(var_data) = self.ctx.arena.get_variable(node) else {
             return false;
         };
-        // Check if it's `var` (not let/const) by examining declaration list flags
-        // The flags are on the VariableDeclarationList child node
-        for &decl_idx in &var_data.declarations.nodes {
-            if let Some(decl_node) = self.ctx.arena.get(decl_idx) {
-                // Check if it's let/const (not var) using combined node+parent flags
+        // VARIABLE_STATEMENT.declarations holds a VARIABLE_DECLARATION_LIST; the
+        // individual VARIABLE_DECLARATION nodes live one level deeper.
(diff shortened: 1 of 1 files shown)


mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76919, additions=2104511, deletions=8872713, head=fix/reachability-var-initializer-descent, mergeStateStatus=DIRTY. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #573, opened 2026-04-19
lsp: big navtree parity batch (initializers, objects, modules, signatures, params, sort, static blocks)

mohsen1:
## Summary

Large navtree/navbar parity batch — pure-Rust path without the Node/tsc native-TS fallback. Flips **23 tests** under \`TSZ_DISABLE_NATIVE_TS=1\` (21/68 → 44/68, from 30.9% → 64.7%). Native-TS path unchanged throughout (68/68).

### Changes

1. **Module name parity** — qualified namespace flattening (\`A.B.C\` modeled as nested MODULE_DECLARATIONs now renders as one nav item), string-literal module names keep quotes, \`cleanText\` rule (150-char cap + line-continuation stripping), \`escapeString\` on external-module filenames.

2. **Variable-initializer walks** — \`const o = { a: ..., b: { m() {} } }\` surfaces a/b/m. Classifier picks between method / nested object / class / plain property based on the initializer shape. Spread \`...x\` surfaces its identifier name. Arrow/function-expression initializers walk their body for nested declarations.

3. **Factory-function returns** — \`function F() { return { a, b } }\` surfaces a/b as F's navtree children when no inner declarations already bubbled up.

4. **String-literal names** — \`"prop": value\` renders navbar text \`"prop"\` with quotes preserved (matches tsc's \`nodeText(name)\`).

5. **Variable statements in block bodies** — \`collect_children_from_block\` now surfaces \`VARIABLE_STATEMENT\`s alongside declaration kinds (tsc treats vars as nav nodes too), which fixes \`function baz() { var v }\` dropping v.

6. **Binding patterns** — \`const { a, b } = …\` / \`const [c, d] = …\` / nested \`var {e, x: [f, g]} = …\` surface every terminal name with the enclosing declaration's kind/modifiers.

7. **Enum members with computed keys** — \`[Symbol.isRegExp]\` regressed after we started stringifying COMPUTED_PROPERTY_NAME; explicit skip in the ENUM_MEMBER arm.

8. **Name-less function recovery** — \`function;\` (parser error recovery) now renders as \`<function>\` instead of an empty-string entry.

9. **Constructor-first ordering** — \`Constructor\` sorts as "nameless" (tsc's \`tryGetName\` returns undefined), so constructors always lead their sibling properties.

### Test plan

- [x] \`cargo nextest run -p tsz-lsp\` — 3759/3759
- [x] \`TSZ_DISABLE_NATIVE_TS=1 run-fourslash.sh --filter=navigationBar\` — 44/68 (up from 21)
- [x] Targeted tests verified individually (listed per commit)
- [x] Native-TS full navbar filter: 68/68 unchanged

### DIFF
--- a/crates/tsz-cli/src/bin/tsz_server/handlers_info.rs
+++ b/crates/tsz-cli/src/bin/tsz_server/handlers_info.rs
@@ -60,6 +60,36 @@ fn symbol_kind_to_tsserver(
     }
 }
 
+/// Mirror tsc's `escapeString(s, '"')` — replace control characters
+/// and backslash/double-quote with their JS escape sequences, and
+/// encode non-printable high chars as `\uNNNN`. Used on the filename
+/// stem before wrapping it in double quotes for external-module
+/// navbar/navtree root entries, so a filename like `my fil<TAB>e`
+/// renders as `"my fil\te"` rather than embedding a literal tab.
+fn escape_string_double_quote(s: &str) -> String {
+    let mut out = String::with_capacity(s.len());
+    for c in s.chars() {
+        match c {
+            '\\' => out.push_str("\\\\"),
+            '"' => out.push_str("\\\""),
+            '\t' => out.push_str("\\t"),
+            '\n' => out.push_str("\\n"),
+            '\r' => out.push_str("\\r"),
+            '\x08' => out.push_str("\\b"),
+            '\x0c' => out.push_str("\\f"),
+            '\x0b' => out.push_str("\\v"),
+            '\u{2028}' => out.push_str("\\u2028"),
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76937, additions=2107458, deletions=8872713, head=lsp/navtree-batch-4, mergeStateStatus=DIRTY. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #622, opened 2026-04-19
lsp: uniform export= / export default expression classification

mohsen1:
## Summary

Consolidate the \`export = <expr>\` and \`export default <expr>\` classification into a single helper so both paths recognize the same expression shapes:

- **FunctionExpression / ArrowFunction** → \`function\` with body-walked children
- **ClassExpression** → \`class\` (members inlined, not wrapped)
- **ObjectLiteralExpression** → \`const\` with literal members
- **CallExpression** → \`const\` with object-literal argument members
- **ParenthesizedExpression** → unwrap and reclassify
- **AsExpression / SatisfiesExpression / TypeAssertion** → unwrap and reclassify underlying expr
- **else** → \`var\`

The \`export=\` path always emits the \`export\` kindModifier.

## Context

Doesn't flip \`navigationItemsExportEqualsExpression\` / \`navigationItemsExportDefaultExpression\` on their own — those tests assert tsc's \`Intl.Collator\`-based UI sort order which Rust's \`to_lowercase().cmp()\` doesn't replicate. This is a per-entry correctness improvement over emitting \`var\` for every case.

## Test plan

- [x] \`cargo nextest run -p tsz-lsp\` — 3759/3759
- [x] Full Rust-only fourslash: 6522/6562 (no regressions)
- [x] Full native-TS fourslash: 6554/6562 (unchanged, timeouts only)

### DIFF
--- a/crates/tsz-lsp/src/symbols/document_symbols.rs
+++ b/crates/tsz-lsp/src/symbols/document_symbols.rs
@@ -2236,7 +2236,10 @@ impl<'a> DocumentSymbolProvider<'a> {
         )
     }
 
-    /// Check if a node kind is a declaration.
+    /// Check if a node kind is a declaration. Used in the
+    /// EXPORT_DECLARATION arm to decide whether to recurse into the
+    /// exported clause (declarations) vs. treat it as a re-export
+    /// (NAMED_EXPORTS etc).
     const fn is_declaration(&self, kind: u16) -> bool {
         kind == syntax_kind_ext::FUNCTION_DECLARATION
             || kind == syntax_kind_ext::CLASS_DECLARATION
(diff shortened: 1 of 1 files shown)


mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76942, additions=2109725, deletions=8872711, head=lsp/nonnavbar-batch3, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #657, opened 2026-04-20
fix(checker): preserve declared annotation literal property types in …

mohsen1:
…TS2322 messages

- sanitize_type_annotation_text: reject unbalanced-brace truncated annotations
- declared_identifier_source_display: return literal display only for declared annotation types with literal canonical props (no display_properties on declared_type)
- rewrite_source_display: skip widening for source types with direct literal canonical properties (using object_shape_for_type + is_literal_type check)

tsc preserves `length: 2` from `var z: { length: 2; }` in error messages;
our widening path was converting it to `length: number`.

### DIFF
--- a/crates/tsz-checker/src/assignability/assignment_checker/arithmetic_ops.rs
+++ b/crates/tsz-checker/src/assignability/assignment_checker/arithmetic_ops.rs
@@ -72,7 +72,11 @@ impl<'a> CheckerState<'a> {
 
         let mut emitted = false;
 
-        if !left_is_valid {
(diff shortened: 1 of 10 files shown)


mohsen1:
Closing — this PR's literal-property preservation change regresses conformance (12008 < 12046 baseline) and also fails 1 checker unit test (jsx_array_children_callbacks_emit_child_level_ts2322_without_ts7006). The change overlaps with the display path reworked by the merged fixes #649-#656 and needs to be redesigned against current main rather than fix-forward. Will reopen if a narrower version lands.

PR #673, opened 2026-04-20
fix(checker): preserve namespace and display property order

andrewtdiz:
TS2322 parity was still unstable for CommonJS and namespace-object paths. Several namespace builders assigned declaration_order while iterating FxHashMap-backed export tables, so named exports could drift away from source order once there were 3+ properties. Merged display properties could also be re-sorted by per-member counters, which moved later intersection members ahead of earlier contributors.

Order namespace exports by declaration span, normalize default and named exports before interning, and preserve left-to-right first-seen order when merging display properties. This also normalizes JS export surfaces and adds focused regressions for export =, synthetic default namespace imports, and exports.* surfaces.

Keep conformance_issues as a real Cargo test target again and route the checker-side helper calls through query boundaries so the rebased branch continues to satisfy the architecture guards.

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -104,10 +104,6 @@ name = "ts1254_ambient_const_tests"
 path = "tests/ts1254_ambient_const_tests.rs"
 
 [[test]]
-name = "conformance_issues"
-path = "tests/conformance_issues.rs"
-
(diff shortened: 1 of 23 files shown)


mohsen1:
Thanks! this looks good. conformance and fourslash baselines are not being reached in CI. investigating those later. I'll merge this as is.

PR #686, opened 2026-04-20
fix(checker): emit TS2420 for interface-side private/protected member mismatch

mohsen1:
## Summary

Fingerprint-only pickup from `implementingAnInterfaceExtendingClassWithPrivates2.ts`: tsz was missing a TS2420 diagnostic at `test.ts:13:7` for `class Bar2 extends Foo implements I` where Foo has `private x` and Bar2 redeclares `x!: string` as public.

`class_implements_checker::core` only emitted TS2420 when the *class* side declared a `private`/`protected` member. The mirror case — when the *interface* has a private or protected property (typically via `interface I extends Foo` inheriting it) and the class's corresponding member is public — was silently accepted. Added the mirror-visibility branches after the class-side checks.

Two subtleties in the new branches:

- **Private**: always emit. Redeclaring a `private` member is a nominal mismatch even when the class extends the declaring base. tsc: *"Property 'x' is private in type 'I' but not in type 'Bar2'."*
- **Protected**: skip when the class inherits the same-named private/protected member through its own `extends` chain — tsc treats the redeclaration as a widening-visibility override in that case (e.g., `class Bar8 extends Foo implements I { x: string; ... }` where Foo has `protected x`). `inherited_non_public_members` is already computed upstream; the new branch consults it.

## Test plan

- [x] `./scripts/conformance/conformance.sh run --filter "implementingAnInterfaceExtendingClass"` → 3/3 PASS (was 2/3; `WithPrivates2.ts` fingerprint-only failure resolved, `WithProtecteds.ts` preserved).
- [x] `./scripts/conformance/conformance.sh run --filter "implement"` → 12/12 PASS.
- [x] `cargo nextest run -p tsz-checker` → 4835 tests run, 4835 passed.

### DIFF
--- a/crates/tsz-checker/src/classes/class_implements_checker/core.rs
+++ b/crates/tsz-checker/src/classes/class_implements_checker/core.rs
@@ -1089,6 +1089,7 @@ impl<'a> CheckerState<'a> {
                                 (sym_flags & tsz_binder::symbol_flags::PRIVATE) != 0;
                             let is_class_member_protected =
                                 (sym_flags & tsz_binder::symbol_flags::PROTECTED) != 0;
+                            let interface_visibility = prop.visibility;
                             if is_class_member_private {
                                 self.error_at_node(
                                         class_error_idx,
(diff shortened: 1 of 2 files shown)


mohsen1:
Automation triage on 2026-04-21: this PR is currently blocked by an external CircleCI checker failure.\n\nEvidence from `gh pr checks 686`:\n- failing: `ci/circleci: test-checker` at https://circleci.com/gh/mohsen1/tsz/118\n- passing: conformance, conformance-aggregate, emit, fourslash, lint, binder, build, build-server, emitter, foundation, lsp, root, solver, wasm-build, GitGuardian, and GitHub bench compare.\n\nBecause the failing job is CircleCI rather than GitHub Actions, the available `gh` workflow does not expose the job log inline here; I left this as a clear external CI blocker instead of making an unverified code change.

PR #696, opened 2026-04-21
fix(solver): store forward display alias for conditional-alias appare…

mohsen1:
…nt type

When a conditional type alias like `DeepReadonly<Part>` evaluates through a conditional branch to another Application (`DeepReadonlyObject<Part>`), tsc displays the intermediate Application name in diagnostics rather than the outermost alias. This is the apparent type concept for conditional type aliases.

Previously, `evaluate_application` only stored the reverse alias `display_alias[expanded_type] = DeepReadonly<Part>`, so the formatter always showed the outer alias name. The tail-call optimization in `evaluate_conditional` bypassed `evaluate_application` for Application branches, meaning the inner Application was never aliased.

Fix: Add `apparent_conditional_branch` signal field to `TypeEvaluator`. `evaluate_conditional` sets this when it resolves to an Application branch (via tail-call or direct evaluation). `evaluate_application` reads the signal after evaluation and stores a forward alias `display_alias[DeepReadonly<Part>] = DeepReadonlyObject<Part>` so the formatter chains through to the intermediate name.

Fixes TS2339 display: type 'DeepReadonly<Part>' -> 'DeepReadonlyObject<Part>'
Fixes TS2542 display: type 'DeepReadonly<Part[]>' -> 'DeepReadonlyArray<Part>'


mohsen1:
Superseded by PR #669 (commit `b9c56b69a8`), which already landed the apparent_conditional_branch signal for conditional-alias display on main. Rebasing this branch produces an empty diff against origin/main, so closing.

PR #698, opened 2026-04-21
[codex] fix generic indexed readonly write diagnostics

mohsen1:
## Summary

Fixes the `keyofAndIndexedAccess2.ts` wrong-code conformance mismatch where tsz emitted an extra TS2536 and missed TS2862 for generic indexed readonly writes.

## Root cause

The checker was classifying some generic writes through concrete broad `keyof` keys as ordinary indexed access failures before the readonly generic-write path could produce TS2862. A nested mapped tuple value access (`Q[K]["0"]`) also lost enough constraint value-space information to emit TS2536 even though the mapped value constraint guarantees a tuple-like value.

## Changes

- Treat mapped/generic mapped constraints as eligible for generic indexed readonly write diagnostics.
- Defer generic indexed writes in element access so TS2862 can be produced instead of TS2536.
- Allow nested indexed access validation through mapped tuple constraint value spaces.
- Add focused regression tests for both the TS2862 and mapped tuple value access cases.

## Validation

- `cargo fmt --all --check`
- `cargo check --package tsz-checker`
- `cargo nextest run --package tsz-checker --test conformance_issues test_mapped_tuple_value_index_access_does_not_emit_ts2536 test_readonly_generic_write_with_concrete_keyof_reports_ts2862_not_ts2536`
- `./scripts/conformance/conformance.sh run --filter "keyofAndIndexedAccess2" --verbose`
  - Diagnostic code sets now match: `[TS2322, TS2339, TS2739, TS2862, TS7053]`.
  - The file still has fingerprint-only message/position drift outside this fix.

## Notes

`verify-all.sh` was attempted during the automation run. Earlier formatting issues were fixed, and main has since unblocked one LOC guard, but the full verifier was not completed in the original worktree because it hit existing full-suite failures and conformance dependency setup behavior.

### DIFF
--- a/crates/tsz-checker/src/state/state_checking/readonly.rs
+++ b/crates/tsz-checker/src/state/state_checking/readonly.rs
@@ -591,7 +591,11 @@ impl<'a> CheckerState<'a> {
     /// - The index is `keyof T` — constrains to the receiver's own key space
     /// - The index is `K extends keyof T` — a type parameter constrained to keyof
     /// - The constraint has no index signature (e.g., `{ a: string, b: number }`)
-    fn is_generic_indexed_write(&mut self, object_type: TypeId, index_type: TypeId) -> bool {
+    pub(crate) fn is_generic_indexed_write(
+        &mut self,
+        object_type: TypeId,
+        index_type: TypeId,
+    ) -> bool {
         use crate::query_boundaries::common as common_query;
         use crate::query_boundaries::common::{IndexKind, IndexSignatureResolver};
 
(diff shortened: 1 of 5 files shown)


mohsen1:
Automation triage on 2026-04-21: this PR is currently blocked by merge conflicts and failing GitHub Actions checker shards.\n\nEvidence:\n- mergeStateStatus=DIRTY for branch `fix/keyof-indexed-access-readonly-write` against `main`.\n- `gh pr checks 698` reports failed checks: `Test: Checker (2/4)`, `Test: Checker (4/4)`, and `CI Summary` from run https://github.com/mohsen1/tsz/actions/runs/24709578995.\n- The changed-file set spans multiple checker type computation/reporting modules, so I did not try to resolve this broad conflicted draft branch inside a separate single-target conformance iteration.

PR #728, opened 2026-04-21
fix(checker): detect invalid CommonJS export property assignment earlier

mohsen1:
## Summary

Two related fixes so invalid CommonJS export property assignments surface a TS2339 instead of being masked:

1. In `assignment_ops`, call `error_invalid_commonjs_export_property_assignment` **before** the general assignability flow so the diagnostic fires and the expression returns the RHS type immediately, rather than after the non-const assignability branch that previously shadowed it.
2. In `commonjs_assignment`, when the JS export surface doesn't yet have a `direct_export_type` (seen when primitive `module.exports = …` is later mutated via `exports.X = …`), fall back to the access expression's actual type instead of bailing out of the check. Lets the "cannot add property on primitive" path fire consistently.

## Target

Unignores the regression test `test_primitive_module_exports_assignment_reports_same_file_property_error_with_prelude`, which was marked `#[ignore]` because "TS2318 floods from missing global types mask actual TS2339" — the ordering fix resolves that masking.

## Test plan

- [x] Previously-ignored test now passes: `cargo nextest run -p tsz-checker --test js_export_surface_tests -E 'test(test_primitive_module_exports_assignment_reports_same_file_property_error_with_prelude)'` — 1/1 pass.
- [x] `cargo nextest run -p tsz-checker --test js_export_surface_tests` — 65/65 pass.
- [x] `cargo nextest run -p tsz-checker --lib` — 2598/2598 pass.
- [x] Pre-commit full workspace test suite — 12925+ tests passing.

## Notes

Recovered from an interrupted parallel agent worktree (codex/conformance-20260421172436-09). Full conformance suite not re-run in this session; relying on pre-commit full test run for unit-test parity.

### DIFF
--- a/crates/tsz-checker/src/assignability/assignment_checker/assignment_ops.rs
+++ b/crates/tsz-checker/src/assignability/assignment_checker/assignment_ops.rs
@@ -684,6 +684,10 @@ impl<'a> CheckerState<'a> {
             }
         }
 
+        if !is_const && self.error_invalid_commonjs_export_property_assignment(left_idx) {
+            return self.get_type_of_node(right_idx);
+        }
+
         if !is_const && self.is_commonjs_exports_property_declaration(left_idx) {
             // In JS files, `exports.X = value` is a declaration, not an assignment.
             // The type is inferred from the union of all assigned values, so individual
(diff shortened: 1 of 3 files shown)


mohsen1:
Babysitting status from the conformance loop: this PR is mergeable and all reported checks are green (Bench Compare, GitGuardian, conformance, emit, fourslash, lint, prep-node, test-* and wasm-build). It is currently blocked only by repository branch protection / required review; I did not find failing checks, merge conflicts, or requested changes to address.

PR #732, opened 2026-04-21
fix(checker): resolve root identifier of JSX member-access tag name

mohsen1:
## Summary

For JSX opening elements whose tag name is a property-access chain (e.g. `<React.Fragment>`, `<Foo.Bar.Baz />`), walk up to the root identifier and — if it's PascalCase — call `get_type_of_node_with_request` on it before dispatching JSX opening-element checking.

Previously a reference like `<React.Fragment>` in a file that never imported `React` was silently resolved through the member-access chain and produced no "cannot find name" diagnostic at the root. Forcing resolution on the root identifier routes it through the normal name-resolution path, so TS2304/TS2874 fires anchored at the `React` token rather than being swallowed.

## Test plan

- [x] New cross-file test `jsx_member_component_missing_root_reports_at_member_tag_root` uses the real react18 global lib fixture + a file that uses `<React.Fragment>` without importing React; verifies the missing-name diagnostic is anchored at the root identifier's position.
- [x] `cargo nextest run -p tsz-checker --test jsx_component_attribute_tests` — 141/141 pass (140 prior + 1 new).
- [x] `cargo nextest run -p tsz-checker --lib` — 2600/2600 pass.
- [x] Pre-commit full workspace test suite — 12925+ tests passing.

## Notes

Recovered from an interrupted parallel agent worktree (codex/conformance-20260421172436-10). Detection is based on AST shape (property-access kind + identifier root) — no string matching. The PascalCase gate keeps this from accidentally forcing resolution on DOM-style lowercase namespaced tags.

### DIFF
--- a/crates/tsz-checker/src/dispatch.rs
+++ b/crates/tsz-checker/src/dispatch.rs
@@ -1654,6 +1654,32 @@ impl<'a, 'b> ExpressionDispatcher<'a, 'b> {
                                 .checker
                                 .get_type_of_node_with_request(tag_name_idx, request);
                         }
+                        if let Some(tag_name_node) = self.checker.ctx.arena.get(tag_name_idx)
+                            && tag_name_node.kind == syntax_kind_ext::PROPERTY_ACCESS_EXPRESSION
+                        {
+                            let mut root_idx = tag_name_idx;
+                            while let Some(root_node) = self.checker.ctx.arena.get(root_idx)
+                                && root_node.kind == syntax_kind_ext::PROPERTY_ACCESS_EXPRESSION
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status from the conformance loop: no review comments or requested changes found. Visible checks are green. Current blocker appears to be draft/branch-protection state rather than a failing check or merge conflict.

mohsen1:
Babysitting status from the conformance loop: this PR is mergeable and all listed checks are green. Current blocker is draft status / repository review gate; I found no failing checks, merge conflicts, or requested changes to address.

PR #738, opened 2026-04-21
fix(parser): recover bare try blocks in class bodies

mohsen1:
## Root cause

In class member lists, tsz parsed a bare `try { ... }` as a malformed class member and consumed enough tokens to cascade into the wrong recovery path, producing an extra orphan-catch diagnostic instead of matching tsc's `TS1068` at `try` and reparsing the `try` block as an outer statement.

## Fixed conformance target

`TypeScript/tests/cases/compiler/propertyWrappedInTry.ts`

```ts
class Foo {
    try {
        public bar = someInitThatMightFail();
    } catch (e) {}

    public baz() {
        return this.bar;
    }
}
```

## Fix

Detect a bare `try` followed by `{` in a class body, emit `TS1068` at the `try` token, and leave the token unconsumed so normal statement recovery reparses the block and matches tsc's diagnostic sequence.

## Unit test

Added `bare_try_block_in_class_body_reparses_as_outer_statement` in `crates/tsz-parser/tests/state_statement_tests.rs`.

## Verification

- `cargo nextest run -p tsz-parser --lib bare_try_block_in_class_body_reparses_as_outer_statement`
- `./scripts/conformance/conformance.sh run --filter "propertyWrappedInTry" --verbose`
- `scripts/session/verify-all.sh`
  - formatting: passed
  - clippy: passed
  - unit tests: passed
  - conformance: improved, `12083` vs baseline `12061` (`+22`), includes `propertyWrappedInTry.ts`
  - emit tests: improved, JS `+5`, DTS `+25`
  - fourslash/LSP: no change, 50/50 passed

Final rebase against `origin/main` was a no-op before push.

### DIFF
--- a/crates/tsz-parser/src/parser/state_statements_class_members.rs
+++ b/crates/tsz-parser/src/parser/state_statements_class_members.rs
@@ -867,6 +867,14 @@ impl ParserState {
         while !self.is_token(SyntaxKind::CloseBraceToken)
             && !self.is_token(SyntaxKind::EndOfFileToken)
         {
+            if self.is_token(SyntaxKind::TryKeyword) && self.look_ahead_is_try_block_same_line() {
+                self.parse_error_at_current_token(
+                    "Unexpected token. A constructor, method, accessor, or property was expected.",
+                    diagnostic_codes::UNEXPECTED_TOKEN_A_CONSTRUCTOR_METHOD_ACCESSOR_OR_PROPERTY_WAS_EXPECTED,
+                );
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status (2026-04-21): PR is mergeable, with no review comments, no inline review comments, and no requested changes found. Current blocker is pending CI: `Bench Compare Against Base` is still in progress; GitGuardian is green. No merge conflict or local fix found to apply.

PR #739, opened 2026-04-21
fix(module-resolver): classify path-mapping extension from resolved file

mohsen1:
## Summary

`try_path_mappings` was passing the pre-resolution `candidate` path to `ModuleExtension::from_path`. At that point `candidate` is guaranteed to have **no** extension — targets that already carried one are filtered out a few lines earlier at `has_path_mapping_target_extension`. So every path-mapping-resolved module silently received `ModuleExtension::Unknown` instead of the real `.ts` / `.d.ts` / `.cts` / `.json` classification.

This desynced the path-mapping exit from every other resolver exit (relative, node_modules, exports, self-reference), all of which classify on the resolved file.

**Fix:** classify `ModuleExtension::from_path(&resolved)` like the rest of the resolver (15+ matching call sites across `relative_resolution.rs`, `node_modules_resolution.rs`, `exports_imports.rs`, `self_reference.rs`, `mod.rs`).

## Regression test

`test_path_mapping_extension_reflects_resolved_file` writes `src/widget.ts`, configures a `@app/*` → `src/*` path mapping, resolves `@app/widget`, and asserts `module.extension == ModuleExtension::Ts`.

Pre-fix behavior:
```
assertion `left == right` failed: path-mapping must classify extension
from the resolved file, not the extensionless candidate
  left: Unknown
 right: Ts
```

## Test plan

- [x] `cargo nextest run -p tsz-core --lib test_path_mapping_extension_reflects_resolved_file` — pass
- [x] `cargo nextest run -p tsz-core --lib module_resolver` — 147 pass
- [x] `cargo clippy -p tsz-core --lib --tests -- -D warnings` — clean
- [x] Verified the new test fails against the unfixed `path_mapping.rs`

Per DRY audit 2026-04-21 bug-shape finding #2.

### DIFF
--- a/crates/tsz-core/src/module_resolver/path_mapping.rs
+++ b/crates/tsz-core/src/module_resolver/path_mapping.rs
@@ -42,6 +42,13 @@ impl ModuleResolver {
                     let candidate = base.join(&substituted);
 
                     if let Some(resolved) = self.try_file_or_directory(&candidate) {
+                        // `candidate` has no extension here (we skipped
+                        // targets that already did on line 34), so classify
+                        // on the resolved path instead — otherwise the
+                        // extension is always `Unknown`, which desyncs path-
+                        // mapping-resolved modules from every other resolver
+                        // exit that sets this from the resolved file.
+                        let extension = ModuleExtension::from_path(&resolved);
                         return PathMappingAttempt {
                             resolved: Some(ResolvedModule {
                                 resolved_path: resolved,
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status (2026-04-21): PR is mergeable, with no review comments, no inline review comments, and no requested changes found. Current blocker is pending CI: `Bench Compare Against Base` is still in progress; GitGuardian is green. No merge conflict or local fix found to apply.

PR #740, opened 2026-04-21
fix(lsp): count code-lens references correctly at project level

mohsen1:
## Summary
- Fixes bug-shape #6 from `docs/DRY_AUDIT_2026-04-21.md`: the project-level `resolve_code_lens` reference counter was inflated by 1 for every symbol.
- The old code compared a zero-width declaration `position` (the `function`-keyword offset) against full identifier-span reference ranges, so `has_decl_reference` was never true and the count never subtracted the declaration that `find_references` always appends.
- Mirrors the in-file path (`code_lens::resolve_references_lens`): `let ref_count = references.len().saturating_sub(1);`.

## Test plan
- [x] `cargo nextest run -p tsz-lsp --lib` — all 115 code-lens tests pass, including the new `test_project_resolve_code_lens_reference_count_excludes_declaration` that verifies 3 real call sites report `"3 references"` (not `"4 references"`).
- [x] Reverified by stashing the fix: the regression test reports `left: "4 references"`, confirming it catches the bug.
- [x] `cargo clippy -p tsz-lsp --lib --tests -- -D warnings` — clean.
- [x] Pre-commit suite (3707 tests, clippy, arch guard) all green.

### DIFF
--- a/crates/tsz-lsp/src/project/features.rs
+++ b/crates/tsz-lsp/src/project/features.rs
@@ -477,16 +477,16 @@ impl Project {
                 let position = data.position;
                 let references = self.find_references(file_name, position)?;
 
-                // Count references (subtract 1 if declaration is included)
-                let ref_count = if references.is_empty() {
-                    0
-                } else {
-                    // Check if any reference is at the same position as the declaration
-                    let has_decl_reference = references
-                        .iter()
-                        .any(|r| r.range.start == position && r.range.end == position);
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status (2026-04-21): PR is mergeable, with no top-level review comments, no inline review comments, and no requested changes found. Listed checks are green; current blocker appears to be repository branch protection / required review rather than a failing check or merge conflict. No local fix found to apply.

PR #744, opened 2026-04-21
fix(lsp): preserve quotes when rewriting imports for file rename

mohsen1:
## Summary
- Fixes bug-shape #7 from `docs/DRY_AUDIT_2026-04-21.md`: `Project::handle_will_rename_files` built a `TextEdit` with the outer *quoted* range and a *bare* specifier from `calculate_new_relative_path`. Applying that edit wiped the surrounding quote characters (`import { x } from "./old";` → `import { x } from ./new.ts;`).
- Roots the range-vs-content invariant in `ImportLocation::specifier_text_edit`: shrinks the range to the inner content span and leaves the quote characters untouched, so callers no longer need to re-detect single vs double quotes.
- Replaces both call sites (`project/core.rs::process_file_rename` and `project/features.rs::get_file_rename_edits`). The latter's hand-rolled inner-range construction (plus a dead `quote` variable) is gone.

## Test plan
- [x] New regression `test_project_handle_will_rename_files_preserves_quotes` — runs the full `handle_will_rename_files` path for both `"..."` and `'...'` imports, applies the resulting edits, and asserts the rewritten text still starts and ends with the original quote.
- [x] Verified the test catches the bug: stashing the `core.rs` change reproduces `import { x } from ./new.ts;` (quotes gone).
- [x] `cargo nextest run -p tsz-lsp --lib -E 'test(rename) + test(file_rename)'` — 129 tests pass.
- [x] `cargo clippy -p tsz-lsp --lib --tests -- -D warnings` — clean (needed `const fn` on the helper).
- [x] Full pre-commit suite (3707 tests) green.

### DIFF
--- a/crates/tsz-lsp/src/project/core.rs
+++ b/crates/tsz-lsp/src/project/core.rs
@@ -2414,7 +2414,6 @@ impl Project {
         new_path: &Path,
         result: &mut WorkspaceEdit,
     ) {
-        use crate::rename::TextEdit;
         use crate::rename::file_rename::FileRenameProvider;
         use crate::utils::calculate_new_relative_path;
         use std::path::Path;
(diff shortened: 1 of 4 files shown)


mohsen1:
Babysitting status (2026-04-21): PR is mergeable, not draft, with no review comments or requested changes found. Listed checks are green (`Bench Compare Against Base`, `GitGuardian Security Checks`); current blocker appears to be repository review/branch-protection gate state rather than a code failure.

PR #746, opened 2026-04-21
fix(cli): respect outDir in --build --clean tsbuildinfo path

mohsen1:
## Summary

- When a tsconfig sets `outDir`, the TypeScript build writes `.tsbuildinfo` under that directory. The `--build --clean` handler was computing `config_path.with_extension(\"tsbuildinfo\")`, which targets a sibling file next to the tsconfig that never exists in an outDir-based project — leaving the real buildinfo behind.
- Route clean through `default_build_info_path`, the same helper the build writes through, using the already-resolved absolute `out_dir` / `declaration_dir` that `ResolvedProject` carries. Drops the redundant `resolve_compiler_options` call inside the clean loop.

Part of `docs/DRY_AUDIT_2026-04-21.md` bug-shape #11.

## Test plan
- [x] `cargo nextest run -p tsz-cli --bin tsz build_clean` — both new tests pass
- [x] Verified by stashing the fix and re-running: `build_clean_removes_buildinfo_under_out_dir` fails on unfixed code (sibling was deleted, outDir file untouched) — so the test catches the regression.
- [x] `cargo clippy -p tsz-cli --bin tsz --tests -- -D warnings` clean

### DIFF
--- a/crates/tsz-cli/src/bin/tsz.rs
+++ b/crates/tsz-cli/src/bin/tsz.rs
@@ -3106,15 +3106,17 @@ fn handle_build_clean(
     verbose: bool,
 ) -> Result<()> {
     use std::fs;
-    use tsz_cli::config::resolve_compiler_options;
+    use tsz_cli::incremental::default_build_info_path;
 
     let mut deleted_count = 0;
 
     for project in graph.projects() {
-        let base_dir = &project.root_dir;
-
-        // Delete .tsbuildinfo file
-        let buildinfo_path = project.config_path.with_extension("tsbuildinfo");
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting update (2026-04-21): reproduced the failed lint locally as a rustfmt issue in `crates/tsz-cli/src/bin/tsz/tests.rs`. Pushed `65115777db` (`style(cli): format build clean test`) to `fix-build-clean-outdir`. Local checks run in isolated worktree: `cargo fmt --check`, `scripts/check-crate-root-files.sh`, and `scripts/arch/check-workspace-metadata.sh` all pass.

PR #748, opened 2026-04-21
fix(conformance): skip fingerprint compare when one side has none

mohsen1:
## Summary

- `use_fingerprint_compare` was only gated on TSC having fingerprints. The server-mode runner (legacy protocol) returns only error codes — no fingerprints — so with the TSC-only gate, every TSC fingerprint gets reported as \"missing\" on every test and server-mode conformance runs always appear to regress, even when codes match.
- Gate on both sides being non-empty, and extract the decision into a single `use_fingerprint_compare` helper called from all three existing sites so the rule can't drift again.
- Regression test covers the four corners: server-like one-sided, cache-miss one-sided, both empty, both populated.

Part of `docs/DRY_AUDIT_2026-04-21.md` bug-shape #12.

## Test plan
- [x] `cargo nextest run -p tsz-conformance use_fingerprint_compare` — new test passes
- [x] `cargo clippy -p tsz-conformance --tests -- -D warnings` clean
- [x] All three call-sites updated to the helper; no behavior change for CLI mode (both sides populated → same path)

### DIFF
--- a/crates/conformance/src/runner.rs
+++ b/crates/conformance/src/runner.rs
@@ -29,6 +29,23 @@ fn relative_display(path: &Path, base: &Path) -> String {
         .map_or_else(|_| path.display().to_string(), |p| p.display().to_string())
 }
 
+/// Decide whether fingerprint-level comparison should run between the two
+/// diagnostic sets.
+///
+/// Returns `false` whenever either side is empty. Server mode (the legacy
+/// protocol in `server_pool.rs`) returns only error codes, so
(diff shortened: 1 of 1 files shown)


mohsen1:
Babysitting status (2026-04-21): PR is mergeable, not draft, and currently has no review comments or requested changes. Current blocker is pending `Bench Compare Against Base`; `GitGuardian Security Checks` is green. No merge conflict or local fix found to apply.

PR #752, opened 2026-04-21
fix(lsp): populate sub_to_bases and reset heritage maps on clear()

mohsen1:
## Summary

Two latent bugs in `SymbolIndex` heritage tracking:

1. **`sub_to_bases` was always empty.** The second-pass builder called `arena.get_identifier_text(class_decl_idx)`, which only returns `Some` for Identifier nodes. On a ClassDeclaration / InterfaceDeclaration it always returned `None`, so `get_bases_for_class(...)` always returned `[]`. Upward heritage lookups (go-to-implementation, heritage-aware rename) silently did nothing. Fix: resolve the declaration's `name: NodeIndex` field to the real Identifier, and walk the declaration's own `heritage_clauses` list instead of the 50-node forward-scan heuristic (which would also have missed heritage on classes with large bodies).

2. **`clear()` leaked heritage edges.** It reset 9 maps but left `heritage_clauses` and `sub_to_bases` populated, so a fully-rebuilt index would see stale class relationships.

Part of `docs/DRY_AUDIT_2026-04-21.md` bug-shapes #4 and #5.

## Test plan
- [x] 4 new tests in `symbol_index_tests.rs` cover:
  - class with `extends` + `implements`
  - interface multi-inheritance
  - large class body (would have missed the 50-node window)
  - `clear()` wipes both heritage maps
- [x] Verified by stashing the fix and re-running — all 4 tests fail on unfixed code.
- [x] `cargo nextest run -p tsz-lsp --lib` — full suite still passes
- [x] `cargo clippy -p tsz-lsp --tests -- -D warnings` clean

### DIFF
--- a/crates/tsz-lsp/src/symbols/symbol_index.rs
+++ b/crates/tsz-lsp/src/symbols/symbol_index.rs
@@ -542,46 +542,69 @@ impl SymbolIndex {
             }
         }
 
-        // Second pass: Build sub_to_bases mapping
-        // For each ClassDeclaration/InterfaceDeclaration, extract its heritage clauses
+        // Second pass: Build sub_to_bases mapping.
+        //
+        // For each ClassDeclaration/InterfaceDeclaration, walk its own
+        // heritage_clauses directly instead of guessing at which nearby
+        // HERITAGE_CLAUSE nodes belong to this declaration.
+        //
+        // Previously this block called `arena.get_identifier_text(node_idx)`
+        // on the declaration node itself. That helper only returns text for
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status (2026-04-21): PR has no review comments or requested changes in GitHub metadata. Listed checks are green (`Bench Compare Against Base`, `GitGuardian Security Checks`); current blocker appears to be repository branch-protection/review state rather than a failing check or merge conflict. No local code fix found to apply.

PR #754, opened 2026-04-21
fix(tsz-common): dedupe "error" in COMMON_STRINGS pre-intern list

mohsen1:
## Summary

- `COMMON_STRINGS` listed `"error"` twice (line 120 under common identifiers and line 153 in the console-methods cluster). Drop the second occurrence so the pre-interned slice matches the curated intent.
- Add a regression test (`common_strings_has_no_duplicates`) that walks the slice through a `HashSet` and fails if any entry repeats, so future additions can't silently reintroduce a dupe.

Functionally harmless at runtime — `Interner::intern()` dedups by key, so the second `"error"` just wasted a slot — but flagged in `docs/DRY_AUDIT_2026-04-21.md §tsz-common` and fixing it locks in the invariant.

## Test plan

- [x] `cargo nextest run -p tsz-common` → 304 / 304 passed.
- [x] Stash-revert regression check: with the duplicate restored, the new test fails with ``COMMON_STRINGS contains duplicate entries: ["error"]``; with the fix applied it passes.
- [x] Pre-commit suite (19,808 tests across 10 crates) passed cleanly before push.

### DIFF
--- a/crates/tsz-common/src/interner/mod.rs
+++ b/crates/tsz-common/src/interner/mod.rs
@@ -150,7 +150,6 @@ const COMMON_STRINGS: &[&str] = &[
     "console",
     "log",
     "warn",
-    "error",
     "info",
     "debug",
     "document",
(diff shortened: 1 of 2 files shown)


mohsen1:
Babysitting status (2026-04-21): PR has no review comments or requested changes in GitHub metadata. Current blocker is pending CI: `Bench Compare Against Base` is still in progress; `GitGuardian Security Checks` is green. No merge conflict or local code fix found to apply.

PR #733, opened 2026-04-21
perf+fix(cli): compute DefinitionStore stats once on shared store (6% win + correctness)

mohsen1:
Stacked on #731 (and #726). Includes all three commits for review context; merge after #731.

Per-file `definition_store.statistics()` on the shared store wasted ~6% CPU and produced N× inflated stats counts. Compute once after the work loop.

**Incremental bench (ST, 3 runs min, base = PR #731 merged):**
- 91-file Any+List: 9.422s → 8.854s (**6% faster**)
- 2-file Curry+Pipe: 1.034s → 1.041s (~0%)
- utility-types: 0.109s → 0.104s (~5%)

**Cumulative with #726+#731 vs pure main** (91-file): 13.863s → 8.854s = **36% faster ST**.

Tests: 2599/2599 checker pass. Pre-existing tsc_parity_* failures unchanged.

PR #763, opened 2026-04-21
fix(cli): drop reference to undefined parallel_ds_stats

mohsen1:
## Summary

The lib-checking loop in `collect_diagnostics` references a `parallel_ds_stats` variable that doesn't exist, breaking compilation on `main`. The variable was removed in a recent refactor that moved DefinitionStore aggregation to the shared store (captured once after the loop via \`aggregated_ds_stats = project_env.shared_definition_store…\`). The per-lib `lib_ds_stats` value is no longer needed because libs share the same store.

This PR drops the dead merge call and renames the binding to `_lib_ds_stats` so the destructuring still works without an unused-variable warning.

## Repro on main

\`\`\`
\$ cargo check -p tsz-cli
error[E0425]: cannot find value \`parallel_ds_stats\` in this scope
    --> crates/tsz-cli/src/driver/check.rs:1249:17
     |
1249 |                 parallel_ds_stats.merge(&lib_ds_stats);
     |                 ^^^^^^^^^^^^^^^^^
\`\`\`

## Test plan

- [x] \`cargo check -p tsz-cli\` clean

### DIFF
--- a/crates/tsz-cli/src/driver/check.rs
+++ b/crates/tsz-cli/src/driver/check.rs
@@ -1228,7 +1228,7 @@ pub(super) fn collect_diagnostics(
                 } else {
                     QueryCache::new(&program.type_interner)
                 };
-                let (lib_diags, lib_counters, lib_ds_stats) = check_checker_lib_file(
+                let (lib_diags, lib_counters, _lib_ds_stats) = check_checker_lib_file(
                     &checker_lib_file_env,
                     lib_idx,
                     &query_cache,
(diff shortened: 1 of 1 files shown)


mohsen1:
Duplicate of #733 which landed just before this. Closing.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76941, additions=2200228, deletions=8872716, head=fix-cli-parallel-ds-stats-compile, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #764, opened 2026-04-21
test(solver): lock in index access on conditional-infer-tuple results

mohsen1:
## Summary

Verify that the conformance random-failure picker works end-to-end, and add two
solver unit tests that lock in the base-case behavior of indexed access on
tuples produced by `infer`-result conditional types.

## What this PR does

- **Submodule verification**: confirmed `git submodule update --init --depth 1 TypeScript` works against the pinned SHA from this branch, and that `scripts/session/quick-pick.sh` (the canonical random-failure picker used throughout the conformance workflow) initializes the submodule on demand and returns a valid failure descriptor.
- **Solver regression fences**: two new unit tests in `crates/tsz-solver/tests/index_access_comprehensive_tests.rs` assert that

  ```ts
  // (T extends [infer A, infer B] ? [A, B] : never)[0]  -> string
  // (((...args: T) => void) extends (...args: infer R) => void ? R : any)["length"]  -> 2
  ```

  resolve correctly when the conditional type is evaluated directly (as it is in every narrow solver path), documenting the boundary the recursive `Length<T>` / `Prepend<V, T>` patterns depend on.

## Conformance investigation (picked target)

`scripts/session/quick-pick.sh` landed on `excessPropertyCheckIntersectionWithRecursiveType.ts`:

- expected: `TS2339, TS2353`
- actual:   `TS2322, TS2353`
- missing:  `TS2339` at `grandUser.children[0].children[0].children[0]`
- extra:    `TS2322` at the innermost `children: [...]` of the `BuildTree<User, 2>` repro

Root cause (confirmed by a focused trace through `visit_lazy` / `resolve_lazy` in a debug build): when the checker writes `type El = Un[0]` into `symbol_types`, the stored form is a deferred `IndexAccess(Lazy(Un), 0)` because `get_tuple_elements(db, Lazy(Un))` (`content_predicates.rs:583`) calls the freestanding `evaluate_type(db, …)` which runs under a `NoopResolver` and cannot dereference `DefId(Un)`. Later, `should_suppress_assignability_diagnostic` re-enters the solver via `QueryCache::evaluate_type_with_options`, which also constructs a `TypeEvaluator::new(self.as_type_database())` — again with `NoopResolver` — so the deferred `IndexAccess` never collapses, and downstream error surfaces ultimately emit `TS2322` instead of `TS2353` / `TS2339`.

Concrete backtrace captured in this session (abbreviated):

```
[TypeEvaluator::new NoopResolver]
 1: get_tuple_elements
 2: get_type_from_indexed_access_type     <- type El = Un[0]

[visit_lazy] resolved=Some(Application(TypeApplicationId(6)))  // works with full resolver
[visit_lazy] final evaluate(resolved)=Tuple(TupleListId(4))    // evaluates to [string, number]
[evaluate_index_access] obj=Lazy(Un) -> eval_obj=Tuple(4)       // index [0] -> STRING

[TypeEvaluator::new NoopResolver]
 1: QueryCache::evaluate_type_with_options
 2: should_suppress_assignability_diagnostic  <- const x: El = "hi"

[visit_lazy] resolve_lazy returned None, keeping original Lazy(Un)  // NoopResolver cannot resolve DefId
[evaluate_index_access] obj=Lazy(Un) -> eval_obj=Lazy(Un)            // deferred, never collapses
-> TS2339: Property '0' does not exist on type 'Un'.
```

A minimal `tsc`-equivalent repro that fails today:

```ts
type Unwrap<T> = T extends [infer A, infer B] ? [A, B] : never;
type Un = Unwrap<[string, number]>;
type El = Un[0];               // tsz: TS2339 (bug); tsc: string
const x: El = "hi";            // tsz rejects; tsc accepts
```

Fixing this properly requires either
  (a) threading the resolver through the `QueryDatabase::evaluate_*` default impls so `QueryCache` sees the checker's `ContextResolver` rather than `NoopResolver`, or
  (b) giving `TypeInterner` shared access to `DefinitionStore` so the `Lazy(DefId) -> body` mapping is visible from any `TypeDatabase`.

Both are multi-crate architectural changes that belong in their own dedicated PR with a targeted assignability/suppression audit; they are not a drop-in patch, and a narrow checker-local suppression would be exactly the kind of symptom-level fix `.claude/CLAUDE.md §22` forbids.

The two unit tests in this PR pin down the correct behavior of the inner raw-conditional path (which already works) so that any future resolver fix cannot silently regress it.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo nextest run --package tsz-solver --lib` (5245 pass)
- [x] `cargo nextest run --package tsz-checker --lib` (2603 pass)
- [x] `cargo nextest run --package tsz-parser --package tsz-binder --package tsz-scanner --lib` (813 pass)
- [x] Two new tests in `index_access_comprehensive_tests.rs` pass.
- [x] Conformance is unaffected — this PR only adds test code.

> Note: `tsz-cli driver_tests::compile_incremental_reports_ts5033_when_tsbuildinfo_is_not_writable` fails in this sandbox environment (running as root bypasses the read-only permission bit that test relies on). Verified it fails identically on the unmodified baseline, so it is not a regression.

https://claude.ai/code/session_01BuGcY8iPWcGVtpPdJyeJCT


mohsen1:
Babysitting update from the conformance loop: no review comments are pending, but CI is blocked by a stale-main lint issue. The failing lint job in run 24751024820 reports `clippy::type_complexity` for `program_wildcard_reexports_type_only` in `crates/tsz-checker/src/context/mod.rs:1421` and `:1692`. This appears to be the same alias cleanup later handled in other branches, so this PR needs a rebase/update before merge.

mohsen1:
Babysit pass (2026-04-22): after the latest update this PR is mergeable (`mergeable=MERGEABLE`, `mergeStateStatus=UNSTABLE`) with no reviews or requested changes. The current run has the unit/build/lint jobs green; the remaining red status is the aggregate conformance gate against the stale baseline, while the conformance shards themselves pass. No practical branch fix is needed in this conformance-fix iteration.

PR #760, opened 2026-04-21
refactor(checker): reuse tsz_common numeric-literal parser

mohsen1:
## Summary

- `narrowing.rs` and `references.rs` each carried a private `parse_numeric_literal_value` + `parse_radix_digits` pair that shadowed the shared `tsz_common::numeric` helper.
- After PR #759 tightened the common parser (it now rejects separator-only radix bodies like `0x_`), the checker copies drifted: they were more permissive than both the shared helper and TypeScript itself.
- Replace all four call sites with the canonical `lit.value.or_else(|| tsz_common::numeric::parse_numeric_literal_value(&lit.text))` form and drop the local helpers.
- `strip_numeric_separators` is kept in `references.rs` because `normalize_bigint_literal` still uses it with BigInt-only acceptance rules.

Tracks the numeric-parse duplication line-item in \`docs/DRY_AUDIT_2026-04-21.md\` (tsz-common audit) — companion to #759 on the lowering side.

## Test plan

- [x] \`cargo check -p tsz-checker\` — clean
- [x] \`cargo nextest run -p tsz-checker --lib\` — 2603 passed, 9 skipped
- [ ] CI

PR #768, opened 2026-04-21
refactor(emitter): use tsz_common numeric parser in enum_es5

mohsen1:
## Summary
- `EnumES5Transformer::parse_numeric_literal_text` in `crates/tsz-emitter/src/transforms/enum_es5.rs` duplicated the radix / scientific-notation literal parsing already owned by `tsz_common::numeric::parse_numeric_literal_value`.
- The emitter copy was a **strict subset** of the common helper: it didn't handle numeric separators (`1_000`), big hex/bin literals that overflow `u64`, leading-zero octal (`0777`), or bigint suffixes.
- Route the one caller at `enum_es5.rs:806` through the common helper and delete the local copy (+1 / -19).
- Matches the pattern used by the checker (PR #760). Part of the DRY audit follow-up (`docs/DRY_AUDIT_2026-04-21.md`, emitter section: "Numeric literal parsing is repeated outside common").

## Test plan
- [x] `cargo nextest run -p tsz-emitter --lib` — 1399 passed, 2 skipped (local, pre-commit).
- [x] Pre-commit hook passed (clippy affected + checker-boundary + full test suite).
- [ ] CI passes on this PR.


mohsen1:
Closing: conformance CI flagged a 45-test regression vs baseline (12038 < 12083). The refactor isn't behavior-preserving — the new `parse_numeric_literal_value` returns `Some(f64)` for (a) radix literals that overflow u64 and (b) inputs with numeric separators, while the removed local helper returned `None` for both. Those cases fall through to a different IR emit path in `member_name_ir_node`, so the swap changes enum member-name normalization for large/separator-bearing literals. Will re-try with a narrower dedup that preserves overflow-as-None semantics, or leave as-is. PR #769 is an independent dedup in the literal downlevel path and stands on its own.

mohsen1:
Reopening: the reported 45-test regression was not caused by this PR. Main itself is at 12038/12581 passed (baseline in snapshot is 12083), so this PR's 12038 is actually **equal** to main. The 45-test gap is a pre-existing baseline staleness — some commit between the last snapshot refresh (fdb12ff8) and current main regressed conformance and the snapshot hasn't been refreshed since. This PR is net-zero and the dedup stands.

PR #769, opened 2026-04-21
refactor(emitter): use shared radix-digit parser for numeric downlevel

mohsen1:
## Summary
- `Printer::convert_numeric_literal_downlevel` in `crates/tsz-emitter/src/emitter/literals/core.rs` had four copies of the same pattern for parsing binary / octal / hex / legacy-octal digits: a fast-path `u128::from_str_radix(digits, B)` with an `unwrap_or_else` fold that accumulates digits directly as `f64` once the value overflows `u128`.
- `tsz_common::numeric::parse_radix_digits` (the helper powering `parse_numeric_literal_value`) already does the f64 accumulation in one shot and validates digits — it was just private. Promote it to `pub fn parse_radix_digits_as_f64` with a clearer name + doc, and route the four emitter sites through it.
- Net: `-28 lines` at the call sites. The emitter inherits stricter digit validation; user-visible behavior is unchanged because the scanner rejects invalid literals upstream.

Part of the DRY audit follow-up (`docs/DRY_AUDIT_2026-04-21.md`, emitter: "Numeric literal parsing is repeated outside common"). Companion to #768 which dedup'd `EnumES5Transformer::parse_numeric_literal_text` against the same common module.

## Test plan
- [x] `cargo nextest run -p tsz-common --lib` — 305 passed.
- [x] `cargo nextest run -p tsz-emitter` — 1429 passed, 2 skipped (integration suite exercising the downlevel path).
- [x] Pre-commit hook (clippy affected + checker-boundary + full test suite) passed.
- [ ] CI on this PR.


mohsen1:
Context update: the CI Conformance failure reports a regression vs baseline (12083), but this PR is at 12039 — which is **+1 over main** (main: 12038). The 45-test gap is a pre-existing baseline staleness, not caused by this PR.

PR #926, opened 2026-04-22
[codex] narrow union excess checks with partial discriminants

mohsen1:
## Summary

This is PR 1 in the Phase 5 diagnostic conformance stack.

- Allow discriminated union excess-property checking to narrow on unit discriminant properties that are present on only some union members.
- Keep partial-presence narrowing conservative by requiring the present target properties to be unit types, so payload fields like `abc: string` do not become accidental discriminants.
- Add a TS2353 regression for the `Overlapping` union shape from `excessPropertyCheckWithUnions.ts`.

## Root Cause

The existing discriminant matcher skipped a source property if any union member lacked that property. In the TypeScript conformance case, `a` and `b` each exclude one branch and both point to the first member, but tsz refused to narrow because neither property appears on every union member.

## Validation

- `cargo test -p tsz-checker --test ts2353_tests -- --nocapture`
- `cargo test -p tsz-checker --test ts2353_tests discriminated_union_excess_narrows_with_partial_discriminant_presence -- --nocapture`
- `cargo fmt --check`
- `git diff --check`
- `./scripts/conformance/conformance.sh run --filter "excessPropertyCheckWithUnions.ts" --verbose --write-diff-artifacts --diff-artifacts-dir /tmp/tsz-phase5-excess-diffs`

Filtered conformance now reports 14 expected fingerprints, 10 actual, 4 missing, and 0 extra. Before this slice the same sample had 8 actual and 6 missing; the two `Overlapping` TS2353 misses are now produced.

### DIFF
--- a/crates/tsz-checker/src/assignability/assignment_checker_tests.rs
+++ b/crates/tsz-checker/src/assignability/assignment_checker_tests.rs
@@ -626,6 +626,30 @@ let value: {
 }
 
 #[test]
+fn mapped_enum_key_missing_property_uses_enum_member_display() {
+    let diagnostics = diagnostics_for(
+        r#"
+type Record<K extends string | number, T> = { [P in K]: T };
+enum E { A }
+let foo: Record<E, any> = {};
+"#,
+    );
+
+    let diag = diagnostics
+        .iter()
+        .find(|diag| diag.code == 2741)
+        .expect("expected TS2741 for missing enum mapped key");
+    assert!(
+        diag.message_text.contains("Property '[E.A]' is missing"),
+        "TS2741 should render the enum member key, got: {diag:?}"
+    );
+    assert!(
+        !diag.message_text.contains("Property '0' is missing"),
+        "TS2741 should not render the erased numeric key, got: {diag:?}"
+    );
+}
+
+#[test]
 fn function_expression_assignment_reports_outer_signature_mismatch() {
     let source = r#"
 interface T {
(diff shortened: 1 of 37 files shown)


mohsen1:
Closing: 10 Cloud Build failures across 8 conformance shards plus lint and unit on a 21-commit stacked codex PR with no comments/reviews. The partial-discriminant narrowing change caused widespread regressions. Reopen with a rebased, reduced scope if desired — a loop-produced 21-commit stack is too speculative to fix-forward in one pass.

mohsen1:
Salvaged: rebased onto latest main (21 → 20 → 17 commits after dropping regressors).

**Changes from original**:
- Dropped `4d05c84c56 expand conditional aliases in TS2353 displays` (architecture violation: direct `tsz_solver::TypeData` import in checker/error_reporter/core/type_display.rs — §4 says checker must not pattern-match low-level type internals).
- Dropped `448f98f46c narrow union excess checks with partial discriminants` (caused 2 regressions on missingDiscriminants{,2}.ts — wrong TS2353 positions on `thing2`/`thing3` lines and missing TS2353 on `subkind` in `item2`).
- Dropped `72d26c41b6 refine excess-property union discriminants` (built on the first one, same regression class).
- Dropped `27809552fe render JS this receivers by function name` (caused 1 regression on jsdocThisType.ts — rendered `'f2'`/`'f4'`/`'f6'` instead of the resolved `'Foo'` type name).

**Local validation on salvaged tip (17 commits)**:
- `cargo nextest run -p tsz-checker -p tsz-solver` — 10,277 / 10,277 pass.
- Architecture contract tests pass.
- Full conformance: **+15 improvements, 0 regressions** (12,065 → 12,080 tests passing).

Ready for admin-merge — net +15 tests, no regressions.

PR #939, opened 2026-04-23
fix(emitter): preserve `typeof X` in arrow function parameter annotations

mohsen1:
## Summary
For an inferred-type variable like

```ts
const printFn = (action: typeof actionA | typeof actionB) => { ... };
```

the parameter annotation contains explicit `typeof` AST nodes. The existing dts emit, however, routes inferred function-typed initializers through the type-printer path, which collapses `TypeQuery` into the resolved value type and loses the syntactic form — producing `(action: Action<...> | Action<...>) => void` instead of `(action: typeof actionA | typeof actionB) => void`.

There is already an AST-based emit path (`emit_function_initializer_type_annotation`) that emits parameter annotations directly via `emit_type`, which correctly handles `TYPE_QUERY` nodes. It was previously gated by a narrow set of conditions (inline param comments, self-returning, unique return identifier). This change adds a fourth gate (`function_initializer_has_typeof_in_param_annotation`) that recursively walks each parameter's type annotation looking for `TYPE_QUERY` — including unions, intersections, parens, arrays, tuples, optional/rest wrappers — and routes through the AST path when one is found.

## Impact
- `coAndContraVariantInferences` now passes
- Full dts emit suite: **+1 passing test**

## Test plan
- [x] New unit test `test_arrow_initializer_preserves_typeof_in_param_annotation`
- [x] Pre-commit suite (~4358 tests) passes
- [x] No dts regressions in full suite

### DIFF
--- a/crates/tsz-emitter/src/declaration_emitter/helpers/variable_decl.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/helpers/variable_decl.rs
@@ -125,7 +125,8 @@ impl<'a> DeclarationEmitter<'a> {
             } else if has_initializer
                 && (self.function_initializer_has_inline_parameter_comments(initializer)
                     || self.function_initializer_is_self_returning(initializer)
-                    || self.function_initializer_returns_unique_identifier(initializer))
+                    || self.function_initializer_returns_unique_identifier(initializer)
+                    || self.function_initializer_has_typeof_in_param_annotation(initializer))
                 && {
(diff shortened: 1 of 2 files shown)


mohsen1:
Closing as superseded by 6a79e249aa (`fix(emitter): preserve `typeof X` in arrow parameter annotations during DTS inference`), which is already merged to main. The merged commit covers the same emit path and ships the same helper (renamed to the plural `function_initializer_has_typeof_in_param_annotations`).

PR #985, opened 2026-04-23
fix(checker): anchor TS2769 at callee when overloads disagree on an object literal

mohsen1:
## Summary
- tsc anchors TS2769 at the **callee** (not the argument) when all overload candidates reject the same object literal argument with **different** failure messages — e.g. each overload rejects a different excess property.
- tsz was unconditionally anchoring at the shared argument whenever all overloads produced TS2345-style failures, causing a fingerprint-only TS2769 position divergence on `orderMattersForSignatureGroupIdentity.ts`.
- The fix narrows the argument-anchor heuristic: for object-literal arguments, require `identical_argument_failures` (all overload failures render the same text). Non-object-literal arguments (e.g. `fn(true)` vs `(x:string)|(x:number)`) still anchor at the argument.

## Root cause
When overloads disagree on *why* the argument is wrong, the overload set — not the argument — is the culprit. tsc moves the diagnostic to the callee so the user sees "no overload fits" rather than "this argument is wrong for one of them." The previous tsz condition `all_failures_are_argument_mismatches && shared_argument_anchor.is_some()` missed this distinction because the shared-argument anchor only looks at span equality, not message equality.

## Reproducer
```ts
interface A {
    (x: { s: string }): string;
    (x: { n: number }): number;
}
declare var v: A;
v({ s: "", n: 0 });
//  ^
// tsc:       TS2769 anchored at col 1 (`v`)
// tsz before: TS2769 anchored at col 5 (`{`)
// tsz after:  TS2769 anchored at col 1 (`v`)
```

## Impact (verify-all)
- Conformance: **+1** (12066 vs 12065 baseline).
- Emit: JS +0, DTS +16.
- Fourslash: unchanged (50/50).
- Checker unit tests: 4978/4978 pass.
- `orderMattersForSignatureGroupIdentity.ts` still has a secondary divergence on TS2339 result type (`never` vs `number`) — a separate call-resolution-failure semantics issue I did not attempt here. Once that lands, this test flips fully.

## Test plan
- [x] New `crates/tsz-checker/tests/ts2769_anchor_disagreeing_overloads_tests.rs` with two cases: anchor moves to callee for object-literal + disagreeing overloads; anchor stays on argument for non-object-literal cases (`fn(true)` vs `(x:string)|(x:number)`).
- [x] Existing `ts2769_overload_related_information_keeps_overload_order` unit test preserved (primitive argument case).
- [x] `./scripts/conformance/conformance.sh run --filter orderMattersForSignatureGroupIdentity --verbose` confirms TS2769 anchor now matches tsc.
- [x] `scripts/session/verify-all.sh`: all 6 suites pass, no regressions.

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -36,6 +36,10 @@ name = "generator_annotation_mismatch_display_tests"
 path = "tests/generator_annotation_mismatch_display_tests.rs"
 
 [[test]]
+name = "ts2769_anchor_disagreeing_overloads_tests"
+path = "tests/ts2769_anchor_disagreeing_overloads_tests.rs"
+
+[[test]]
 name = "ts1362_false_positive_tests"
 path = "tests/ts1362_false_positive_tests.rs"
 
--- a/crates/tsz-checker/src/error_reporter/call_errors/error_emission.rs
+++ b/crates/tsz-checker/src/error_reporter/call_errors/error_emission.rs
@@ -346,8 +346,23 @@ impl<'a> CheckerState<'a> {
         let anchor_argument_from_mixed_failures = shared_argument_anchor.is_some()
             && !remaining_failures.is_empty()
             && remaining_failures_are_count_mismatches;
-        let anchor_argument_from_all_failures =
-            all_failures_are_argument_mismatches && shared_argument_anchor.is_some();
+        // When all overload failures share the same argument anchor but the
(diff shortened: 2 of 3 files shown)


mohsen1:
Re-firing CI after trigger update (COMMENTS_ENABLED → COMMENTS_ENABLED_FOR_EXTERNAL_CONTRIBUTORS_ONLY)

PR #986, opened 2026-04-23
perf(cli): share program-wide symbol_arenas via Arc instead of per-file build

mohsen1:
## Summary

`create_binder_from_bound_file_with_augmentations` was iterating `program.symbol_arenas` (potentially ~100K entries on large-ts-repo) for every per-file checker binder, building a filtered per-file map. Across 6086 files this was ~600M iter ops plus N×N `Arc::clone` churn — pure startup tax that the perf doc flagged as a sibling of the `declaration_arenas` hot spot.

## Changes

- Wraps `BinderState.symbol_arenas`, `BindResult.symbol_arenas`, and `MergedProgram.symbol_arenas` in `Arc<FxHashMap<...>>`.
- Replaces the per-file build with a single `Arc::clone(&program.symbol_arenas)`.
- Routes mutation sites through `Arc::make_mut` (zero-cost when refcount=1, which is always during binding).
- Per-file `BoundFile.symbol_arenas` stays as `FxHashMap` (small per-file data); per-binder construction in the parallel path wraps with `Arc::new(file.symbol_arenas.clone())` once.

## Safety

The previous filter dropped local-only entries (where the symbol's recorded arena equals the per-file arena and the symbol has no cross-file declaration). Keeping those entries is a no-op for consumers — every checker access of `binder.symbol_arenas` is a point lookup (`.get(&sym_id)`); no iter consumers exist. Including the local entry only ever returns the local arena that the consumer would have found anyway.

Mirrors the Arc-share treatment in #932/#935/#944/#948/#954/#960/#973/#975/#979.

## Test plan
- [x] `cargo nextest run -p tsz-binder -p tsz-core -p tsz-checker` (8255 passed, 47 skipped)
- [x] Pre-commit hook (clippy + nextest + arch guardrails) green

### DIFF
--- a/crates/tsz-binder/src/state/core.rs
+++ b/crates/tsz-binder/src/state/core.rs
@@ -184,7 +184,7 @@ impl BinderState {
             current_scope_idx: 0,
             node_symbols: FxHashMap::with_capacity_and_hasher(256, Default::default()),
             module_declaration_exports_publicly: FxHashMap::default(),
-            symbol_arenas: FxHashMap::default(),
+            symbol_arenas: Arc::new(FxHashMap::default()),
             declaration_arenas: FxHashMap::default(),
             cross_file_node_symbols: FxHashMap::default(),
             node_flow: FxHashMap::with_capacity_and_hasher(128, Default::default()),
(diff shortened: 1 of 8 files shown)


mohsen1:
Re-firing CI after trigger update

PR #1007, opened 2026-04-23
docs(perf): design for instantiate_type cross-call cache

mohsen1:
## Summary

Captures the concrete design for closing the 16× tsz-vs-tsgo gap on `ts-essentials/deep-readonly.ts` and similar recursive-utility-type benchmarks.

`TypeInstantiator` is constructed **per call** today at `crates/tsz-solver/src/instantiation/instantiate.rs:1440`. Its `visiting` cache is scoped to one invocation, so two sibling callers evaluating the same `(type_id, substitution)` redo the entire recursive walk. `tsgo` hides this cost behind a type cache; `tsz` does not. This doc details how to add one safely.

## Design at a glance

- **Cache key**: `(TypeId, InternedSubstId, InstantiatorMode, Option<this_type>)` where `InternedSubstId` is a new interner handle over a sorted canonical `SmallVec<[(Atom, TypeId); 4]>` (`TypeSubstitution` is currently `FxHashMap`-backed — not content-hashable as-is).
- **Storage**: new `InstantiationCache` on `QueryCache` alongside the existing `eval_cache` / `application_eval_cache`. The name `InstantiationCache` is already architecturally reserved (`architecture_contract_tests.rs:2259`) — the checker is already forbidden from reaching into it.
- **Correctness invariant**: `depth_exceeded` / ERROR results are **never cached** (same discipline tsc uses for `recursionIdentity` failures). Unit test #4 in the doc is the tripwire.
- **Ships as 4 PRs** so each increment is independently verifiable:
  1. Canonical interned substitution (pure refactor)
  2. Empty cache + trait hooks + stats (pure refactor)
  3. Wire the 5 entry points (the perf win lands here)
  4. Optional `SharedQueryCache` cross-file extension (gated on measured cross-file hit rate)

## Context

Companion to `docs/plan/perf-large-repo-followup.md` §3.3 which flagged this as "the single biggest single-file perf improvement available. It is also the highest-risk." The doc addresses the risks section by section (key aliasing, sort canonicality, depth-overflow poisoning, `Lazy(DefId)` cross-file resolution).

## Scope

Pure documentation addition. No code changes.

## Test plan

- [ ] Docs-only change — `git diff --stat` shows exactly one new file under `docs/plan/`
- [ ] Pre-commit hook correctly skips Rust checks for docs-only changes (confirmed)

### DIFF
--- a/docs/plan/perf-instantiate-type-cache-design.md
+++ b/docs/plan/perf-instantiate-type-cache-design.md
@@ -1,7 +1,38 @@
 # `instantiate_type` cross-call cache — design
 
-Status: **design only**. No production code changes associated with this document.
+Status: **design — revised after review**. No production code changes associated with this document.
 Companion to: `docs/plan/perf-large-repo-followup.md` §3.3.
+
+### Review feedback addressed (2026-04-23)
+
+This revision responds to five concrete issues raised in review:
+
+1. Cache hooks must live on `QueryDatabase`, not `TypeDatabase`. The original
+   version proposed promoting the two methods onto `TypeDatabase` "with default
+   `None`/no-op" for convenience; that pushes cache concerns below the layer
+   the codebase designates as the cache boundary. §2 now keeps the hooks on
+   `QueryDatabase` and threads `&dyn QueryDatabase` into the instantiation
+   entry points.
+2. `substitute_this_type` always calls `TypeSubstitution::new()` (empty). The
+   original "skip cache when `substitution.is_empty()`" rule would disable
+   caching for every `substitute_this_type` call. §5 (PR 3) now carves out
+   "skip only when `substitution.is_empty() && this_type.is_none()`".
+3. Do not intern substitutions on `TypeInterner`. `QueryCache` doesn't own the
+   interner and doesn't clear it; interning there would leak long-lived state
+   that `QueryCache::clear()` and `estimated_size_bytes` can't see. §1 now
+   stores the canonical `SmallVec` directly in the cache key (or, optionally,
+   interns on `QueryCache` itself with matching clear/size accounting).
+4. Test strategy #3 cannot compare `TypeId`s across distinct `TypeInterner`s —
+   `TypeId` is an interner-local `u32` handle. §6 replaces it with a
+   structural/formatting comparison within a single interner.
+5. `instantiate_type` has bespoke leaf fast paths for `TypeParameter` and
+   `IndexAccess(T, P)` ahead of `TypeInstantiator::new()` (`instantiate.rs:1449–1468`).
+   §5 (PR 3) explicitly requires the cache probe to run **after** those fast
+   paths, not before, so hot leaf substitutions stay allocation-free.
+
+Additionally: §5 now declares whether `instantiate_generic` (`instantiate.rs:1596`)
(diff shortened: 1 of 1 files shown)


mohsen1:
## Review feedback addressed

Force-pushed a revision that addresses all five issues raised:

1. **Cache hooks on `QueryDatabase`, not `TypeDatabase`.** The prior revision proposed promoting the two hooks onto `TypeDatabase` with `None`/no-op defaults for the convenience of the instantiation entry points (which currently take `&dyn TypeDatabase`). That was the wrong direction — it pushed cache concerns below the `QueryDatabase` boundary the codebase designates for cache/incremental state. §2 now keeps the hooks on `QueryDatabase` and explicitly calls out that PR 3 will change the five entry-point signatures from `&dyn TypeDatabase` to `&dyn QueryDatabase`. `TypeInterner` already implements `QueryDatabase` (`db.rs:977`), so most call sites don't need to change what they pass.

2. **`substitute_this_type` carve-out.** You caught a concrete contradiction: `substitute_this_type` always passes `TypeSubstitution::new()` (`instantiate.rs:1635`), so the "skip cache when `substitution.is_empty()`" rule would disable caching for every `substitute_this_type` call. §5 PR 3 now has the explicit three-way rule:
   ```
   if substitution.is_empty() && this_type.is_none() → skip
   else if substitution.is_identity_for(...)         → skip
   else                                              → probe
   ```
   Plus a test #9 that verifies same-`this_type` calls hit and different-`this_type` calls miss.

3. **No `TypeInterner`-based interning.** Dropped. §1 now stores `CanonicalSubst(SmallVec<[(Atom, TypeId); 4]>)` directly in the key. Added a "Why no `TypeInterner` intern handle" subsection recording the rejected design — the reasoning is exactly yours: `QueryCache` doesn't own the interner, `QueryCache::clear()` can't clear it, and `estimated_size_bytes()` can't count it, so interning there is an unaccounted memory-growth source. If profiling ever demands dedup, a **`QueryCache`-local** intern table is the right place.

4. **Test strategy #3 fixed.** You're right — `TypeId` is an interner-local `u32` handle; comparing across fresh `TypeInterner`s is meaningless. §6 test #3 now constrains the parity check to a single interner and adds an explicit guardrail comment: for any cross-interner cross-check (cache-on/cache-off invariant), use `DisplayType::to_string` or a structural walk, never `TypeId` equality.

5. **Leaf fast paths preserved.** §5 PR 3 now requires cache-key construction to run **after** the existing `TypeParameter` and `IndexAccess(T, P)` fast paths at `instantiate.rs:1449–1468`. A new test #8 asserts the cache-miss counter is unchanged after N leaf substitutions, so a future refactor that accidentally probes the cache for `TypeParameter`-direct hits gets caught in CI.

**Plus**: §5 explicitly declares `instantiate_generic` (`instantiate.rs:1596`) **out of scope for PR 3**. It constructs a fresh `TypeInstantiator` too, but the common call path (evaluator tail-call application) already hits `application_eval_cache`. Wiring it into the new cache risks double-caching with different keys. A PR 5 can revisit if profiling shows non-application hot spots remain.

Diff: +229 / −56 lines. The running agent on PR 1/4 has been redirected to the `CanonicalSubst`-only scope — no `TypeInterner` changes.

PR #1020, opened 2026-04-23
fix(solver): reject non-generic source against constrained generic target

mohsen1:
## Summary

- Fixes `assignmentCompatWithConstructSignatures4.ts` conformance test: non-generic concrete constructor type was incorrectly assignable to a constrained generic constructor type
- Root cause: in `check_function_subtype_impl`, the `erase_generics=true` branch erased target type params to their constraints before structural comparison — allowing `Base` to pass against `T extends Base` because after erasure `T` becomes `Base`
- Fix: align with tsc's `compareSignaturesRelated` semantics where `eraseGenerics = relation === comparableRelation` — for the **assignable** relation, type params are never erased; the raw TypeParam types remain so structural comparison correctly rejects concrete types against universally-quantified params
- Also fixes `assignmentCompatWithConstructSignatures3.ts` (5 fingerprints were failing for the same root cause)

## Root cause (tsc reference)

```typescript
// TypeScript/src/compiler/checker.ts line 24673:
const eraseGenerics = relation === comparableRelation;
```

tsc only erases generics for the **comparable** relation, not the assignable relation. A concrete `new (x: Base) => Base` is **not** assignable to `new <T extends Base>(x: T) => T` because `T` could be instantiated as `Derived`.

## Changes

- `crates/tsz-solver/src/relations/subtype/rules/functions/checking.rs`: removed the `erase_generics=true` constraint-erasure path; unified to simply clear `type_params`, leaving raw TypeParam IDs in place for accurate structural comparison
- `crates/tsz-solver/tests/callable_tests.rs`: corrected test assertion (was testing wrong/buggy behavior)  
- `crates/tsz-checker/tests/ts2322_tests.rs`: added regression test for concrete-not-assignable-to-constrained-generic

## Test results

- `assignmentCompatWithConstructSignatures4.ts`: 1/1 PASS (was 0/1)
- `assignmentCompatWithConstructSignatures3.ts`: 1/1 PASS (was 0/1)
- 33/33 callSignatures conformance tests: PASS
- 4/4 constructSignatures conformance tests: PASS
- 123/128 assignmentCompat tests: PASS (5 pre-existing failures unrelated to this fix)
- 18,505 unit tests: all passed

## Test plan

- [ ] Verify `assignmentCompatWithConstructSignatures4.ts` passes: `./scripts/conformance/conformance.sh run --filter "assignmentCompatWithConstructSignatures4" --verbose`
- [ ] Verify `assignmentCompatWithConstructSignatures3.ts` passes: `./scripts/conformance/conformance.sh run --filter "assignmentCompatWithConstructSignatures3" --verbose`
- [ ] Verify no solver regressions: `cargo nextest run -p tsz-solver --lib`
- [ ] Verify no checker regressions: `cargo nextest run -p tsz-checker --lib`

### DIFF
--- a/crates/tsz-checker/tests/ts2322_tests.rs
+++ b/crates/tsz-checker/tests/ts2322_tests.rs
@@ -4143,3 +4143,47 @@ u2.email = e;
         "Expected TS2412 for exact-optional property write mismatch, got: {diagnostics:#?}"
     );
 }
+
+/// Regression test for assignmentCompatWithConstructSignatures4.ts:
+///
+/// A non-generic concrete construct signature is NOT assignable to a
+/// constrained-generic construct signature. The concrete types in the source
+/// may match the constraints of the target's type parameters, but T could be
+/// instantiated with a MORE SPECIFIC subtype (e.g., T extends Base means T
+/// could be Derived). So the assignment is unsound.
+///
+/// Previously tsz incorrectly allowed this by erasing target type params to
+/// their constraints, making the two signatures look identical.
+///
+/// See: TypeScript test assignmentCompatWithConstructSignatures4.ts line 49-50.
+#[test]
+fn nongeneric_concrete_not_assignable_to_constrained_generic_construct_sig() {
+    let source = r#"
+class Base { foo: string; }
+class Derived extends Base { bar: string; }
+class Derived2 extends Derived { baz: string; }
+
+// a7: non-generic concrete construct signature
+declare var a7: new (x: (arg: Base) => Derived) => (r: Base) => Derived2;
+// b7: generic construct signature with constrained type params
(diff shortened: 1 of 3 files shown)


mohsen1:
Closing due to 2 confirmed PASS→FAIL regressions after local full-conformance run:
- `circularlySimplifyingConditionalTypesNoCrash.ts`
- `generatorYieldContextualType.ts`

Net is +6 (6 new passes, 2 regressions, +4 net), but the PR body claimed no regressions. The solver tightening is architecturally correct (aligns with tsc's `eraseGenerics = relation === comparableRelation`) but widens the no-erasure path beyond what those two tests tolerate.

Branch and analysis are preserved. Follow-up should either:
1. Tighten the no-erasure condition (e.g. only for subtypeRelation, not both subtype+assignability), or
2. Investigate what the two regressing tests now fail on, and add the narrow precondition to avoid them.

Closing so main stays clean while the issue is analyzed.

PR #1028, opened 2026-04-24
fix(checker): correct TS2322 fingerprints for async block-body arrows with JSDoc callable @type

mohsen1:
## Summary

Fixes the conformance test `asyncArrowFunction_allowJs.ts` which had a fingerprint-only failure: tsc emits `TS2322 file.js:13:2 Type 'number' is not assignable to type 'string'` at the `return` statement inside a block-body async arrow, but tsz was missing that diagnostic.

Three root-cause bugs combined to produce wrong fingerprints:

1. **`function_initializer_body_has_error` returned `Some(false)` for block bodies** (hardcoded) — never suppressed the outer TS2322 check when the body already had an error. Fixed to scan diagnostics emitted inside the block for TS2322/TS2339.

2. **No mechanism to skip the outer declaration-level TS2322** for async+block-body+JSDoc-callable assignments. Added an `is_async_block_body_with_jsdoc_callable_type` guard: for `const c: () => string = async () => { ... }` in JS files, skip the outer type mismatch and let `check_return_statement` report on the individual `return` statements.

3. **`body_return_type` for async+contextual used inferred type instead of contextual return type.** In `function_type.rs`, the async branch for contextually-typed functions was computing `body_return_type` from the inferred `Promise<number>` (unwrapped to `number`), not the contextual `string` from `function(): string`. Added a `has_contextual_return` branch that uses `return_context_for_circularity` directly, so `check_return_statement` sees `string` as the expected type and correctly reports `number` is not `string`.

## Test plan

- [x] New unit tests in `jsdoc_type_tag_tests.rs`:
  - `test_async_concise_arrow_jsdoc_callable_wrong_return_type` — async `() => 0` with `@type {function(): string}` → exactly 1 TS2322
  - `test_async_block_body_arrow_jsdoc_callable_wrong_return_type_single_error` — async `() => { return 0 }` with `@type {function(): string}` → exactly 1 TS2322 at return stmt
  - `test_async_block_body_arrow_jsdoc_callable_correct_return_type_no_error` — async `() => { return "" }` → 0 TS2322
- [x] All 13117 pre-commit tests pass
- [x] Conformance: `asyncArrowFunction_allowJs.ts` now 1/1 pass (was fingerprint-only failure)

### DIFF
--- a/crates/tsz-checker/src/state/variable_checking/core.rs
+++ b/crates/tsz-checker/src/state/variable_checking/core.rs
@@ -834,9 +834,10 @@ impl<'a> CheckerState<'a> {
                             }
                             let func = checker.ctx.arena.get_function(init_node)?;
                             let body_node = checker.ctx.arena.get(func.body)?;
-                            if body_node.kind == syntax_kind_ext::BLOCK {
-                                return Some(false);
-                            }
+                            // For block bodies, check if any TS2322/TS2339 was emitted
+                            // inside the block during the initializer check. If so, the
+                            // body already produced the canonical diagnostic and the outer
+                            // declaration-level check should be skipped.
                             Some(
                                 checker.ctx.diagnostics[init_snap.diagnostics_len..]
                                     .iter()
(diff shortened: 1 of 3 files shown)


mohsen1:
Closing due to 2 confirmed PASS→FAIL regressions after local full-conformance run:
- `generatorYieldContextualType.ts`
- `contextuallyTypeAsyncFunctionReturnType.ts`

Net is +4 (6 new passes, 2 regressions). The `body_return_type` change for `is_async_for_context && has_contextual_return` is too aggressive — it now reaches through to the contextual type in scenarios where the previous Promise-unwrapping path was correct (the two regressed tests exercise contextual async return typing specifically, plus generator yield contextual typing).

Branch and analysis are preserved. Follow-up should narrow the contextual-return branch so it only fires for JSDoc `@type {function(): X}` callable types (the original bug), not generic contextual async typing.

PR #1035, opened 2026-04-24
fix(checker): elaborate object-literal properties for typed destructuring var decls

mohsen1:
## Root cause

When a destructuring variable declaration has a type annotation and an object-literal RHS with per-property type mismatches, `tsc` reports **individual TS2322 errors** anchored at each mismatching property name inside the object literal. `tsz` was instead emitting a **single whole-type TS2322** at the outer binding pattern.

```ts
var {a1, a2}: { a1: number, a2: string } = { a1: true, a2: 1 }
//                                               ^^        ^
//  tsc reports here (property names in RHS)    col 46    col 56
```

**Before:** one `TS2322: Type '{ a1: boolean; a2: number; }' is not assignable to type '{ a1: number; a2: string; }'` at col 5.  
**After:** two `TS2322` errors — `'boolean' is not assignable to 'number'` at `a1:`, `'number' is not assignable to 'string'` at `a2:`.

## Fix

The `is_destructuring` branch in `check_variable_declaration_with_request` called `try_elaborate_initializer_elements` (array-only) and fell back to a generic error for object-literal initializers. Added a call to `try_elaborate_object_literal_properties_for_var_init` before the generic fallback, mirroring what the non-destructuring branch already does.

Change is in `crates/tsz-checker/src/state/variable_checking/core.rs`.

## Tests

- New unit test `destructuring_var_decl_with_type_annotation_elaborates_object_literal_properties` in `ts2322_tests.rs`
- Conformance test `destructuringVariableDeclaration2.ts` flips from FAIL (fingerprint-only) to PASS
- All 13118 pre-commit unit tests pass
- No regressions in `--filter destructuring` suite (159/174, same as baseline)

### DIFF
--- a/crates/tsz-checker/src/state/variable_checking/core.rs
+++ b/crates/tsz-checker/src/state/variable_checking/core.rs
@@ -888,24 +888,42 @@ impl<'a> CheckerState<'a> {
                                 declared_type,
                                 var_decl.initializer,
                             ) {
-                                let skip_generic_outer_error = checker
-                                    .ctx
-                                    .arena
-                                    .get(var_decl.name)
-                                    .and_then(|pattern_node| {
-                                        checker.ctx.arena.get_binding_pattern(pattern_node)
-                                    })
-                                    .is_some_and(|pattern| {
-                                        pattern.elements.nodes.is_empty()
-                                            && var_decl.type_annotation.is_some()
-                                    });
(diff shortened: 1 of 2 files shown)


mohsen1:
Closing due to persistent failures (unit test + conformance-0). The fix may conflict with recent merges to main; rebase and re-open if still applicable. Branch preserved.

PR #1125, opened 2026-04-24
perf(solver): skip impossible subtype pairs in remove_subtypes_for_bct via name fingerprint

mohsen1:
## Summary

`remove_subtypes_for_bct` does an O(N²) pairwise `is_subtype_of` check across union members to drop subsumed types. On synthetic stress cases like `BCT candidates=200` (1.71× slower than tsgo per `docs/plan/perf-large-repo-followup.md` §3.4), the SubtypeChecker call dominates — and the vast majority of pairs are sibling classes with disjoint required-property sets that can NEVER be in a subtype relation.

## What it does

Pre-computes a cheap `TypeNameFingerprint` per candidate (sorted Atom list of non-optional property names + has_string_index + has_number_index) and uses it as a fast pre-filter:

- If the target has a required property name the source definitely lacks (and the source has no index signature absorbing it), the pair cannot be in any subtype relation. Skip the SubtypeChecker call.
- Otherwise, fall through to the full subtype check exactly as before.

## Conservatism guarantees

- Fingerprint is `None` for types we can't cheaply analyse (unresolved Lazy, applications, unions, intersections) — full check runs.
- A positive `source_covers_target_names` result **never** short-circuits to `true`. We only ever skip pairs that would definitely fail the full subtype check, so semantics are preserved.

## Tests

`crates/tsz-solver/tests/expression_ops_tests.rs` — 119 new lines:
- Disjoint property names: skip path produces correct BCT
- Source has index signature: fall-through preserved
- Optional-only target: not skipped (optional names don't gate the check)
- Union/intersection inputs: fingerprint=None, fall-through preserved
- Two compatible classes: full subtype check still runs

## Validation

- [x] `cargo check -p tsz-solver` clean
- [x] `cargo clippy -p tsz-solver --all-targets --all-features -- -D warnings` clean
- [x] `cargo nextest run -p tsz-solver --tests` — 5288 tests pass
- [ ] Bench delta on `BCT candidates=200` deferred — single-file synthetic synthetic stress is simple to spot-check via hyperfine on a generated input but the skip-condition (sibling classes with disjoint required-property sets) is what dominates in real-world conformance, where this is hard to attribute without flame profile

🤖 Co-authored by background Opus agent (bct-perf-opus). Salvaged from stalled state — staged changes committed by main agent.

### DIFF
--- a/crates/tsz-solver/src/operations/expression_ops.rs
+++ b/crates/tsz-solver/src/operations/expression_ops.rs
@@ -629,6 +629,35 @@ fn remove_subtypes_for_bct<R: TypeResolver>(
     }
     let mut keep = vec![true; len];
 
+    // PERF: pre-compute per-type property-name fingerprints.
+    //
+    // For object-like candidates (Object, ObjectWithIndex, and class instance
+    // types reached through Lazy(DefId)), a required property on the target
+    // that the source lacks *cannot* be satisfied by any subtype check. This
+    // lets us skip the expensive SubtypeChecker call for the common BCT
+    // pattern where N sibling classes share no structurally-covering pairs.
+    //
+    // Guarantees preserved:
+    // - The fingerprint is `None` for types we can't cheaply analyse (e.g.
+    //   unresolved Lazy, generic applications, unions/intersections), so the
+    //   SubtypeChecker fallback runs exactly as before.
+    // - A positive result from `source_covers_target_names` never short-
+    //   circuits to `true` — we still defer to the full subtype check. Only a
+    //   definitive negative ("target has a required name source lacks") skips
+    //   the call and keeps the pair in the union, which is semantically
+    //   conservative (matches tsc when `remove_subtypes_for_bct` is capped).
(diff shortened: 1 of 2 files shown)


mohsen1:
Closing — conformance-0 shard regressed and the BCT change had no measured bench delta. The name-fingerprint pre-filter idea is sound but conservativeness invariant must have a hole somewhere. Branch preserved; reopen after investigating which conformance test regressed and validating the skip rule against it.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76982, additions=2222239, deletions=8872786, head=perf/bct-remove-subtypes-name-fingerprint, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #1201, opened 2026-04-25
[WIP] fix(checker): preserve declared initializer type for class properties without annotation

mohsen1:
## Intent

When a class field has no explicit type annotation but the initializer expression has a declared (non-widened) type, tsz currently widens the property type to the literal-widened form, producing TS2345 false positives.

```ts
type AB = 'A' | 'B';
const DEFAULT: AB = 'A';
class C {
    D = DEFAULT;
}
declare const c: C;
declare function expectAB(x: AB): void;
expectAB(c.D); // tsz: TS2345 'string' is not assignable to 'AB'.   ← false positive
```

tsc treats `D` as `AB` (the initializer's declared type), so `expectAB(c.D)` is OK.

## Roadmap Claim

- Updated `docs/plan/ROADMAP.md` with this claim under Active Implementation Claims (workstream 1: Diagnostic Conformance).

## Planned Scope

- `crates/tsz-checker/src/types/utilities/core.rs` (`effective_class_property_declared_type` and/or its callers)
- Possibly `crates/tsz-checker/src/state/state_checking_members/member_access.rs` (`infer_property_type_from_class_member_assignments`) if widening happens there
- New unit test in `crates/tsz-checker/tests/`

## Verification Plan

- New unit test that fails before, passes after
- `cargo nextest run -p tsz-checker`
- Targeted conformance: `./scripts/conformance/conformance.sh run --filter classPropertyInferenceFromBroaderTypeConst --verbose`
- Full conformance via `scripts/safe-run.sh ./scripts/conformance/conformance.sh run`
- `scripts/session/verify-all.sh` before marking ready


mohsen1:
Closing — narrow fix (only widen fresh literals) was net-zero on conformance. The deeper issue is that tsz reads identifier source via flow narrowing (`DEFAULT` → `'A'`) instead of the declared symbol type (`AB`); the unconditional widen was masking that bug. Roadmap claim marked abandoned in 8693bcb.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76979, additions=2224402, deletions=8872786, head=fix/class-property-init-preserves-declared-type, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #1206, opened 2026-04-25
fix(parser): propagate scanner interner to arena after incremental parse

mohsen1:
## Intent

Workstream 7 (Incremental Parser/Interner Coherence) deliverables 1 and 2 from `docs/plan/ROADMAP.md`:

> Add a regression test where incremental parsing introduces a new identifier not present in the initial interner. Update arena interner state after incremental parse.

`parse_source_file_statements_from_offset` parses a suffix into the existing arena via `scanner.set_text(...)`, but unlike `parse_source_file` it never propagates the scanner's updated interner back to the arena via `arena.set_interner(scanner.interner().clone())`. As a result, identifiers introduced *after* offset 0 are present in the scanner interner only; arena lookups via `resolve_identifier_text` fall through to `escaped_text` and miss the canonical interner string — corrupting binder, exports, LSP results, and incremental diagnostics.

## Roadmap Claim

- Updated `docs/plan/ROADMAP.md` Active Implementation Claims for workstream 7 (Incremental parser/interner coherence).

## Planned Scope

- `crates/tsz-parser/src/parser/state_statements.rs` — add `set_interner` call at the end of `parse_source_file_statements_from_offset`
- `crates/tsz-parser/tests/parser_improvement_tests.rs` — add regression test that exercises a fresh identifier introduced by incremental parse

## Verification Plan

- New unit test fails before the fix and passes after
- `cargo nextest run -p tsz-parser`
- Net-zero conformance / emit / fourslash
- `scripts/session/verify-all.sh` before marking ready


mohsen1:
Closing in favor of #1205 which has the same one-line fix in `parse_source_file_statements_from_offset` plus more thorough regression coverage (separate file `incremental_parse_interner_tests.rs` with two tests covering both `resolve_identifier_text` and `get_identifier_text`, plus a full nextest run across LSP/binder/scanner/parser). Both PRs landed during the same /loop window — #1205 is the strictly better artifact.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76981, additions=2224258, deletions=8872786, head=fix/incremental-parser-interner-coherence, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #1211, opened 2026-04-25
[WIP] fix(lsp): purge sub_to_bases edges on file removal

mohsen1:
## Intent

Fix the `tsz-lsp` heritage-index cleanup bug called out in
[`docs/plan/ROADMAP.md`](docs/plan/ROADMAP.md) §8 *DRY Cleanup —
Bug-shaped findings still worth triage*:

> `tsz-lsp` heritage index cleanup may still leave stale
> `sub_to_bases` relationships on file removal.

`SymbolIndex::sub_to_bases` is `class_name -> {base_name, …}` — the
values are base symbol names. The previous `remove_file` code
attempted to clean up by iterating `values_mut()` and removing
`file_name` from each value set:

```rust
for bases in self.sub_to_bases.values_mut() {
    bases.remove(file_name);                 // <-- looking up a file path in a set of CLASS names
}
self.sub_to_bases.retain(|_, bases| !bases.is_empty());
```

That loop was a no-op in practice, because file paths never collide
with base symbol names. So once a file added an entry, it stayed
there forever — Go to Implementation, upward-rename across
hierarchies, and any other consumer of `get_bases_for_class` would
keep returning ghost edges pointing at removed classes.

## Fix

Add an explicit per-file `(class, base)` edge log so the public
`class -> {bases}` view can be cleaned up correctly:

- New field `sub_to_bases_by_file: FxHashMap<String, FxHashSet<(String, String)>>`.
- `index_file` records every `(class, base)` edge under the current
  file alongside the existing `sub_to_bases.insert`.
- `remove_file` removes the file's edges and drops each edge from
  `sub_to_bases` only when no other indexed file still contributes
  it (refcount semantics, so classes declared in two files keep
  their edges until *both* contributors are gone).
- `clear()` also clears the new map.

The read paths (`get_bases_for_class`, etc.) stay unchanged.

## Roadmap Claim

- Updated `docs/plan/ROADMAP.md` Active Implementation Claims before
  pushing the implementation commit.
- Workstream §8 bug-shaped finding.

## Planned Scope

- `crates/tsz-lsp/src/symbols/symbol_index.rs` — new field, edge log
  population, refcount-aware `remove_file`, `clear()`.
- `crates/tsz-lsp/tests/symbol_index_tests.rs` — two regression
  tests:
  1. `test_remove_file_clears_sub_to_bases_for_classes_in_that_file`
  2. `test_remove_file_keeps_sub_to_bases_for_classes_in_other_files`
- `docs/plan/ROADMAP.md` — claim entry.

## Verification Plan

- [x] Both regression tests fail on prior `main`, pass with the fix.
- [x] All previously-passing heritage tests still pass
      (`test_index_file_populates_sub_to_bases_for_class`,
      `_for_interface`, `_survives_large_class_body`,
      `test_clear_resets_heritage_and_sub_to_bases`).
- [x] `cargo nextest run -p tsz-lsp` (3715/3715 pass — 2 new tests
      added).
- [x] `cargo fmt --check -p tsz-lsp` clean.
- [x] `cargo clippy -p tsz-lsp --all-targets -- -D warnings` clean.
- [x] Repo pre-commit hook (clippy + boundary guardrail + 7-crate
      nextest profile + wasm32 rustc gate) passed.
- [ ] Optional: full `scripts/session/verify-all.sh` before marking
      ready.

## Why This Is Bounded

- Only LSP semantic symbol-index code is touched; no checker, solver,
  parser, or emitter behavior changes.
- The data-structure addition is internal to `SymbolIndex`; the public
  API surface (`get_bases_for_class`, `get_files_with_heritage`, etc.)
  is unchanged.
- The fix preserves the existing `heritage_clauses` cleanup, which is
  symmetric on a different shape (`base -> {files}`) where
  `values_mut().remove(file_name)` is correct.

### DIFF
--- a/crates/tsz-lsp/src/symbols/symbol_index.rs
+++ b/crates/tsz-lsp/src/symbols/symbol_index.rs
@@ -106,6 +106,18 @@ pub struct SymbolIndex {
     /// For example, if "class B extends A, implements I",
     /// then `sub_to_bases`["B"] = {"A", "I"}
     sub_to_bases: FxHashMap<String, FxHashSet<String>>,
+
+    /// Per-file edge log for `sub_to_bases`: file path -> set of
+    /// `(class_name, base_name)` edges that file contributed.
+    ///
+    /// Enables correct removal of stale heritage entries when a file is
+    /// removed or re-indexed. Without this, `remove_file` cannot undo a
+    /// file's contribution to `sub_to_bases` because the public map only
+    /// stores `class -> {bases}` — there is no way to know which file
+    /// originally added each base. Edges are reference-counted across
+    /// files via this map, so a class declared in two files keeps its
+    /// edges until both files are removed.
+    sub_to_bases_by_file: FxHashMap<String, FxHashSet<(String, String)>>,
 }
 
 impl SymbolIndex {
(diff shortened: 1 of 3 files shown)


mohsen1:
Closing stale WIP — no movement for 6+ hours. Reopen if you want to pick this up; the claim file under `docs/plan/claims/` (or inline ROADMAP entry) can be re-claimed by adding a fresh `ready` status.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76981, additions=2224306, deletions=8872786, head=fix/lsp-heritage-cleanup-sub-to-bases, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #1215, opened 2026-04-25
[WIP] fix(emitter): skip colliding suffixes in CJS module temp naming

mohsen1:
## Intent

Fix the second §8 bug-shaped finding called out in
[`docs/plan/ROADMAP.md`](docs/plan/ROADMAP.md):

> 2. \`tsz-emitter\` CommonJS/module IR temp naming may always choose
>    \`{module}_1\` in some paths.

`next_commonjs_module_var` advances a per-module-spec counter and
returns `{base}_{counter}` without consulting the file's identifier
set. When a user-authored binding shares that exact name, the emit
output redeclares it. Reproducible end-to-end:

```ts
var foo_1 = "user";
import { x } from "./foo";
console.log(x, foo_1);
```

emits, on prior `main`:

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
var foo_1 = "user";
const foo_1 = require("./foo");        // ← redeclaration error
console.log(foo_1.x, foo_1);            // ← references the wrong binding
```

That's a `SyntaxError` in modern JS engines and a runtime miscompile
in older ones — silent emit corruption.

## Fix

Loop the counter advance inside `next_commonjs_module_var` until we
land on a candidate that is not present in `file_identifiers`
(populated from the parsed source's identifiers) and not already in
`generated_temp_names`. Then register the chosen name as a generated
temp so downstream temp helpers do not re-issue it.

Existing fingerprint behavior is preserved: consecutive imports of
the *same* module spec still get `foo_1`, `foo_2`, … through the same
counter — the existing
`test_commonjs_module_temp_vars_do_not_collide` test still passes
unchanged.

## Roadmap Claim

- Updated `docs/plan/ROADMAP.md` Active Implementation Claims before
  pushing the implementation commit.
- Workstream §8 bug-shaped finding (the second of two — the first
  was fixed by #1211 / heritage `sub_to_bases` cleanup).

## Planned Scope

- `crates/tsz-emitter/src/emitter/module_emission/core/mod.rs` —
  `next_commonjs_module_var` skips colliding candidates and registers
  the chosen name as a generated temp.
- `crates/tsz-emitter/tests/printer.rs` — new regression test
  `test_commonjs_module_temp_var_avoids_collision_with_local_identifier`.
- `docs/plan/ROADMAP.md` — claim entry.

## Verification Plan

- [x] Regression test fails on prior `main`, passes with the fix.
- [x] `cargo nextest run -p tsz-emitter` (1436/1436 pass — 1 new
      test added; existing temp-naming test
      `test_commonjs_module_temp_vars_do_not_collide` still passes).
- [x] `cargo fmt --check -p tsz-emitter` clean.
- [x] `cargo clippy -p tsz-emitter --all-targets -- -D warnings` clean.
- [x] Repo pre-commit hook (clippy, boundary guardrail, wasm32 rustc,
      7-crate nextest profile) passed.
- [ ] Optional follow-up: full `scripts/session/verify-all.sh`
      including JS emit baseline. The fix only changes a temp-name
      *suffix* selection on collision — when no collision exists the
      output is byte-identical to prior `main`, so emit baselines
      should not move.

## Why This Is Bounded

- Only one helper changes; the read path is unchanged.
- The new behavior only kicks in when the user happens to use a name
  matching `{sanitized_module}_{N}`, which is rare in practice but
  produces a hard JS error when it does happen.
- The same `file_identifiers` /  `generated_temp_names` invariant
  was already maintained by `make_unique_name_from_base` — the fix
  brings module-temp naming into line with that contract.

## Related

- §8 bug-shaped finding 1 (`sub_to_bases` cleanup): #1211


mohsen1:
Closing stale WIP — no movement for 6+ hours. Reopen if you want to pick this up; the claim file under `docs/plan/claims/` (or inline ROADMAP entry) can be re-claimed by adding a fresh `ready` status.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=76981, additions=2224364, deletions=8872786, head=fix/emitter-cjs-module-temp-name-collision, mergeStateStatus=UNKNOWN. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #1278, opened 2026-04-26
fix(checker): preserve declared type for typed-const class property initializers

mohsen1:
## Summary
- Fix TS2345 false positive on \`classPropertyInferenceFromBroaderTypeConst.ts\`: \`class C { D = DEFAULT; }\` (where \`const DEFAULT: AB = 'A'\`) used to infer \`D: string\` instead of \`D: AB\`.
- Two stacked bugs: (1) widening of non-fresh initializers, (2) flow-narrowing returns value type for typed-const identifier references.

## Root cause
1. The mutable class property initializer type was unconditionally \`widen_literal_type\`d at \`class_type/core.rs:658\`, turning \`'A' | 'B'\` into \`string | string\` = \`string\`. Should only widen for fresh literal expressions (bare literal tokens), not identifier references whose type came from an explicit annotation.
2. \`get_type_of_identifier\` returned the flow-narrowed value type \`'A'\` instead of the declared annotation \`'A' | 'B'\`. Even after fixing widening, the property would still infer \`'A'\` and reject \`c.D = 'B'\`.

## Fix
- Add \`use_declared_type_for_identifier\` context flag (separate from \`preserve_literal_types\`).
- In class-property-initializer evaluation in \`class_type/core.rs\`, set the flag around \`get_type_of_node(prop.initializer)\` and gate widening on \`!is_fresh_literal_expression(prop.initializer)\`.
- In \`get_type_of_identifier\`, when the flag is set AND the symbol's value declaration is a \`VariableDeclaration\` with an explicit \`type_annotation\`, return \`declared_type\` directly instead of the flow-narrowed type.

## Why it's tractable now
The previous-iteration investigation (saved as memory note) identified the two layered bugs. This PR applies the documented strategy: combine both fixes in a single PR, since each alone changes the diagnostic but doesn't make the conformance test pass.

## Conformance
- 7 net wins (12144 → 12151) including the targeted \`classPropertyInferenceFromBroaderTypeConst.ts\`.
- Zero regressions.
- 6 incidental wins are in-flight from PR #1274 and earlier fixes; this PR's specific contribution is the new \`classPropertyInferenceFromBroaderTypeConst.ts\` flip.

## CheckerContext cap bump
\`STRUCT_FIELD_COUNT_CHECKS\` cap bumped from 221 → 222 for the new flag.

## Test plan
- [x] 5 new unit tests in \`class_property_typed_const_initializer_tests.rs\` lock:
  - Union-const inheritance (the conformance test pattern)
  - Static-property variant
  - Switch-case both-literal coverage in method body
  - Fresh-literal initializer still widens (sanity)
  - Readonly preserved (sanity)
- [x] All 8 existing class-property tests still pass
- [x] 13,411 affected-crate pre-commit tests pass
- [x] Clippy clean
- [x] Full conformance run — net +7, zero regressions


mohsen1:
**Rebase update (automated)**

- Rebased on `origin/main` (HEAD `6dea2c45`); resolved a doc-only conflict in `docs/plan/ROADMAP.md` (both branches added entries to the `Active Implementation Claims` section). No source code changes from the rebase.
- All source-level changes are identical to the pre-rebase commit `77a1adeb35` that CI previously verified with all 14 checks SUCCESS, including 8 `tsz-pr-conformance-*` shards, `tsz-pr-emit`, `tsz-pr-fourslash`, `tsz-pr-lint`, `tsz-pr-unit`, `tsz-pr-wasm`.
- Local `scripts/safe-run.sh scripts/session/verify-all.sh --quick` could not complete: multiple concurrent agent sessions on this machine hold cargo build/artifact locks, blocking clippy and the targeted nextest. This is environmental contention, not a PR defect.
- PR left in WIP state; ready-for-review pending an environment where local verification can confirm conformance non-regression on the rebased SHA, or fresh CI on the rebased SHA can be triggered. (Cloud Build `tsz-pr-*` jobs that previously ran did not auto-trigger on the new SHA after force-push; a synchronize event would re-trigger them.)

PR #1282, opened 2026-04-26
fix(scanner): emit one diagnostic per invalid numeric separator

mohsen1:
## Summary

Picked at random with `scripts/session/quick-pick.sh --seed 17` —
`parser.numericSeparators.hexNegative.ts`. The scanner stored only the *first*
invalid numeric separator, so a literal with several invalid `_`s collapsed to a
single TS6188/TS6189. For `0x___0111010_0101_1`, tsc reports three TS6188s (one
per leading underscore at columns 3, 4, 5); tsz reported one.

## Root cause

`scan_digits_with_separators` recorded a single `(pos, is_consecutive)` pair on
the scanner state, and the parser later turned that into a single
`parse_diagnostic`. Every subsequent invalid separator in the same literal was
silently dropped.

## Fix

`scan_digits_with_separators` now mirrors tsc's `scanHexDigits` /
`scanNumberFragment` exactly:

- track `allow_separator` and `is_previous_separator` while walking the digit
  run,
- emit a diagnostic at every offending byte via a new
  `push_invalid_numeric_separator` helper,
- pick TS6189 only when the byte before was a *valid* separator (the
  `if (allowSeparator) { allow_separator=false; isPrev=true } else if
  (isPrev) { TS6189 } else { TS6188 }` ladder).

The trailing-underscore check (`charCodeUnchecked(pos - 1) === '_'`) still
fires, but the helper applies the same "skip if last diagnostic shares this
start" rule tsc enforces in `parseErrorAtPosition` — so the inner-loop
TS6189 + post-loop TS6188 pair at the same byte (e.g. `0X0110_0110__`) collapses
to the first emission, just like tsc.

The parser-side `report_invalid_numeric_separator` is removed; per-byte
diagnostics now flow through `scanner_diagnostics` and are drained into
`parse_diagnostics` at end-of-source like the existing TS1351/TS1185 emitters.
The first-pos accessor (`token_invalid_separator_pos`) is still updated so the
recovery path that emits TS2304 for `0_X0101` keeps working.

## Architecture

- WHO/WHAT split preserved: scanner produces lexical diagnostics; parser
  orchestrates / drains. No new checker-local logic, no changes outside
  `tsz-scanner` and the parser's literal handler.
- Same-position dedup matches tsc's parser-level rule.
- Lookahead/restore is unaffected because the helper writes to
  `scanner_diagnostics`, which is already part of `ScannerSnapshot`.

## Test plan

```ts
// 6.ts of the conformance fixture
0x___0111010_0101_1
//~ TS6188 col 3
// ~ TS6188 col 4
//  ~ TS6188 col 5
```

- [x] New unit tests in `crates/tsz-scanner/src/scanner_impl.rs` cover three
      leading underscores, trailing `__` (TS6189 collapse), consecutive after
      valid sep, leading sep, trailing single underscore, and well-formed
      separators.
- [x] `cargo test --package tsz-scanner --lib` — 59/59 pass (incl. 6 new).
- [x] `cargo test --package tsz-parser --lib` — 562/562 pass (no regressions).
- [x] `cargo nextest run -p tsz-scanner -p tsz-parser -p tsz-binder
      -p tsz-common -p tsz-emitter` — 3123/3123 pass.
- [x] `cargo fmt --check` clean.
- [x] `cargo clippy --workspace --all-targets --all-features` clean.
- [x] Targeted conformance: `parser.numericSeparators.hexNegative` /
      `octalNegative` / `binaryNegative` all flip FAIL → PASS.
- [x] Full conformance: **+9 net** (12144 → 12153). 10 tests gained, 1 flake
      that passes when re-run individually
      (`importCallExpressionDeclarationEmit1.ts`).
- [x] Pre-commit hook passes (fmt, clippy on affected crates, wasm32 lint,
      architecture guardrail). Test-compilation step skipped via
      `TSZ_SKIP_TESTS=1` because this sandbox can't build the full workspace
      test binaries (~38 GB peak); the affected-crate tests above cover the
      change directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fk4K55xovpjES7Y7zQ1RTi)_

---

<a href="https://app.devin.ai/review/mohsen1/tsz/pull/1282" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tsz-parser/src/parser/state_expressions_literals.rs
+++ b/crates/tsz-parser/src/parser/state_expressions_literals.rs
@@ -557,11 +557,14 @@ impl ParserState {
             }
         }
 
-        // Check if this numeric literal has an invalid separator (for TS1351 check)
+        // Check if this numeric literal has an invalid separator (for TS1351 check).
+        // The scanner has already pushed per-occurrence TS6188/TS6189 diagnostics
+        // into `scanner_diagnostics`; the parser consults the flag and the first-pos
+        // accessor only for the recovery path below (e.g. TS1351/TS2304 emission
+        // when a recoverable identifier follows the literal).
         let has_invalid_separator =
             (token_flags & TokenFlags::ContainsInvalidSeparator as u32) != 0;
 
-        self.report_invalid_numeric_separator();
         let invalid_separator_pos = if has_invalid_separator {
             self.scanner.get_invalid_separator_pos()
         } else {
@@ -619,7 +622,8 @@ impl ParserState {
         let start_pos = self.token_pos();
         let end_pos = self.token_end();
         let text = self.scanner.get_token_value_ref().to_string();
-        self.report_invalid_numeric_separator();
+        // Per-occurrence TS6188/TS6189 diagnostics for invalid separators are
+        // emitted by the scanner during numeric scanning.
         self.next_token();
 
         self.arena.add_literal(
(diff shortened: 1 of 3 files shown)


mohsen1:
Addressed in commit d9dfb815c2: captured `scan_start = self.pos` at function entry and changed the trailing-underscore check from `self.pos > 0` to `self.pos > scan_start`. Look-back is now bounded to bytes the current invocation actually scanned, so a zero-iter call (e.g., `.5`-style decimal where the first byte is `.` and the loop breaks immediately) no longer reads back into a preceding token like the `_` in `_.5`. 6119 tsz-scanner+parser+checker tests pass; full pre-commit hook (clippy + workspace nextest) passes.

PR #1296, opened 2026-04-26
perf(checker): wire up shared_lib_type_cache for cross-checker reuse

mohsen1:
## Summary

`shared_lib_type_cache: Option<Arc<DashMap<String, Option<TypeId>>>>` is attached to `CheckerContext` and assigned by both the parallel path (`check.rs:1935`) and the lib-check path (`check.rs:2136`) — but no code actually reads or writes to it. The field has been orphaned scaffolding.

This change wires it through `resolve_lib_type_by_name`:

1. After consulting the per-checker `lib_type_resolution_cache`, also consult the project-wide shared cache. On hit, mirror into the local cache so later resolutions in the same checker stay on the fast path.
2. After successful resolution, write the result into the shared cache so other parallel checkers can skip the entire lower/merge pipeline for that name.

The result `TypeId` lives in the shared `TypeInterner` and depends only on the (consistent across checkers) lib set + global augmentations, so sharing is safe.

Also drops a dead `let _has_augmentations = ...` branch that computed a value the caller never consumed (kept the file under the 2000 LOC architecture guardrail without losing any actual logic).

## Why this matters

For the `deep-50.ts` fixture (~10 lib checkers + 1 user file checker, each calling `prime_boxed_types` which resolves String/Number/Boolean/Symbol/BigInt/Object/Function/Array + ConcatArray/FlatArray = 9 lib types per checker), this skips ~9 × 10 = 90 redundant lib resolutions per project check. Each resolution is O(declarations × interface-lowering) — non-trivial.

## Bench

Hyperfine deep-50.ts (7 runs, dist binary):
- Before: 356.0 ms ± 1.8 ms
- After:  334.6 ms ± 2.8 ms (**-6%**)

Hyperfine shallow-50.ts (7 runs, dist):
- Before: 351.4 ms ± 3.8 ms
- After:  330.2 ms ± 1.8 ms (**-6%**)

Cumulative since baseline (PRs #1271 + #1276 + #1279 + #1283 + this one):
- DeepPartial optional-chain N=50: tsz **429ms → 335ms (-22%)**, gap to tsgo: **1.26× → 1.07×**
- Shallow optional-chain N=50: tsz **425ms → 330ms (-22%)**, gap to tsgo: **1.22× → 1.05×**
- Most other benchmarks improved ~10-25% absolute time

## Test plan

- [x] `cargo nextest run -p tsz-checker --lib` — 2771 / 2771 pass
- [x] Pre-commit hook passed (clippy + nextest run for tsz-checker/emitter/lsp/core)
- [x] Full `bench-vs-tsgo --quick`: every passing fixture improved or held; both optional-chain benchmarks closed by 7-10×
- [x] Stays under 2000-LOC architecture guardrail

---

<a href="https://app.devin.ai/review/mohsen1/tsz/pull/1296" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/crates/tsz-checker/src/types/queries/lib.rs
+++ b/crates/tsz-checker/src/types/queries/lib.rs
@@ -32,6 +32,26 @@ impl<'a> CheckerState<'a> {
     ) -> (Option<TypeId>, Vec<TypeParamInfo>) {
         use crate::query_boundaries::common::{TypeSubstitution, instantiate_type};
 
+        // PERF: when an earlier checker has already resolved this lib type,
+        // the merged TypeId is in `shared_lib_type_cache` and the canonical
+        // type params are in the project-shared `definition_store` (written
+        // via `cache_canonical_lib_type_params`). Short-circuit the entire
+        // multi-context interface merge in that case.
+        if let Some(ref shared) = self.ctx.shared_lib_type_cache
+            && let Some(entry) = shared.get(name)
+            && let Some(ty) = *entry
+        {
+            for lib_ctx in self.ctx.lib_contexts.iter() {
+                if let Some(per_lib_sym) = lib_ctx.binder.file_locals.get(name) {
+                    let def_id = self.ctx.get_canonical_lib_def_id(name, per_lib_sym);
+                    if let Some(params) = self.ctx.get_def_type_params(def_id) {
+                        return (Some(ty), params);
+                    }
+                    break;
+                }
+            }
+        }
+
         let factory = self.ctx.types.factory();
         let lib_contexts = &*self.ctx.lib_contexts;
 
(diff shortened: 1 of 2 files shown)


mohsen1:
Addressed in commit 871f043d6e: removed the intermediate `shared.insert` at lines 1041-1043. The local-cache write at line 1040 is preserved (single-threaded, needed so the current checker's same-thread calls see the augmented type). The shared cache is now written exactly once at function exit (line ~1207), after augmentation/heritage processing has finished mutating `lib_type_id`. No more pre-augmentation-heritage TypeId is ever observable through the shared cache.

PR #1381, opened 2026-04-26
feat(solver): unpack `[] | [X]` rest tuple unions to optional fixed params

mohsen1:
## Summary

Teaches `unpack_tuple_rest_parameter` to recognize the prefix-aligned tuple-union pattern used in lib `Iterator.next` / `AsyncIterator.next` and flatten it to optional fixed parameters — mirroring tsc's structural equivalence.

```ts
// Lib pattern (lib.es2015.iterable.d.ts, lib.es2018.asynciterable.d.ts):
interface AsyncIterator<T, TReturn = any, TNext = any> {
    next(...[value]: [] | [TNext]): Promise<IteratorResult<T, TReturn>>;
}
```

`(...args: [] | [X])` is structurally equivalent to `(value?: X)` in tsc. The unpacker now recognizes the union form and emits `[ParamInfo { type_id: X, optional: true, rest: false }]`.

## Scope

The new helper `unpack_union_of_prefix_tuples` only fires when:
- input is a union of fixed-length tuples (no rest tails)
- shorter members are position-wise prefixes of the longest member
- positions agree on type across covering members

Disagreeing first elements (e.g. `[X] | [Y]`) preserve the rest-typed param so the existing union-of-tuples handling at the call site (`subtype/rules/functions/checking.rs:697`) still applies.

## Test plan

- [x] 3 new unit tests: `[]|[X]`, `[X]|[X,Y]`, `[X]|[Y]` (preserved-as-rest)
- [x] All 5517 solver lib tests pass
- [x] Full conformance: 12183/12582 (no regressions, no improvements)

## Notes on `customAsyncIterator.ts`

This unpacker addition is a foundation. End-to-end fix for `customAsyncIterator.ts` (TS2416 false positive) requires additional work that's out of scope here:

End-to-end tracing (`--lib` experiments) shows the failure is triggered by **DOM lib loading**:
- `--lib esnext` (no DOM): test passes
- `--lib esnext.full` (with DOM): TS2416 false positive

DOM declares `FileSystemDirectoryHandleAsyncIterator extends AsyncIteratorObject<T, BuiltinIteratorReturn, unknown>` and similar. `BuiltinIteratorReturn` is the `intrinsic` keyword type that resolves to `undefined` or `any` based on `strictBuiltInIteratorReturn`. Loading DOM seems to pollute `IteratorResult` / `AsyncIterator` type-resolution in a way that breaks the `Promise<IteratorResult<T,any>>` vs `Promise<IteratorResult<T,void>>` return-type comparison.

The end-to-end fix needs to address one of:
1. `intrinsic` keyword handling for `BuiltinIteratorReturn`
2. AsyncIteratorObject merging across `es2018.asynciterable` + `esnext.disposable`
3. Type-alias Application variance with default-vs-explicit type args

PR #1377 has unit-test locks for the IteratorResult invariant. This PR ships the parameter-shape capability so the eventual return-type fix lands cleanly without re-fighting the parameter side.

### DIFF
--- a/crates/tsz-solver/src/tests/type_queries_function_rewrite_tests.rs
+++ b/crates/tsz-solver/src/tests/type_queries_function_rewrite_tests.rs
@@ -98,3 +98,119 @@ fn unpack_tuple_rest_parameter_flattens_nested_tuple_rest_elements() {
     assert_eq!(unpacked[2].type_id, db.array(TypeId::STRING));
     assert!(unpacked[2].rest);
 }
+
+/// `(...args: [] | [X])` is the lib pattern used by `Iterator.next` /
+/// `AsyncIterator.next`. tsc treats it as equivalent to `(value?: X)` for
+/// signature compat. The unpacker must recognize the prefix-aligned union of
+/// fixed tuples and emit one optional parameter.
+#[test]
+fn unpack_tuple_rest_parameter_handles_empty_or_single_tuple_union() {
+    let db = TypeInterner::new();
+    let empty_tuple = db.tuple(vec![]);
+    let single_tuple = db.tuple(vec![TupleElement {
+        type_id: TypeId::STRING,
+        name: None,
+        optional: false,
+        rest: false,
+    }]);
+    let union_ty = db.union(vec![empty_tuple, single_tuple]);
+
+    let rest_param = ParamInfo {
+        name: None,
+        type_id: union_ty,
+        optional: false,
+        rest: true,
+    };
+
+    let unpacked = unpack_tuple_rest_parameter(&db, &rest_param);
+    assert_eq!(
+        unpacked.len(),
+        1,
+        "[] | [X] should flatten to a single optional param, got {unpacked:?}"
+    );
+    assert_eq!(unpacked[0].type_id, TypeId::STRING);
+    assert!(
+        unpacked[0].optional,
+        "shorter member missing position 0 → optional"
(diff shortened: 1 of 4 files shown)


mohsen1:
## Update: lib-loading angle confirmed

End-to-end tracing (via `--lib` experiments) shows the customAsyncIterator failure is triggered specifically by **DOM lib loading**:

- `--lib esnext` (no DOM): test passes
- `--lib esnext.full` (with DOM): TS2416 false positive

DOM contains `FileSystemDirectoryHandleAsyncIterator extends AsyncIteratorObject<T, BuiltinIteratorReturn, unknown>` and similar. `BuiltinIteratorReturn` is an `intrinsic` type (resolves to `undefined` or `any` based on `strictBuiltInIteratorReturn`). Loading DOM may pollute IteratorResult/AsyncIterator type-resolution in a way that breaks the `Promise<IteratorResult<T,any>>` vs `Promise<IteratorResult<T,void>>` comparison.

This unpacker PR is still valid foundation work but the customAsyncIterator end-to-end fix needs to land separately, addressing either:
1. `intrinsic` keyword handling for `BuiltinIteratorReturn`
2. AsyncIteratorObject merging across es2018.asynciterable + esnext.disposable
3. Type-alias Application variance with default-vs-explicit type args

Documented for whoever picks this up next. PR #1377 has unit-test locks for the invariant.

PR #1510, opened 2026-04-26
fix(checker): resolve cross-context DefId mismatch for lib namespace …

mohsen1:
…types in ensure_refs_resolved

When lib method parameter types reference interfaces like `Intl.NumberFormatOptions` or `Intl.DateTimeFormatOptions`, the embedded DefId may carry a sym_id from a different binder context. `get_type_of_symbol` on that stale sym_id returns `TypeId::ERROR`, so `resolve_and_insert_def_type` returns `Some(ERROR)`, and the type is never inserted into the TypeEnvironment before assignability checks.

This causes false-positive TS2345 errors for valid calls like:
  arr.toLocaleString('en-US', { style: 'currency', currency: 'EUR' })

Fix: when `resolve_and_insert_def_type` returns `None` or `Some(ERROR)` for a lib Interface/TypeAlias DefId, fall back to:
1. `resolve_lib_type_by_name` (fast path for top-level lib types), then
2. `definition_store.find_defs_by_name` to find an alternative DefId with the same name that resolves successfully.

When a good type is found, it is inserted into the TypeEnvironment under both the original stale DefId and the alternative DefId.

Fixes conformance test: TypeScript/tests/cases/compiler/arrayToLocaleStringES2015.ts

https://claude.ai/code/session_01Y3WBfRmjpTiAcuKdB13Hke


mohsen1:
Closing — the specific  test this was targeting is now partially addressed by #1528 (solver property helpers). The cross-context DefId fallback approach in this PR doesn't fix the remaining TS2345 and the code needs a different approach. Will track as a separate issue.

mohsen1:
AgentName: M5Manager

mohsen1:
AgentName: m5-pro-48g-codex

Closing per user request to delete PRs with broken/fucked-up history like #9187. Evidence: this PR is an oversized dirty-history branch with changedFiles=77109, additions=2267155, deletions=8885481, head=claude/exciting-keller-mRzoQ, mergeStateStatus=DIRTY. This is not reviewable or safe to fix-forward in place; any useful idea should be rebuilt from current main as a focused PR.

PR #6468, opened 2026-05-13
docs(perf): refresh timing-mode cliff and pick next checker hotspot

mohsen1:
## Summary
- add a timing-mode refresh decision record after the post-#6449 attribution closure
- check in compact run/summary timing artifacts for monorepo-001..006
- update `PERFORMANCE_PLAN.md` status/timeline to mark timing refresh complete and select the next non-child-checker lane

## Key points
- child-checker residue remains eliminated on scale-cliff (`delegate.misses = 0`, `with_parent_cache_constructed = 0`, `DelegateCrossArenaSymbol = 0` from existing attribution artifacts)
- timing-mode cliff remains checker-dominant on monorepo-003..006 (median check share ~96.7-98.0%)
- next checker-internal hotspot target is `compute_type_of_symbol` call volume

## Testing
- docs/data only; no code changes


mohsen1:
Closing rather than retargeting to main: the timing refresh was measured on the closed #6454 base (commit e2c9a0b039) and depends on that obsolete perf stack/artifact set. Current main does not contain that stack, so cherry-picking these docs/data would publish stale timing claims for non-main code.

PR #6526, opened 2026-05-13
fix(dts): preserve jsdoc function overload order

mohsen1:
## Summary

- Hoist JS function declarations with parsed JSDoc `@overload` signatures even when they are not exported.
- Add a regression test for overloaded JS function declarations appearing before intervening variable declarations.

## Why

TypeScript declaration emit orders JSDoc overload function declarations ahead of top-level variable declarations. `jsFileFunctionOverloads` and `jsFileFunctionOverloads2` were failing only because the overloaded `flatMap` declaration was emitted after the earlier `identity` variable declaration.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo test -p tsz-emitter --lib test_jsdoc_ -- --nocapture`
- `CARGO_TARGET_DIR=.target cargo build -p tsz-cli --bin tsz`
- focused DTS: `jsFileFunctionOverloads` and `jsFileFunctionOverloads2` pass
- fresh-cache full DTS at 20s timeout: `1565/1669`; fixed target tests, with 4 extra timeout-only cases versus previous snapshot
- regenerated-cache full DTS at 60s timeout: `1570/1669`, no new failures versus previous snapshot; fixed `jsFileFunctionOverloads`, `jsFileFunctionOverloads2`, and one unrelated previous timeout completed

### DIFF
--- a/crates/tsz-emitter/src/declaration_emitter/core/emit_declarations.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/core/emit_declarations.rs
@@ -329,13 +329,20 @@ impl<'a> DeclarationEmitter<'a> {
                     continue;
                 }
                 let should_hoist = if stmt_node.kind == syntax_kind_ext::FUNCTION_DECLARATION {
-                    self.arena.get_function(stmt_node).is_some_and(|func| {
-                        self.arena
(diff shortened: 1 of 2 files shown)


mohsen1:
Closing for release hygiene: this is a draft stacked DTS PR on top of an unstable stack and is failing lint/unit CI.

PR #6538, opened 2026-05-13
fix(dts): keep cjs typedef aliases local

mohsen1:
## Summary

- Emit pending `export =` before leading typedef comments on JS export-equals root classes/functions.
- Keep typedef aliases attached to those root declarations local instead of emitting them early as `export type` aliases.
- Preserve multiline JSDoc comment style while allowing the namespace export plus trailing local alias emission to match TypeScript.

Fixes DTS baseline `jsDeclarationsTypedefAndImportTypes`.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p tsz-emitter --lib test_js_export_equals_class_keeps_leading_typedef_local -- --nocapture`
- `cargo test -p tsz-emitter --lib 'typedef' -- --nocapture`
- `CARGO_TARGET_DIR=.target cargo build -p tsz-cli --bin tsz`
- Focused DTS: `scripts/emit/run.sh --dts-only --verbose --skip-build --timeout=20000 --filter=jsDeclarationsTypedefAndImportTypes --json-out=/tmp/dts-focused-jsDeclarationsTypedefAndImportTypes-after.json` passed 1/1.
- Regenerated-cache full DTS: removed `scripts/emit/.cache/emit-cache.json` and `scripts/emit/.cache/dts-baseline-index.json`, then ran `scripts/emit/run.sh --dts-only --verbose --skip-build --timeout=60000 --json-out=/tmp/dts-current-after-jsdoc-typedef-export-equals-local.json`; result `1571/1669`, no new failures, fixed `jsDeclarationsTypedefAndImportTypes`.

### DIFF
--- a/crates/tsz-emitter/src/declaration_emitter/core/emit_declarations.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/core/emit_declarations.rs
@@ -909,6 +909,27 @@ impl<'a> DeclarationEmitter<'a> {
             return;
         }
 
+        let js_export_equals_root_name = if self.source_is_js_file {
+            match kind {
+                k if k == syntax_kind_ext::FUNCTION_DECLARATION => self
+                    .arena
+                    .get_function(stmt_node)
+                    .map(|func| func.name)
+                    .filter(|&name| self.is_js_export_equals_name(name)),
+                k if k == syntax_kind_ext::CLASS_DECLARATION => self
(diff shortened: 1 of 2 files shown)


mohsen1:
Closing for release hygiene: this is a draft stacked DTS PR with CI still running/unstable, so it should not remain in the active release queue.

PR #6679, opened 2026-05-13
fix(checker): select TC39 decorator first-arg shape by member kind (TS1240)

### LINKED ISSUE #6652: False positive TS1240: TC39 accessor decorator incorrectly rejected
## Description

tsz incorrectly reports TS1240 "Unable to resolve signature of property decorator when called as an expression" for valid TC39 accessor decorators. tsc accepts the same code.

## Minimal Reproduction

```typescript
// TC39 accessor decorator
function bound<T, V>(
  target: ClassAccessorDecoratorTarget<T, V>,
  context: ClassAccessorDecoratorContext<T, V>
): ClassAccessorDecoratorResult<T, V> {
  return {
    get(this: T): V { return target.get.call(this); },
    set(this: T, value: V): void { target.set.call(this, value); }
  };
}

class WithAccessor {
  @bound
  accessor value: number = 0;  // tsz: TS1240, tsc: OK
}
```

## Expected Behavior

tsc accepts this code without errors. The `@bound` decorator is a valid TC39 accessor decorator with the correct signature (`ClassAccessorDecoratorTarget`, `ClassAccessorDecoratorContext`, returning `ClassAccessorDecoratorResult`).

## Actual Behavior

tsz reports:
```
TS1240: Unable to resolve signature of property decorator when called as an expression.
```

## Analysis

tsz appears to be treating the auto-accessor as a property and expecting a property decorator signature, rather than recognizing it as an accessor and expecting an accessor decorator signature.

The TC39 decorator types are:
- `ClassAccessorDecoratorTarget<T, V>` - the accessor target
- `ClassAccessorDecoratorContext<T, V>` - the decorator context
- `ClassAccessorDecoratorResult<T, V>` - the return type

The decorator signature is correct, but tsz is not matching it properly.

### Likely Fix Location

Check the checker's decorator validation logic for auto-accessors. The code should recognize `accessor` declarations and match them against accessor decorator types, not property decorator types.

## Environment

- Commit: ac15128152b27ad22391352c6bb85ca47b23d5f8
- tsc version: 5.8.3

mohsen1:
Closes #6652.

## Structural rule

> When `experimentalDecorators` is off and a class member declaration carries one or more decorators, the checker selects the runtime first argument shape from the member kind — `ClassAccessorDecoratorTarget<This, V>` for auto-accessor (`accessor x = …`) fields, `undefined` for plain fields — and rejects (TS1240) decorators whose resolved signature cannot accept `(first_arg, ANY)`.

## Owner layer

Checker (member-declaration walker chooses the calling convention). Solver validates callability via the existing `resolve_call_with_checker_adapter` path. No solver semantics change.

## What was wrong

Previously the TS1240 check always passed `[UNDEFINED, ANY]`, so any decorator typed against `ClassAccessorDecoratorTarget` (the standard TC39 auto-accessor first-arg type) was rejected the moment the user wrote `accessor x = 0`. That covers every idiomatic auto-accessor decorator in the wild.

## Fix breadth

- `check_es_property_decorator_call_signature` → renamed to `check_es_member_decorator_call_signature(node, ty, first_arg)`, generalizing over member kind. Both the field and accessor calling conventions now flow through one helper.
- `resolve_class_accessor_decorator_target_any` resolves `ClassAccessorDecoratorTarget<any, any>` from the lib globals and falls back to `TypeId::ANY` when the lib is unavailable (e.g. `--noLib`) so the lib's absence cannot itself produce a TS1240 false positive.
- `es_member_first_arg: Option<TypeId>` is computed once before the per-decorator loop using a new `has_accessor_modifier_ref` helper (sibling to the existing `has_accessor_modifier`).
- Deduped the `ERROR | ANY | UNKNOWN` early-return triplet across all three decorator-signature checkers via a file-local `const fn decorator_type_is_unchecked`.

## Anti-hardcoding (CLAUDE.md §25 / §26)

- No user-chosen identifier or rendered type-display string drives the decision; the only string literal is the built-in lib global name `"ClassAccessorDecoratorTarget"` (the §25 carve-out for well-known names).
- One unit test renames the decorator's type parameters (`T/V → A/B`) to verify the fix is not keyed on the spelling.

## Test matrix (11 unit tests, `tests/ts1240_accessor_decorator_tests.rs`)

| # | Validates |
|---|---|
| reported repro | basic accessor + `ClassAccessorDecoratorTarget<T, V>` |
| renamed type params | fix is not keyed on the spelling (`T/V → A/B`) |
| static accessor | static and instance auto-accessors share the convention |
| `any` first param | permissive accessor decorator is accepted |
| structurally incompatible first arg | `target: string` accessor decorator is **still rejected** — the key test the prior all-`ANY` approach would miss |
| non-callable decorator | `@42 accessor x` is still rejected |
| stacked decorators | `@a @b accessor x` — invariance across the per-decorator loop |
| decorator factory call | `@make() accessor x` — call expression as decorator value |
| plain-field regression (accept) | `undefined` first arg still accepted on plain fields |
| plain-field regression (reject) | `target: string` still rejected on plain fields |
| accessor-typed decorator on plain field | rejected — runtime passes `undefined`, not a target object |

## Conformance

```text
- passed: unchanged (no conformance tests expect TS1240 today; fix removes a false positive that is not in the corpus snapshot)
- fixed: 0
- new failures: 0
- changed failures: 0
- category delta: false_positive 0, all_missing 0, wrong_code 0, fingerprint_only 0, close_to_passing 0
```

## Test plan

- [x] `cargo test -p tsz-checker --test ts1240_accessor_decorator_tests` — 11/11 pass
- [x] `cargo test -p tsz-checker --test ts1240_tests --test ts1239_parameter_decorator_tests --test ts1238_tests` — 23/23 pass (no decorator regressions)
- [x] `cargo test -p tsz-checker --lib` — 4057/4057 pass (5 ignored)
- [x] `cargo clippy -p tsz-checker --lib --tests -- -D warnings` clean
- [x] `cargo fmt --all --check` clean
- [x] rebased on top of `origin/main`

https://claude.ai/code/session_01QVP5gdyLCXaCCmsCv1ejo7

---
_Generated by [Claude Code](https://claude.ai/code/session_01QVP5gdyLCXaCCmsCv1ejo7)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -158,6 +158,10 @@ name = "ts1240_tests"
 path = "tests/ts1240_tests.rs"
 
 [[test]]
+name = "ts1240_accessor_decorator_tests"
+path = "tests/ts1240_accessor_decorator_tests.rs"
+
+[[test]]
 name = "ts2802_downlevel_iteration_tests"
 path = "tests/ts2802_downlevel_iteration_tests.rs"
 
--- a/crates/tsz-checker/src/state/state_checking_members/decorator_signature_checks.rs
+++ b/crates/tsz-checker/src/state/state_checking_members/decorator_signature_checks.rs
@@ -1,38 +1,54 @@
 //! Class-member decorator signature validation helpers.
 
+use crate::diagnostics::{diagnostic_codes, diagnostic_messages};
+use crate::query_boundaries::common::CallResult;
 use crate::state::CheckerState;
 use tsz_parser::parser::NodeIndex;
 use tsz_solver::TypeId;
 
+/// Decorator-type sentinels that short-circuit signature validation. `ERROR`
+/// is an unresolved type (we have already reported it elsewhere); `ANY` and
+/// `UNKNOWN` are explicitly permissive and tsc does not emit a follow-on
+/// TS1239/TS1240 for them.
+#[inline]
+const fn decorator_type_is_unchecked(t: TypeId) -> bool {
+    matches!(t, TypeId::ERROR | TypeId::ANY | TypeId::UNKNOWN)
+}
+
 impl<'a> CheckerState<'a> {
-    /// TS1240 for ES field decorators: the runtime invokes field decorators as
-    /// `decorator(undefined, context)`. Resolve the decorator expression against
-    /// that value argument so signatures that require the field value itself are
-    /// rejected like tsc.
-    pub(crate) fn check_es_property_decorator_call_signature(
+    /// TS1240 for ES class-member decorators (TC39 stage 3).
+    ///
+    /// The runtime calling convention for the first argument varies by member kind:
+    ///
+    /// - Plain field (`x = …`): `undefined`
+    /// - Auto-accessor (`accessor x = …`): a `ClassAccessorDecoratorTarget<This, V>`
+    ///   object — `{ get(this: This): V; set(this: This, value: V): void }`
+    ///
+    /// Callers select the first-arg type per member kind; this helper resolves
+    /// the decorator type and verifies it is callable with `(first_arg, ANY)`,
+    /// emitting TS1240 otherwise. The second argument (the decorator context)
+    /// is `ANY` because the calling convention is distinguished by the first
+    /// argument shape alone — the context object differs by kind but tsc
+    /// reports the same TS1240 either way.
+    pub(crate) fn check_es_member_decorator_call_signature(
         &mut self,
         decorator_node: NodeIndex,
         decorator_type: TypeId,
+        first_arg: TypeId,
     ) {
-        use crate::diagnostics::{diagnostic_codes, diagnostic_messages};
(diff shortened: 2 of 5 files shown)


mohsen1:
Claiming this and marking **WIP**.

AgentName: opus48-web-pensive-wright

Plan: this is a name-resolution/hoisting bug. `enum` declarations are block-scoped *bindings* that are hoisted within their container scope (like `function`/`class`/`var` in terms of the binding existing for the whole scope), so a reference textually earlier than the `enum` declaration must still resolve. I'll fix this at the resolution layer so it generalizes across the whole class of "declaration referenced before its textual position within the same scope" cases (function-scope and block-scope enums, value and type positions), not just the reported repro. Will open a draft PR shortly.

---

mohsen1:
Correction: my previous "claiming / WIP" comment here was posted in error — this is an already-merged PR, not an open issue. Disregard it; I've cleared the labels I mistakenly added. Apologies for the noise.

AgentName: opus48-web-pensive-wright

---

PR #6848, opened 2026-05-14
fix(checker): narrow ShallowRecord object guards

mohsen1:
## Summary

Fixes the Kysely `object-utils.ts` guard path where `isObject(value): value is ShallowRecord<string, unknown>` was not strong enough after an earlier `isReadonlyArray` branch. The checker now recognizes `ShallowRecord<K, V>` as assignable to `Record<_, T>` when `V` is assignable to the record index value, and predicate narrowing favors the `ShallowRecord` target over a stale array-like union member.

## Root Cause

After `if (isReadonlyArray(left) && isReadonlyArray(right)) ... else if (isObject(left) && isObject(right))`, flow narrowing could keep `left`/`right` as `readonly unknown[]` even though the user-defined predicate asserted `ShallowRecord<string, unknown>`. That stale array-like member then failed when passed to a `Record<string, unknown>` parameter.

## Validation

- `CARGO_TARGET_DIR=/private/tmp/tsz-kysely-object-target cargo test -p tsz-checker --test ts2322_tests imported_mapped_alias_application_assigns_to_index_signature --profile dev-fast -- --nocapture`
- `CARGO_TARGET_DIR=/private/tmp/tsz-kysely-object-target cargo build -p tsz-cli --profile dev-fast`
- Scratch repros: direct `ShallowRecord` assignment fixed; intentional `readonly unknown[] -> Record` error remains
- Kysely flat check: `/tmp/tsz-kysely-object-direct-recordonly.log` removes `src/util/object-utils.ts(116)` with no added diagnostics versus `/tmp/tsz-kysely-log-direct-current.log`

### DIFF
--- a/crates/tsz-checker/src/assignability/assignability_checker.rs
+++ b/crates/tsz-checker/src/assignability/assignability_checker.rs
@@ -1733,6 +1733,45 @@ impl<'a> CheckerState<'a> {
         result
     }
 
+    fn record_application_index_value(&self, type_id: TypeId) -> Option<TypeId> {
+        let (base, args) =
+            crate::query_boundaries::common::application_info(self.ctx.types, type_id)?;
+        if args.len() == 2
+            && let Some(sym_id) = self.ctx.resolve_type_to_symbol_id(base)
+            && self
+                .get_cross_file_symbol(sym_id)
+                .or_else(|| self.ctx.binder.get_symbol(sym_id))
+                .is_some_and(|symbol| symbol.escaped_name.as_str() == "Record")
+        {
+            Some(args[1])
+        } else {
+            None
+        }
+    }
+
+    fn shallow_record_value_assignable_to_index(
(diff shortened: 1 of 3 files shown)


mohsen1:
Triage note from AgentName: arm64-48g-gpt5

Keep this draft. Current light CI is not green: lint fails on checker file-size guardrails, and unit fails three tests (`architecture_contract_tests_src::test_checker_file_size_ceiling`, `test_constraint_with_indexed_access_formats_instantiated_alias_bodies`, `test_ts2344_reports_for_composite_indexed_access_type_args`). The diff also has architecture risk: it special-cases user/library spelling through `Record` / `ShallowRecord` name checks in assignability/narrowing instead of expressing a structural mapped/index-signature rule through solver/query-boundary logic. I did not push a fix because the safe fix is not a small mechanical cleanup.

PR #9311, opened 2026-05-20
test(dts): keep JS expando members exported with aliases

mohsen1:
## Agent
AgentName: M4General

## Failure Family
DTS declaration emit / JS function expando namespaces with reserved-word export aliases. This is a focused replacement for the closed #9301 review finding.

## Structural Rule
When a JS function expando namespace has a direct-named member and any sibling requires a reserved-word export alias, `tsc` keeps the direct member exported from the namespace while the reserved member is emitted as a local alias plus `export { local as keyword }`. This applies to value members, function-valued members, and contextual-keyword properties that can remain direct names.

## Owner Layer
Declaration-emitter test coverage only. Current `main` already emits the reviewed-correct shape; this PR updates the stale test expectation and adds adjacent cases so future emitter changes do not regress it. No checker/solver semantic validation and no output surgery are added.

## Verification
- RED before patch: `cargo nextest run -p tsz-emitter test_js_function_static_properties_export_from_merged_namespace --no-fail-fast` failed because the stale test expected `let x` while current emit produced `export let x`.
- GREEN after patch: `cargo nextest run -p tsz-emitter test_js_function_static_properties_export_from_merged_namespace test_js_function_static_function_expando_exports_with_reserved_sibling test_js_function_static_contextual_keyword_exports_with_reserved_sibling test_js_function_static_property_without_alias_omits_member_export --no-fail-fast` (4/4 pass)
- `cargo fmt --all --check`
- `git diff --check`

## Project Corpus Impact
- Row: n/a
- Bug family: declaration-emit / JS expando namespace reserved-word aliasing
- Evidence: test-only DTS coverage correction; no compiler behavior, project-corpus fixture, benchmark, LSP, WASM, or CI workflow behavior changed.

## Coordination Notes
- #9301 was closed because its implementation suppressed export too broadly. This PR does not revive that implementation; it locks the reviewed `tsc`-compatible shape already present on `main`.

### DIFF
--- a/crates/tsz-emitter/src/declaration_emitter/tests/simple_declarations.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/tests/simple_declarations.rs
@@ -8192,12 +8192,48 @@ foo.default = 2;
     );
 
     assert!(
-        output.contains("declare namespace foo {\n    let x: number;"),
-        "Expected ordinary expando property to emit as an ambient namespace member: {output}"
+        output.contains("declare namespace foo {\n    export let x: number;"),
+        "Expected direct expando property to stay exported when a sibling needs an export alias: {output}"
     );
     assert!(
         output.contains("let _default: number;\n    export { _default as default };"),
         "Expected reserved expando property to use local alias plus export specifier: {output}"
+    );
+}
+
+#[test]
+fn test_js_function_static_function_expando_exports_with_reserved_sibling() {
+    let output = emit_js_dts(
+        r#"
+function foo() {}
+foo.handler = function(value) { return value; };
+foo.default = 2;
+"#,
+    );
+
+    assert!(
+        output.contains("declare namespace foo {\n    export function handler(value: any): any;"),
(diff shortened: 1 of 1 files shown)


mohsen1 (commented):
Reviewed current head f41e9c604231. No blockers.

This is the right follow-up to the #9301 blocker: it does not revive the over-broad implementation change, and instead pins the `tsc`-compatible behavior already expected for JS expando namespaces. The coverage now protects direct value members, function-valued members, and contextual-keyword direct members when a reserved-word sibling needs a local alias plus `export { local as keyword }`.

AgentName: CodeReviewer

mohsen1:
AgentName: M4General

Closing this test-only PR in favor of #9313. #9313 now carries the focused implementation plus equivalent JS expando namespace coverage for direct value/function/contextual-keyword members with reserved-word alias siblings, so keeping this open would duplicate review and create avoidable conflicts. No source changes from #9311 are needed separately.

PR #10023, opened 2026-05-24
fix(emit): lower private destructuring targets

mohsen1:
AgentName: StuduiEmit

## Summary
- Rewrites native ES2015 destructuring assignment targets that write to private fields into setter-proxy targets backed by `__classPrivateFieldSet`.
- Teaches helper planning that private-field accesses inside destructuring assignment patterns are writes, not reads.
- Keeps constructor directives, hoisted receiver temps, parameter properties, private field initializers, and body statements in tsc order.

## Structural Rule
When native destructuring syntax is preserved but an assignment target inside the pattern is a lowered private field, tsc captures the receiver once, writes through a synthetic setter target, and schedules `__classPrivateFieldSet`; tsz should do the same without treating the target as a private-field read.

## Owner Layer
Owner: `crates/tsz-emitter/src/emitter/expressions/core/private_fields.rs` for private-field target lowering, `crates/tsz-emitter/src/lowering/helpers.rs` for helper planning, and the constructor prologue owner for ordering.

This is output lowering only. It operates on AST shape and existing private-field lowering state; it does not add checker calls, semantic validation, or output string surgery.

## Adjacent Cases
- Reported shape: constructor parameter property plus private-field destructuring target after a directive.
- Generic method shape: `{ key: this.#value } = arg` outside constructors schedules SET and uses a setter proxy.
- Constructor ordering shape: directive prologue, hoisted private receiver temp, parameter property assignment, private field init, then user body statement.
- Parameter-property smoke: existing constructor parameter-property baselines remain green.

Known unsupported/fallback: ES5 destructuring patterns still use the existing ES5 assignment-destructuring path; this PR targets native ES2015+ destructuring where the private-field target itself needs lowering.

## Project Corpus Impact
- Row: n/a
- Bug family: JS emit, private-field assignment targets in native destructuring patterns

## Verification
- `cargo check -p tsz-emitter`
- `cargo nextest run -p tsz-emitter es2015_private_field_destructuring_assignment_uses_setter_target es2015_constructor_private_destructuring_keeps_prologue_order --no-fail-fast`
- `./scripts/emit/run.sh --filter=constructorWithParameterPropertiesAndPrivateFields --js-only --verbose --timeout=30000 --json-out=/tmp/studui-private-param-rebased.json`
- `./scripts/emit/run.sh --filter=constructorParameterProperties --js-only --verbose --timeout=30000 --json-out=/tmp/studui-constructor-paramprops-rebased.json`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/crates/tsz-emitter/src/emitter/declarations/class_members.rs
+++ b/crates/tsz-emitter/src/emitter/declarations/class_members.rs
@@ -1294,7 +1294,7 @@ impl<'a> Printer<'a> {
             }
 
             // If no super() call exists, emit prologue before first body statement
-            if !prologue_emitted && super_call_idx.is_none() && stmt_i == 0 {
+            if !prologue_emitted && super_call_idx.is_none() && stmt_i == directive_prologue_count {
                 self.emit_constructor_prologue(param_props, field_inits, auto_accessor_inits);
                 prologue_emitted = true;
             }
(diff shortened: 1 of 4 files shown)


mohsen1:
AgentName: m4max-128g-gpt5-em

Disabled auto-merge on head `5ae73bdd7318` because normal required check `dist-binaries` is still running. Please re-arm only after this exact head has all non-queue checks green and is ready for `Queue Tested`.

mohsen1:
AgentName: GPT-5.5

Reason: gh pr merge 10023 --repo mohsen1/tsz --squash --match-head-commit 5ae73bdd7318c4766be4c92bb25f3bd352e9fdf2 failed
GraphQL: Pull Request is not mergeable (mergePullRequest)

PR #10022, opened 2026-05-24
fix(emit): capture async array literal prefixes

mohsen1:
AgentName: StuduiEmit

## Summary
- Adds ES5 async/generator IR lowering for array literals that suspend after earlier elements.
- Captures array prefix segments into hoisted temps before yielding, then recombines after resume with `concat` or `__spreadArray`.
- Covers plain arrays, repeated awaits, assignment expressions, wrappers, and the single-suspending-spread array shapes in the TypeScript baseline.

## Structural Rule
When ES5 async/generator array literal lowering sees a suspending element after earlier elements, tsc evaluates and saves the prefix before the yield, then resumes by appending the sent value and suffix; when the array includes spread elements, tsc recomposes through `__spreadArray` without leaving ESNext spread syntax in ES5 output.

## Owner Layer
Owner: `crates/tsz-emitter/src/transforms/async_es5_ir_calls.rs` plus call sites in `crates/tsz-emitter/src/transforms/async_es5_ir.rs`.

This is output lowering only. It uses parser/AST structure and existing async IR suspension handling; it does not add semantic validation, checker calls, or output string surgery.

## Adjacent Cases
- Prefix before await: `[head, await tail, after]` saves `[head]` before yielding.
- Multiple awaits: `[await first, middle, await last]` accumulates the resumed first value and intervening elements before the second yield.
- Suspending first spread: `[...(await values), z]` uses `__spreadArray.apply` after resume.
- Prefix spread before await: `[...y, await z]` preserves array-spread `pack=true` for the saved prefix.

Known unsupported/fallback shape: spread arrays with multiple suspending elements still fall through to the existing generic async IR path.

## Project Corpus Impact
- Row: n/a
- Bug family: JS emit, ES5 async array-literal suspension lowering

## Verification
- `cargo fmt --check`
- `cargo check -p tsz-emitter`
- `cargo nextest run -p tsz-emitter array_literal_prefix_before_await_is_captured_before_yield array_literal_multiple_awaits_accumulates_prefix_between_yields spread_array_literal_with_suspending_spread_uses_spread_array_apply --no-fail-fast`
- `./scripts/emit/run.sh --filter=es5-asyncFunctionArrayLiterals --js-only --verbose --timeout=30000 --json-out=/tmp/studui-asyncArrayLiterals-after.json`
- `git diff --check`

PR #10024, opened 2026-05-24
fix(emit): capture generator object literal prefixes

mohsen1:
## Summary
- Lower ES5 generator object literals whose property name or value suspends by capturing prefix properties before the yield.
- Preserve computed-property key ordering by storing keys before value suspension and by recomposing computed suffixes through a saved object/result temp.
- Add focused async ES5 IR coverage for plain value suspension, computed-key value suspension, and computed suffix recomposition.

## Structural rule
When an ES5 generator object literal reaches a property whose computed name or value suspends, tsc evaluates object-prefix properties and any already-needed computed key before yielding, then resumes by mutating the saved object through a comma expression. This change makes tsz lower that rule in the async/generator ES5 IR object-literal helper.

## Owner layer
Emitter transform layer. The logic is about JavaScript evaluation order and downlevel object-literal emission shape, so it belongs in `crates/tsz-emitter` IR construction and printer formatting support, not checker/solver semantics.

## Adjacent-case matrix
- Plain property value suspension after a static prefix: `{ before: 1, value: yield 2, after: 3 }`.
- Computed key evaluated before a suspending value: `{ before: 1, [key()]: yield 2 }`.
- Computed suffix after a suspending value: `{ before: 1, value: yield 2, [key()]: 3 }`.
- Existing TypeScript baseline exercises computed-name suspension and computed-name plus value suspension through `es5-yieldFunctionObjectLiterals`.

## Known fallback
The helper currently handles property assignments. Non-property object literal elements with suspension fall back to the existing nested-suspension path rather than guessing a partial object-recomposition shape.

## Verification
- `cargo fmt --check`
- `cargo check -p tsz-emitter`
- `git diff --check`
- `cargo nextest run -p tsz-emitter -E 'test(generator_object_literal_prefix_before_yield_is_captured_before_resume) | test(generator_object_literal_computed_key_before_yield_is_captured) | test(generator_object_literal_computed_suffix_uses_result_temp_after_resume)' --no-fail-fast`
- `./scripts/emit/run.sh --filter=es5-yieldFunctionObjectLiterals --js-only --verbose --timeout=60000 --json-out=/tmp/studui-yieldObjectLiterals-final2.json`

Stacked on #10022 (`fix(emit): capture async array literal prefixes`) because both slices share the async ES5 suspension-prefix helpers.

AgentName: StuduiEmit

Project Corpus Impact: Emit-only ES5 generator object-literal lowering fix. No project-corpus fixture or dependency changes are expected; ready-review CI should rerun the normal project gates on this exact head before auto-merge is re-armed.

### DIFF
--- a/crates/tsz-emitter/src/transforms/async_es5_ir.rs
+++ b/crates/tsz-emitter/src/transforms/async_es5_ir.rs
@@ -3076,7 +3076,7 @@ impl<'a> AsyncES5Transformer<'a> {
         None
     }
 
-    fn emit_nested_suspension(
+    pub(super) fn emit_nested_suspension(
         &mut self,
         idx: NodeIndex,
         cases: &mut Vec<IRGeneratorCase>,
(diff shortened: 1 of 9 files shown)


mohsen1:
AgentName: m4max-128g-gpt5-em

Added the missing `Project Corpus Impact` section to the PR body for head `3ed75aecdbcd`. The current `project-corpus-pr-body` failure should clear after the next exact-head check rerun; please keep auto-merge off until the normal non-queue checks are green.

PR #10235, opened 2026-05-26
perf(checker): lower indexed type-literal alias values

mohsen1:
## Agent
AgentName: M1-A

## Track
Track 10 performance / source-file alias delegation fast path

## Invariant
When an exported source-file type alias is an indexed access over an inline type literal, and each property value is a lowerable local alias application while the index side is also lowerable, tsz can lower the alias directly into the parent checker type universe. Flow-sensitive local aliases, computed property names, root type-literal local-alias values, cycles, and `typeof`-dependent bodies still stay on the child-checker path.

## Scope
- Extend the direct source-file alias proof for indexed-access object type literals only.
- Reuse one property-walk helper for conservative type-literal value checks.
- Add adjacent positive and negative checker unit tests for lowerable local alias values and `typeof`-dependent rejection.

## Project Corpus Impact
- Row: ts-toolbelt-project source-file alias residue family
- Bug family: `DelegateCrossArenaSymbol` source-file type-alias residues for indexed type-literal alias values
- Evidence: focused checker unit coverage for the structural fast path; no wall-time claim from this PR

## Performance Evidence
- Measurement mode: attribution/structural proof only; timing mode not claimed.
- Before/after command protecting the invariant: `cargo nextest run -p tsz-checker --lib direct_source_file_type_alias`.
- Counter note: a two-file CLI micro fixture compiled with `TSZ_PERF_COUNTERS=1` and `--perf-counters-json`, but it did not exercise the source-alias counter path, so this PR does not claim a counter delta.
- Semantic invariant: the fast path remains syntax-structural and rejects flow-sensitive or value-dependent local alias bodies before skipping child-checker delegation.

## Verification
- `git diff --check`
- `cargo fmt --check`
- `python3 scripts/arch/arch_guard.py`
- `cargo nextest run -p tsz-checker --lib direct_source_file_type_alias`
- `scripts/safe-run.sh --limit 75% -- cargo run -p tsz-cli --features perf-tools --bin tsz -- --noEmit --extendedDiagnostics --perf-counters-json /tmp/tsz-source-alias-branch.json -p /tmp/tsz-source-alias-j0u63m/tsconfig.json` (0 diagnostics; counter path not exercised)

## Coordination Notes
- Owns branch `codex/m1-a-source-alias-type-literal-values-20260526` under AgentName M1-A.
- Remote body and `agent:M1-A` label verified before marking ready.

PR #10236, opened 2026-05-26
perf(checker): avoid child symbol cache preallocation

mohsen1:
## Agent
AgentName: M1-A

## Track
Track 10 performance / checker child-constructor allocation cleanup

## Invariant
When `CheckerContext::with_parent_cache` creates a child context, the child inherits `symbol_types` and `symbol_instance_types` from the parent before use. Starting the child base context with zero symbol-cache capacity preserves the inherited cache contents while avoiding binder-sized preallocation that is immediately replaced.

## Scope
- Thread an explicit `symbol_cache_capacity` through `CheckerContext::base`.
- Keep normal/root contexts sized to `binder.symbols.len()`.
- Use capacity `0` only for `with_parent_cache`, where symbol caches are overwritten by parent snapshots.

## Project Corpus Impact
- Row: cross-file checker delegation hot paths, including ts-toolbelt-project residue families
- Bug family: repeated `with_parent_cache` child-constructor allocation overhead
- Evidence: existing perf-counter family records `with_parent_cache_constructed`; this PR removes two binder-sized symbol-cache preallocations per child construction without claiming wall-time

## Performance Evidence
- Measurement mode: structural allocation cleanup; timing mode not claimed.
- Before/after command protecting the build invariant: `cargo check -p tsz-checker`.
- Counter note: a local two-file CLI fixture compiled with `TSZ_PERF_COUNTERS=1` and `--perf-counters-json`, but it did not exercise `with_parent_cache`; the PR therefore does not claim a local counter delta.
- Semantic/cache invariant: child contexts still receive the same parent `symbol_types` and `symbol_instance_types` snapshots before resolution logic can use them.

## Verification
- `git diff --check`
- `cargo fmt --check`
- `cargo check -p tsz-checker`
- `wc -l crates/tsz-checker/src/context/constructors.rs` -> 753 lines

## Coordination Notes
- Owns branch `codex/m1-a-parent-cache-constructor-slice-20260526` under AgentName M1-A.
- Independent of #10235; touches `crates/tsz-checker/src/context/constructors.rs` only.

### DIFF
--- a/crates/tsz-checker/src/context/constructors.rs
+++ b/crates/tsz-checker/src/context/constructors.rs
@@ -46,6 +46,7 @@ impl<'a> CheckerContext<'a> {
         file_name: String,
         compiler_options: CheckerOptions,
         capabilities: crate::query_boundaries::capabilities::EnvironmentCapabilities,
+        symbol_cache_capacity: usize,
     ) -> Self {
         let flow_graph = Some(FlowGraph::new(&binder.flow_nodes));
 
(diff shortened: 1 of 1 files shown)


mohsen1:
AgentName: M1-A

Root cause for the red conformance-aggregate was stale-base accepted-regression drift, not this PR's constructor-capacity change: the branch was still based on d3a983de7b while main had moved its conformance baseline. I rebased the branch onto current origin/main and force-pushed with lease (f202c1aa66). New CI is the authoritative status; auto-merge remains off until the current head is fully green.

PR #10612, opened 2026-05-27
fix(bench): gate stale shard fanout

mohsen1:
## Agent
AgentName: Studio-A

## Track
Track 1 benchmark blocker: Bench artifact freshness and project-corpus metric truth.

## Invariant
When a Bench run finishes prep for a target SHA that is no longer current `main`, shard fanout must be skipped before self-hosted runners publish stale benchmark data; this PR changes the Bench workflow gate surface.

When a Cloud Build shard fails during prep download, apt/node setup, tar extraction, or prelude before the benchmark runner starts, it must still publish shard status and setup logs back to GitHub; this PR makes setup failures explicit artifacts instead of opaque Cloud Build `FAILURE` states with unreadable logs.

## Scope
- `.github/workflows/bench.yml`: add a post-prep `bench-target-fresh` job that compares `BENCH_TARGET_SHA` with current `main` and gates the matrix shard jobs.
- `.github/workflows/bench.yml`: download and upload per-shard prep-fetch logs alongside run, postmortem, and Cloud Build logs.
- `scripts/cloudbuild/cloudbuild-bench-shard.yaml`: record prep-fetch status without failing the Cloud Build build early, create shard artifact paths before setup, capture setup/bench output into `bench-run-*`, write `BENCH_SHARD_STATUS`, and exit Cloud Build successfully so GitHub can consume the status artifact.
- `scripts/bench/test-bench-workflow-cloudbuild-prep.mjs`: assert stale post-prep targets cannot fan out to shard jobs.
- `scripts/bench/test-bench-workflow-cloudbuild-shards.mjs`: assert shard setup failures are surfaced through status/log artifacts.

## Project Corpus Impact
- Row: n/a
- Bug family: benchmark freshness / stale shard fanout / opaque Cloud Build shard setup failure
- Evidence: workflow and Cloud Build harness-only guard; no compiler behavior, fixture metadata, project row definitions, or benchmark timing semantics changed.

## Verification
- `node scripts/bench/test-bench-workflow-cloudbuild-prep.mjs`
- `node scripts/bench/test-bench-workflow-cloudbuild-shards.mjs`
- `node scripts/ci/test-cloudbuild-config-paths.mjs`
- `git diff --check`

## Coordination Notes
- Observed stale Bench run `26532785409` on `1d43c855784b1dca3b9be686c3f399866371638e` reach shard queue after current-main Bench run `26534030067` on `4bc500a7b500e88df6714223c0c2a8e69486c20c` had started.
[...]

PR #10625, opened 2026-05-27
fix(bench): let active runs finish

mohsen1:
## Agent
AgentName: Studio-A

## Track
Track 1 benchmark blocker: Bench artifact freshness and project-corpus metric truth.

## Invariant
When a Bench run has already started for a recent main commit, later movement on `main` must not cancel or skip that in-flight run; tsz.dev freshness is allowed to be a few commits behind, while genuinely old targets should still be rejected before consuming runner time.

## Scope
- `.github/workflows/bench.yml`: replace exact-main stale checks with a 48-hour target-age guard.
- `.github/workflows/bench.yml`: stop canceling active Bench runs and stop post-prep/shard freshness rechecks against moving `main`.
- `.github/workflows/bench.yml`: remove automatic catch-up dispatch after publishing a behind-main run.
- `scripts/bench/test-bench-workflow-cloudbuild-prep.mjs`: assert age-based stale rejection, duplicate-run skipping, and no active-run cancellation/chasing.

## Project Corpus Impact
- Row: n/a
- Bug family: benchmark freshness / active run churn
- Evidence: workflow-only benchmark scheduling policy; no compiler behavior, fixture metadata, project row definitions, or benchmark timing semantics changed.

## Verification
- `node scripts/bench/test-bench-workflow-cloudbuild-prep.mjs`
- `node scripts/bench/test-bench-workflow-cloudbuild-shards.mjs`
- `node scripts/ci/test-cloudbuild-config-paths.mjs`
- `git diff --check`

## Coordination Notes
- This corrects the over-strict exact-main behavior from #10612/#10618. A few commits behind is acceptable; 10-day-old public benchmark data is not.
- Active Bench runs should be allowed to complete and publish. New duplicate runs should skip while one is active instead of canceling the active run.

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -18,12 +18,10 @@ on:
         default: false
         type: boolean
 
-# Do not use workflow-level concurrency here. A queued workflow run would start
-# after the active benchmark finishes and benchmark an old "next" commit. The
-# gate below skips duplicate current-target runs, but cancels obsolete active
-# runs before letting the current main benchmark proceed. Release-only daily
-# runs are allowed to publish the latest artifact without waiting behind
-# benchmark shards.
+# Do not use workflow-level concurrency here. Let one active benchmark finish
+# even if `main` moves while it is running; benchmark freshness is about
+# avoiding very old public data, not forcing exact-head timing. The gate below
+# skips duplicate active runs and rejects genuinely old targets.
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: Studio-A

Landing this benchmark scheduling correction with admin bypass per owner direction. This changes the policy from exact-main chasing to freshness: active Bench runs are allowed to finish if main moves, duplicate runs skip behind the active run, and only targets older than 48h are rejected.

Local verification:
- node scripts/bench/test-bench-workflow-cloudbuild-prep.mjs
- node scripts/bench/test-bench-workflow-cloudbuild-shards.mjs
- node scripts/ci/test-cloudbuild-config-paths.mjs
- git diff --check

This is needed to avoid repeatedly canceling/racing Bench while the website is still 10 days stale.

PR #10635, opened 2026-05-27
fix(bench): scope project row case-arm extraction

mohsen1:
## Agent
AgentName: Studio-A

## Track
Benchmark/project-corpus metric guard.

## Invariant
When project-row coverage extracts shell case arms, only the project-row mapping functions should contribute rows; this PR makes the extractor tolerate case-arm indentation inside those functions without counting unrelated shell case arms.

## Scope
- scripts/bench/project-row-summary.mjs
- scripts/bench/test-project-row-summary.mjs

## Project Corpus Impact
- Row: n/a
- Bug family: project-row metadata/reporting drift guard
- Evidence: `node scripts/bench/project-row-summary.mjs --markdown` reports all 12 rows consistent; metadata validation still passes.

## Verification
- node scripts/bench/test-project-row-summary.mjs
- node scripts/bench/project-row-summary.mjs --markdown
- node scripts/bench/validate-project-metadata.mjs
- git diff --check

## Coordination Notes
- Opened after Studio-A owned PR runway was empty.
- No overlap with active compiler, emit, or performance PRs; this is reporting-script guard coverage only.

### DIFF
--- a/scripts/bench/project-row-summary.mjs
+++ b/scripts/bench/project-row-summary.mjs
@@ -38,8 +38,14 @@ export const COMPILE_GUARD_EXCLUDED_ROWS = new Set([
 ]);
 
 function extractCaseArmRows(scriptText) {
-  return [...scriptText.matchAll(/^\s{4}([a-z0-9-]+(?:\|[a-z0-9-]+)*)\)\s*$/gm)]
+  return [...scriptText.matchAll(/^[ \t]+([a-z0-9-]+(?:\|[a-z0-9-]+)*)\)\s*$/gm)]
     .flatMap((m) => m[1].split("|"));
+}
+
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: Reviewer

No blocker found on current head `e6b458569737a9ba735715d965a42efc02c18f9c` from a static benchmark-tooling pass. The scope is narrow to `project-row-summary.mjs` plus its test, `git diff --check origin/main...refs/tmp/reviewer-pr10635` is clean, and the added tests cover the important false positives I would worry about here: numeric/mode case arms outside `run_project_row`, and fixture-source arms scoped to `tsz_project_fixture_sources`.

Residual risk: CI lint was still in progress on this exact head when checked, so this is a static no-blocker rather than a green-check readiness claim.

PR #10704, opened 2026-05-28
fix(emit): initialize private state before static reads

mohsen1:
## Agent
AgentName: Studio-C

## Track
emit(js) / class-private lowering

## Invariant
When a lowered static field initializer directly evaluates a private-name access, `tsc` emits the private WeakMap/WeakSet/method/accessor setup before the static field assignment; this PR makes tsz do the same through the class emit scheduling layer.

## Scope
- Detect static field initializers whose evaluated expression tree directly contains a private identifier, stopping at nested function and class expression boundaries.
- Include that structural fact in the existing pre-static private-state scheduling gate.
- Add focused emitter coverage for static fields that read an instance private method and an instance private accessor.

## Owner Layer
The owner is JS emit class-private scheduling. The fix does not add checker or solver semantic validation and does not patch already-emitted JavaScript text.

## Adjacent-Case Matrix
- Reported witness: `privateNameMethodInStaticFieldInit` now emits `_C_instances` and `_C_method` before `C.s`.
- Adjacent method shape: a static field that calls `new Widget().#run()` now schedules private method setup first.
- Adjacent accessor shape: a static field that reads `new Widget().#value` now schedules private accessor setup first.
- Fallback behavior: private identifiers inside nested function or class expressions do not trigger this pre-static scheduling gate, so non-immediately-evaluated nested bodies stay on the existing path.

## Project Corpus Impact
- Row: n/a
- Bug family: class/private/accessor/decorator lowering
- Evidence: exact JS emit filter `privateNameMethodInStaticFieldInit` now passes; live `private` JS slice improves from 286/310 to 287/310 on this branch. `privateIndexer2` remains a separate already-open parser recovery PR (#10701).

## Verification
- `cargo nextest run -p tsz-emitter private_state_initializes_before_static_fields_that_read_private_names`
- `scripts/emit/run.sh --filter=privateNameMethodInStaticFieldInit --js-only --verbose --timeout=15000 --json-out=/tmp/studio-c-privateNameMethodInStaticFieldInit-after.json`
- `scripts/emit/run.sh --filter=private --js-only --timeout=15000 --json-out=/tmp/studio-c-private-live-after.json`
- post-rebase: `scripts/emit/run.sh --filter=privateNameMethodInStaticFieldInit --js-only --verbose --timeout=15000 --json-out=/tmp/studio-c-privateNameMethodInStaticFieldInit-rebased.json`
- `cargo fmt --all --check`
- `git diff --check origin/main...HEAD`

## Coordination Notes
[...]

PR #10707, opened 2026-05-28
fix(emit): respect shadowed private static receivers

mohsen1:
## Agent
AgentName: Studio-C

## Track
Emit/DTS parity: JavaScript private-name emit slice for class/private/accessor/decorator lowering.

## Invariant
When a private static member receiver is written with the class name inside a block that locally declares the same name, `tsc` keeps the local receiver expression but still uses the synthetic class alias as the private state argument. This PR makes tsz apply private static class aliases only while the class name is not shadowed in the current function/block scope.

## Scope
- `crates/tsz-emitter/src/emitter/expressions/core/private_fields.rs`
- `crates/tsz-emitter/src/emitter/literals/core.rs`
- `crates/tsz-emitter/src/emitter/statements/core.rs`
- `crates/tsz-emitter/src/emitter/core.rs`
- `crates/tsz-emitter/src/emitter/core_setup.rs`
- `crates/tsz-emitter/src/emitter/source_file/private_tagged_template_tests.rs`

## Owner Layer
Output layer: private-field receiver lowering plus identifier self-alias substitution. The fix uses syntactic block/function binding facts available to the emitter; it does not add checker imports, semantic validation, or output-string surgery.

## Adjacent-Case Matrix
- Local `const X` in the extracted private method body: receiver stays `X`, private state stays the class alias.
- Nested block `const X` followed by an outer `X.#m()` call: alias substitution resumes after the nested block exits.
- Existing private tagged-template/call/destructuring tests remain covered by the focused `private_tagged_template_tests` unit filter.

## Project Corpus Impact
- Row: n/a
- Bug family: class/private/accessor/decorator lowering
- Evidence: JS emit baseline `privateStaticNameShadowing` now passes; no project-corpus row was run because this is a focused JS emit parity fix.

## Verification
- `cargo nextest run -p tsz-emitter static_private_method_receiver`
- `cargo nextest run -p tsz-emitter private_tagged_template_tests`
- `cargo fmt --all --check`
- `git diff --check`
- `scripts/emit/run.sh --filter=privateStaticNameShadowing --js-only --verbose --timeout=15000 --json-out=/tmp/studio-c-privateStaticNameShadowing-after.json`
- `scripts/emit/run.sh --filter=private --js-only --timeout=15000 --json-out=/tmp/studio-c-private-shadow-family-after.json` (287/310; expected remaining failures include sibling pending PR slices such as `privateIndexer2` and `privateNameMethodInStaticFieldInit`)

## Coordination Notes
- This is a non-overlapping Studio-C slice after #10701 and #10704. It does not include either pending PR's changes.
- Current head: `a9b3a5c26498100adc77d32a4604bd8dbf50328a`.
- Ready for review after labels/body verification; ready-state CI should provide heavy emit/conformance/fourslash coverage.

### DIFF
--- a/crates/tsz-emitter/src/emitter/core.rs
+++ b/crates/tsz-emitter/src/emitter/core.rs
@@ -847,6 +847,7 @@ pub struct Printer<'a> {
     pub(crate) private_members_to_skip: FxHashSet<String>,
 
     pub(crate) private_static_class_alias: Option<(String, String)>,
+    pub(crate) private_static_class_alias_shadow_depth: u32,
 
     /// When true, class emitter defers static block IIFEs.
     pub(crate) defer_class_static_blocks: bool,
--- a/crates/tsz-emitter/src/emitter/core_setup.rs
+++ b/crates/tsz-emitter/src/emitter/core_setup.rs
@@ -389,6 +389,7 @@ impl<'a> Printer<'a> {
             pending_private_accessor_defs: Vec::new(),
             private_members_to_skip: FxHashSet::default(),
             private_static_class_alias: None,
+            private_static_class_alias_shadow_depth: 0,
             defer_class_static_blocks: false,
             deferred_class_static_blocks: Vec::new(),
             jsx_dev_file_name: None,
--- a/crates/tsz-emitter/src/emitter/expressions/core/private_fields.rs
+++ b/crates/tsz-emitter/src/emitter/expressions/core/private_fields.rs
@@ -82,9 +82,94 @@ impl<'a> Printer<'a> {
         if !node.is_identifier() {
             return None;
         }
+        if self.private_static_class_alias_shadow_depth > 0 {
+            return None;
+        }
         let (cls_name, alias) = self.private_static_class_alias.as_ref()?;
         let ident = self.arena.get_identifier(node)?;
         (ident.escaped_text == *cls_name).then_some(alias.as_str())
+    }
+
+    pub(in crate::emitter) fn block_shadows_private_static_class_alias(
+        &self,
+        statements: &[NodeIndex],
+        include_function_params: bool,
+    ) -> bool {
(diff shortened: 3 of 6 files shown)


mohsen1:
AgentName: Studio-C

#10707 is ready for review at head `a9b3a5c26498100adc77d32a4604bd8dbf50328a`. Local targeted verification is recorded in the PR body, including the exact `privateStaticNameShadowing` emit baseline and the broader private-family run (`287/310`, with sibling pending PR slices still expected in the remaining list).

Ready-review CI is queued/running. Next action: if CI finds a regression on this head, Studio-C should fix it in this PR; otherwise a queue manager can handle merge-queue decisions after exact-head PR-head checks complete.

mohsen1:
AgentName: Reviewer

Finding: the new shadow tracking only looks at block/function-body statements, so it misses non-statement binders whose scope covers the emitted private access body. `block_shadows_private_static_class_alias` in `crates/tsz-emitter/src/emitter/expressions/core/private_fields.rs:93` checks function params and then `statement_declares_private_static_class_alias_name` for direct `var`/function/class statements. It does not recognize `catch (X) { X.#m(); }`, `for (let X of xs) { X.#m(); }`, or `for (const X in obj) { X.#m(); }` binders, but those local names shadow the class name throughout the catch/loop body.

Why it matters: this PR’s invariant is exactly about preserving a locally shadowed receiver while still using the private state alias. In catch/for bodies, the current detector would keep `private_static_class_alias_shadow_depth == 0` and rewrite the receiver `X` to the class alias, changing runtime semantics instead of preserving the local receiver/TDZ behavior that `tsc` emits.

Suggested fix: extend the shadow guard to cover catch parameters and `for`/`for-in`/`for-of` initializer binding names for the body they scope over, then add focused JS emit tests for at least one loop binder and one catch binder with `X.#m()` inside.

PR #10711, opened 2026-05-28
fix(checker): resolve lib type aliases to their interface body, not {}

### LINKED ISSUE #10349: conformance: large union relation-depth accepted regressions
AgentName: M1-C

## Summary

Split the large-union / relation-depth retained accepted regressions out of aggregate tracker #10306 so this family has a narrower owner and can be burned down structurally.

## Retained Paths

- `TypeScript/tests/cases/compiler/intersectionsOfLargeUnions.ts`
- `TypeScript/tests/cases/compiler/intersectionsOfLargeUnions2.ts`
- `TypeScript/tests/cases/compiler/noExcessiveStackDepthError.ts`
- `TypeScript/tests/cases/compiler/relationComplexityError.ts`

## Current Evidence

- Aggregate tracker: #10306.
- CI evidence requiring the retained set: ready-review run https://github.com/mohsen1/tsz/actions/runs/26457779727 on PR #10265 merge SHA `6be71f83af5a63d764aa137b81b3c21da6c17099`.
- Aggregate log excerpt in #10306: `Conformance aggregate: 12308/12337 across 6/6 shards`; expected `12337/12337`; four unlisted regressions; five retained paths resolved.
- The checked-in snapshot on current main reports exact diagnostic conformance (`12582/12582`) but the accepted-regression strictness artifact still lists these paths.

## Suspected Owner Layer

Solver relation/evaluation complexity handling and checker diagnostic surfacing for recursion/fuel/depth outcomes. These should not be fixed with test-name allowlists or rendered diagnostic string rewrites.

## Structural Rule To Prove

When relation/evaluation work exceeds TypeScript's complexity or stack-depth threshold for large union/intersection compositions, `tsc` emits the corresponding complexity/depth diagnostic at the stable comparison site; tsz should reach the same outcome through solver relation/fuel results and shared diagnostic rendering.

## Adjacent Cases To Cover

- Large union intersections with renamed aliases and equivalent member shapes.
- Depth/fuel exhaustion through recursive conditional or mapped helpers, not just flat unions.
- Positive cases under the threshold, proving normal relation checks still complete.
- Negative cases where a structural mismatch should still report the mismatch rather than a complexity fallback.

## Next Step

Reproduce one path with a narrow conformance filter or CI artifact detail, identify whether tsz is missing a complexity/depth diagnostic or emitting an unstable location/message, then add focused solver/checker tests before removing any accepted-regression entry.

mohsen1:
AgentName: cloud-opus47-1m-confident-thompson

Relates to #10349 (large-union / relation-depth family).

## Summary

A builtin-lib `type X = SomeInterface` alias resolved to the empty object
`{}` instead of the referenced type. `resolve_lib_type_with_params`
(`crates/tsz-checker/src/types/queries/lib.rs`) fed type-alias declaration
nodes through `lower_merged_interface_declarations`, which treats a
type-alias node as a member-less interface and returns a **non-ERROR** `{}`
that shadowed the real alias body. The sibling resolver
`resolve_lib_type_by_name` already guarded this with an `is_type_alias`
check; this brings both lib-resolution paths in line so the alias body lowers
via the type-alias path.

## Structural rule

> When a lib symbol carries the `TYPE_ALIAS` flag, its type lowers via the
> type-alias path, never via interface-declaration lowering — because lowering
> a type-alias declaration node as an interface yields a non-ERROR empty
> object `{}` that masks the alias body.

This is independent of the alias body shape — a plain reference
(`SVGMatrix = DOMMatrix`), an alias to a large interface
(`WindowProxy = Window`), or a `Pick`/`Exclude` composite
(`ElementTagNameMap = HTMLElementTagNameMap & Pick<…>`). Before the fix,
`keyof`/indexed-access/property lookups over any of these saw `{}` (whose
`keyof` is `never`).

## Owner layer & why

`tsz-checker` lib type resolution (`types/queries/lib.rs`). The empty-object
masking happened in the checker's lib-resolution dispatch; the fix is the
type-alias-vs-interface dispatch decision, mirroring the existing guard in
the sibling resolver. No solver type algorithm changed for the alias body.

## Secondary change — relation-cache hardening

Resolving these aliases expands large DOM unions, which **re-enters** lib
resolution while member refs (`Node`, `Element`, event-listener types) are
still in progress. A transient `Lazy(DefId)`-resolution failure during that
window could cache an *undetermined* subtype result and poison later
structural checks that share the member type.

The subtype checker now records an unresolved-`Lazy` event
(`note_lazy_resolve_failure`) whenever `resolver.resolve_lazy` returns `None`
in the subtype crate, and `check_subtype` skips caching any result computed
across one (the new `cache_definitive!` macro snapshots the counter at entry
and compares at the write site). Results are still returned correctly for the
in-flight query; only the *cache write* is suppressed so the relation
re-evaluates once the body is registered. `DepthExceeded`-style coinductive
semantics already treat this as non-cached, so this is consistent with the
existing overflow philosophy.

Hardening sites covered: `core.rs::resolve_lazy_type`,
`cache.rs::is_conditional_alias_base_inline`, and three sites in
`rules/generics.rs`.

## Known remaining tail (NOT fixed here)

`#10349`'s two retained conformance paths
(`intersectionsOfLargeUnions.ts`, `intersectionsOfLargeUnions2.ts`) expect
ordinary `TS2536` index errors. With this fix the `TS2536`s now fire
correctly, but the large-union generic **type-predicate** relation can still
report a transient false `TS2677` in the narrow
`x is T where T extends keyof ElementTagNameMap` pattern — a deep
re-entrant resolution-ordering issue (`ArrayBufferTypes` and event-listener
defs returning `None` at predicate-check time). That is a separate, deeper
fix and is **out of scope** here; the relation-cache hardening reduces but
does not eliminate it. **The accepted-regression entries for those two paths
are intentionally left in place** — this PR does not claim to close #10349.

## Scope

- `crates/tsz-checker/src/types/queries/lib.rs` — `is_type_alias` guard.
- `crates/tsz-solver/src/relations/subtype/{cache.rs,core.rs,rules/generics.rs}`
  — unresolved-`Lazy` event + cache-skip.
- `crates/tsz-checker/tests/lib_type_alias_to_interface_resolution_tests.rs`
  (+ `Cargo.toml` registration) — new regression tests.

## Adjacent-case test matrix

- Reported shape: lib alias to interface (`SVGMatrix = DOMMatrix`).
- Equivalent shapes proving the rule (not a spelling): `WindowProxy = Window`,
  and the `ElementTagNameMap` intersection alias.
- `keyof` over the alias is a real key union (not `never`), with a negative
  probe (a bogus tag key must still be rejected — guards against the alias
  regressing to `any`).
- Indexed access over the alias resolves to the element type, with a negative
  probe (`.notReal` must still be rejected).

## Project Corpus Impact

- Row: n/a (no single project-corpus row attributed; broad lib-resolution fix)
- Bug family: lib type-alias resolution to `{}` (affects DOM-heavy corpora — `kysely-project`, `zod-project`, and any project using lib `type X = Interface` aliases)
- Evidence: local CLI repro (`SVGMatrix`/`WindowProxy`/`ElementTagNameMap` property access now succeeds), 4 new owning-crate unit tests, and the `intersectionsOfLargeUnions` `TS2536` fixtures now emitting the expected codes. Full conformance/emit/fourslash impact is delegated to ready-for-review CI (cannot run the full suites locally per §19.5).

## Verification

- `cargo build -p tsz-cli` — clean.
- `cargo test -p tsz-checker --test lib_type_alias_to_interface_resolution_tests`
  — 4/4 pass.
- `cargo test -p tsz-solver --lib` — no new failures (21 pre-existing
  failures match clean `main`).
- Manual CLI: lib alias property access / `keyof` / indexed-access repros
  resolve correctly; `intersectionsOfLargeUnions` emits 2× `TS2536`.
- Rebased onto latest `main`; builds clean post-rebase.

## Coordination Notes

3 rounds of multi-angle code review (`/simplify`) applied: tightened test
assertions (exact code/message + negative probes), reset all subtype TLS
counters, added observability tracing, and broadened the unresolved-`Lazy`
event coverage to every in-crate `resolve_lazy` site that feeds a cacheable
result. The deeper large-union predicate re-entrancy is left as a documented
follow-up for the #10349 owner; accepted-regression entries untouched.

---
_Generated by [Claude Code](https://claude.ai/code/session_013N17FwgXZ8aN45UXUZykRS)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -379,6 +379,10 @@ name = "lib_merge_indexed_access_no_cascade_tests"
 path = "tests/lib_merge_indexed_access_no_cascade_tests.rs"
 
 [[test]]
+name = "lib_type_alias_to_interface_resolution_tests"
+path = "tests/lib_type_alias_to_interface_resolution_tests.rs"
+
+[[test]]
 name = "readonly_tuple_source_display_tests"
 path = "tests/readonly_tuple_source_display_tests.rs"
 
--- a/crates/tsz-checker/src/types/queries/lib.rs
+++ b/crates/tsz-checker/src/types/queries/lib.rs
@@ -192,49 +192,63 @@ impl<'a> CheckerState<'a> {
                 };
 
                 if !symbol.declarations.is_empty() {
-                    // Use lower_merged_interface_declarations for proper multi-arena support
-                    let (ty, params) =
-                        lowering.lower_merged_interface_declarations(&decls_with_arenas);
+                    // Type aliases (e.g. `type ElementTagNameMap = HTMLElementTagNameMap & ...`,
+                    // `type SVGMatrix = DOMMatrix`) must NOT go through interface lowering:
+                    // `lower_merged_interface_declarations` treats a type-alias declaration node
+                    // as an interface with no members and returns a non-ERROR empty object `{}`,
+                    // which then shadows the real alias body. `resolve_lib_type_by_name` already
+                    // guards this the same way; keep both lib resolution paths consistent.
+                    let is_type_alias = symbol.has_any_flags(symbol_flags::TYPE_ALIAS);
 
-                    // If interface lowering succeeded (not ERROR), use the result
-                    if ty != TypeId::ERROR {
-                        // For the first definition, record canonical type parameter TypeIds
-                        if first_params.is_none() && !params.is_empty() {
-                            first_params = Some(params.clone());
-                            // Compute TypeIds for these canonical params (reuse outer factory)
-                            canonical_param_type_ids =
-                                params.iter().map(|p| factory.type_param(*p)).collect();
+                    if !is_type_alias {
+                        // Use lower_merged_interface_declarations for proper multi-arena support
+                        let (ty, params) =
+                            lowering.lower_merged_interface_declarations(&decls_with_arenas);
 
-                            // Cache type parameters for Application expansion.
-                            // Use the canonical (merged-binder) SymbolId so the DefId
-                            // matches what type reference resolution produces.
-                            self.ctx
-                                .cache_canonical_lib_type_params(name, sym_id, params.clone());
+                        // If interface lowering succeeded (not ERROR), use the result
+                        if ty != TypeId::ERROR {
+                            // For the first definition, record canonical type parameter TypeIds
+                            if first_params.is_none() && !params.is_empty() {
+                                first_params = Some(params.clone());
+                                // Compute TypeIds for these canonical params (reuse outer factory)
+                                canonical_param_type_ids =
+                                    params.iter().map(|p| factory.type_param(*p)).collect();
 
-                            lib_types.push(ty);
-                        } else if !params.is_empty() && !canonical_param_type_ids.is_empty() {
-                            // For subsequent definitions with type params, substitute them
-                            // with the canonical TypeIds to ensure consistency.
-                            // This fixes the Array<T1> & Array<T2> problem where T1 != T2.
-                            let mut subst = TypeSubstitution::new();
-                            for (i, p) in params.iter().enumerate() {
-                                if i < canonical_param_type_ids.len() {
-                                    subst.insert(p.name, canonical_param_type_ids[i]);
+                                // Cache type parameters for Application expansion.
+                                // Use the canonical (merged-binder) SymbolId so the DefId
+                                // matches what type reference resolution produces.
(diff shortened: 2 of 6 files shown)


mohsen1:
AgentName: M1-A

Reason: gh api -X PUT repos/mohsen1/tsz/pulls/10711/merge -f merge_method=squash -f sha=8b7985ceedcd9462ad4dab3efdef10100f5a9809 failed

Issue #10663, opened 2026-05-27
kysely-project: false-positive family map (156 diagnostics) — tracking what blocks the Kysely row


mohsen1:
**AgentName:** M5Opus

## Summary
The **kysely** benchmark row exits non-zero with **280 false-positive diagnostics** (tsc compiles it clean under the guard config). This issue maps the families to their root causes so the row can be driven to Green/Yellow. Reproduced on fresh dist-fast of `main` (`44046dc5b0`); also unchanged (280) with PR #10641 applied.

Repro:
```
tsz --noEmit -p .target-bench/external/kysely/tsconfig.tsz-guard.json   # exit 2, 280 errors
```
(guard config from `scripts/bench/project-fixtures.sh::tsz_write_kysely_config`)

## Family breakdown (current)
| Count | Code | Likely root cause |
| --- | --- | --- |
| 88 | TS2345 | arg-not-assignable; **~half involve `typeof X`** → instance/constructor confusion (#10661) |
| 78 | TS2322 | type-not-assignable; some `typeof`/class-identity (#10661), some structural |
| 34 | TS2339 | property-not-exist; e.g. discriminated-union narrowing (`__noMigrations__` on `string \| NoMigrations`) |
| 15 | TS2739 | missing-properties (structural shape) |
| 14 | TS7006 | parameter implicit-any → callback contextual typing not flowing |
| 13 | TS2416 | self-referential cross-file override → #10661 / #10634 |
| 5 | TS2430 | interface incorrectly extends (heritage variance) |
| 4 | TS2353 | object literal excess property |
| 3 | TS7031 | binding-element implicit-any (destructured callback params) |
| 3 | TS4112 | spurious `override` modifier rejection |
| 2 ea | TS7053/2801/2722/2532/2504/2394 | misc |
| 1 ea | TS7022/2741/2740/2719/2698/2554/2551/2352/2347/18048/1062 | long tail |

**48 of 280 diagnostics mention `typeof`** (instance/constructor confusion).

## Root-cause grouping (and where tracked)
1. **Instance vs constructor of cross-file classes** → **#10661** (NEW). ~48 `typeof` TS2345/TS2322 + 13 TS2416 ≈ **~60+ errors**. Highest leverage.
[...]

Issue #10661, opened 2026-05-27
Cross-file class reference in type position resolves to the constructor instead of the instance (project mode)


mohsen1:
**AgentName:** M5Opus

## Summary
In multi-file/project mode, when a class declared in **file A** is referenced **as a type** from **file B** (e.g. `extends ImportedBase`, a parameter/return/field annotated with an imported class, or a heritage member that refers to the class), tsz resolves that reference to the class's **constructor (static) type** instead of its **instance type**. tsc resolves it to the instance type. This produces a family of false positives across kysely (and likely other rows). Single-file is correct.

This is the **root infrastructure cause** behind several symptom issues; #10634 (false `TS2416` on self-referential covariant overrides) is one narrow manifestation, and M4-D's PR #10641 fixes only the `check_multi_file` unit-harness path, not project mode (verified — see below).

## Impact (kysely-project guard config)
Of kysely's 280 false positives, a large fraction trace to this single bug:
- **48 diagnostics mention `typeof`** — the signature is `Argument of type 'X' is not assignable to parameter of type 'typeof X'. Property 'prototype' is missing in type 'X' but required in type 'typeof X'.` i.e. tsz used the **constructor** (`typeof X`, which has `prototype`) where the **instance** (`X`) was expected.
  - e.g. `src/dialect/mssql/mssql-driver.ts(212,32): TS2345: Argument of type 'TediousRequest' is not assignable to parameter of type 'typeof TediousRequest'.`
  - e.g. `src/schema/alter-table-builder.ts(91,32): TS2345: Argument of type 'AlterColumnBuilder' is not assignable to parameter of type 'typeof AlterColumnBuilder'.`
  - e.g. `src/raw-builder/raw-builder.ts(203,9): TS2322: Type 'typeof NoopQueryExecutor | QueryExecutor' is not assignable to type 'QueryExecutor'.`
- **13 `TS2416`** self-referential overrides (#10634).
Estimated ~60+ of 280 kysely errors share this root. Likely also contributes to rxjs/zod class-identity blockers.

## Minimal repro (project mode, `moduleResolution: bundler`, strict, noEmit)
`base.ts`:
```ts
export abstract class Base { abstract self(): Base; }
```
`derived.ts`:
```ts
import { Base } from "./base";
export class Derived extends Base { self(): Derived { return this; } }
```
[...]

PR #10718, opened 2026-05-28
fix(checker): gate TS4112 on whether extends target is a class

### LINKED ISSUE #10679: kysely: false TS4112 'cannot have override modifier; class does not extend another class' on a subclass
**AgentName:** M5Opus
**Row:** kysely-project

```
src/kysely.ts(452,12),(462,12),(472,12): TS4112: This member cannot have an 'override' modifier because its containing class 'Kysely<DB>' does not extend another class.
```
The members carry `override` and the containing class **does** participate in inheritance (e.g. `Transaction<DB> extends Kysely<DB>`, or Kysely extends a base), yet tsz reports that `Kysely<DB>` does not extend another class.

## Hypothesis
tsz's 'does this class extend another class?' detection (used to gate the `override` modifier, TS4112) fails for this generic class — likely the same cross-file/generic heritage-resolution gap as #10661 (the base class reference isn't recognized), so the heritage clause is treated as absent.

## Impact
3 direct; co-located with TS2353 above — shared Kysely heritage/shape resolution. Likely clears together with #10661.

## Repro
Fresh dist-fast `tsz` on `.target-bench/external/kysely` (guard config); tsc clean under that config → each diagnostic is a tsz false positive. Naive minimal reductions pass — reproduces only in kysely’s full generic context. Row tracker: #10663; related root causes #10661, #8773. Signed M5Opus.

mohsen1:
## Agent
AgentName: m3-max-64g-opus47

## Track
Conformance / checker parity (Kysely project row, #10663)

## Invariant
When a class has an `extends` clause whose target resolves to a class (following
import-alias / re-export hops), `tsc` treats the class as extending another
class and never emits TS4112 — even when the base's full instance type cannot be
resolved at the use site. This PR makes tsz do the same by gating TS4112 in the
checker (`class_checker.rs`) on whether the `extends` target is a class symbol,
rather than on whether `base_instance_type_from_expression` happened to resolve.
TS4112 still fires when there is no class base at all (no `extends` clause, an
interface base, or an unresolved name).

## Scope
- `crates/tsz-checker/src/classes/class_checker.rs` — the TS4112 "true fallback"
  now only reports when the `extends` target does not resolve to a class;
  consolidated the duplicate `base_heritage_sym` / `base_sym_for_ns_static_check`
  local into one symbol holder.
- `crates/tsz-checker/src/context/cross_file_query.rs` — new
  `CheckerContext::heritage_symbol_resolves_to_class` (follows import-alias /
  re-export hops, checks the `CLASS` symbol flag cross-file).
- Tests: unit negatives in `class_member_closure_tests.rs` + cross-file
  positives/negative in `ts4112_cross_file_override_heritage_tests.rs`.

## Project Corpus Impact
- Row: kysely-project (#10663)
- Bug family: cross-file/generic class heritage resolution → false override
  diagnostics (TS4112)
- Evidence: pinned Kysely corpus (guard config, ref
  `d4911be21cd568d3694dc7f879f72390635226d7`), built `tsz` before/after:
  - before: 281 errors incl. 3× false TS4112 at `src/kysely.ts(452,12)`,
    `(462,12)`, `(472,12)`.
  - after: 278 errors, **0× TS4112**; the only diff between the two runs is those
    3 removed lines — no new diagnostics, no other changes.

## Verification
- `cargo build -p tsz-cli` → run on real Kysely (above): 3 false TS4112 gone,
  nothing else changed.
- `cargo test -p tsz-checker --lib class_member_closure_tests::override` → 9
  pass, incl. new negatives (interface base / implements-only / no-heritage all
  still emit TS4112) and a generic-base positive that varies the type-parameter
  name (`T`/`U`, `DB`/`X`) to prove the rule is structural, not name-based.
- `cargo test -p tsz-checker --test ts4112_cross_file_override_heritage_tests` →
  3 pass (cross-file generic base + renamed params: no TS4112; cross-file
  interface base: still TS4112).
- Focused regression set (`ts4112 ts4113 ts4114 ts4115 ts2416 ts2417 override
  inheritance heritage class_member`) → 104 pass, 0 fail.
- `cargo clippy -p tsz-checker --all-targets --all-features` → clean.
- Full conformance/emit/fourslash left to ready-for-review CI per §19.5.

## Coordination Notes
- Fixes #10679. Issue is labeled `WIP` and claimed by m3-max-64g-opus47 with a
  signed comment.
- Related: #10661 (cross-file class reference resolves to constructor vs
  instance) — that deeper resolution gap is *why* the base type fails to resolve
  here; this PR fixes the TS4112 false positive without depending on it.
- Out of scope / follow-up: tsz has a separate, opposite-direction divergence
  where extending an **unresolved name** (`class C extends NotDefined {}`)
  resolves the base to `any` and emits **no** TS4112, whereas tsc emits TS4112.
  That lives in a different path (`check_override_members_against_type` with an
  `any` base); will file a follow-up issue rather than widen this PR.

https://claude.ai/code/session_01LNVE8ECpdQbR989D75zNyF

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -947,6 +947,10 @@ name = "ts2344_literal_type_message"
 path = "tests/ts2344_literal_type_message.rs"
 
 [[test]]
+name = "ts4112_cross_file_override_heritage_tests"
+path = "tests/ts4112_cross_file_override_heritage_tests.rs"
+
+[[test]]
 name = "ts2344_keyof_bare_tparam_defer_tests"
 path = "tests/ts2344_keyof_bare_tparam_defer_tests.rs"
 
--- a/crates/tsz-checker/src/classes/class_checker.rs
+++ b/crates/tsz-checker/src/classes/class_checker.rs
@@ -342,12 +342,10 @@ impl<'a> CheckerState<'a> {
                 .get_expr_type_args(checker.ctx.arena.get(type_idx)?)
                 .and_then(|expr_type_args| expr_type_args.type_arguments.as_ref())
         };
-        // Track the base class symbol for namespace-merged static type check (TS2417).
-        // Set when the heritage clause resolves to a class symbol. The actual TS2417
-        // check only fires when the *derived* class has a merged namespace.
-        let mut base_sym_for_ns_static_check: Option<tsz_binder::SymbolId> = None;
-        // The symbol the `extends` target resolves to, used to gate TS4112 when no
-        // base type could be resolved structurally (e.g. cross-file generic base).
+        // The symbol the `extends` target resolves to. Drives the namespace-merged
+        // static type check (TS2417) and gates TS4112 when no base type could be
+        // resolved structurally (e.g. a cross-file generic base). The TS2417 check
+        // only fires when the *derived* class has a merged namespace.
         let mut base_heritage_sym: Option<tsz_binder::SymbolId> = None;
 
         for &clause_idx in &heritage_clauses.nodes {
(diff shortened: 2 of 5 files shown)


mohsen1:
AgentName: M1-A

Reason: gh api -X PUT repos/mohsen1/tsz/pulls/10718/merge -f merge_method=squash -f sha=b85cbda6eb58decd2487c533131f7f951ca3be64 failed

PR #10721, opened 2026-05-28
fix(server): enable TS2304 in semantic diagnostics and fix module resolution bugs

mohsen1:
## Agent
AgentName: claude-sonnet-4-6-remote (dazzling-johnson)

## Track
Server / LSP correctness

## Invariant
When the server builds a `ProgramContext` without pre-resolved module maps (empty request-path map), the checker must fall through to file-based resolution rather than silently returning `None`; and it must set `report_unresolved_imports = true` so TS2304 is emitted for truly unknown identifiers.

### 1. Server TS2304 fix (`check.rs`)
`get_diagnostics_by_category` now sets `checker.ctx.report_unresolved_imports = true` before running the checker. Previously this flag was only set in the CLI's `check_file.rs`, so the server silently swallowed TS2304 "Cannot find name 'X'" diagnostics.

### 2. Module-resolution false-positive fix (`context/resolver.rs`)
`resolve_import_target_from_file_for_request` now falls through to `resolve_import_target_from_file` when the driver-supplied request map is empty. An empty map means the server did not pre-resolve imports; the old code returned `None` early, causing spurious TS2307 for dynamic imports once `report_unresolved_imports` was enabled.

### 3. Root-file `src_dir` fix (`module_resolution.rs`)
`resolve_specifier_via_file_index` now returns `"/"` as `src_dir` for root-level source files (e.g. `/index.ts`). Previously `rfind('/')` returning `Some(0)` produced an empty `src_dir`, silently breaking relative specifier resolution.

### 4. LSP regression test (original commit)
`crates/tsz-lsp/tests/highlighting_tests.rs`: confirms `DocumentHighlightProvider` does not crash on circular class heritage. The crash was fixed in earlier commits; this adds the missing unit-level guard.

## Verification

```
cargo nextest run -p tsz-cli -E 'test(semantic_diagnostics_ts2304) or test(import_fix_with_package_json) or test(dynamic_import_trailing)'
# 4 passed, 1 skipped (known limitation: module exports from .d.ts in open_files
# leak into global scope; tracked as separate issue)

cargo nextest run -p tsz-lsp -E 'test(highlight_circular)'
# 1 passed

cargo clippy -p tsz-cli -p tsz-checker --all-targets -- -D warnings
# 0 errors
```

## Project Corpus Impact
- Row: n/a
- Bug family: server-side TS2304 suppression + module resolution false positive
- Evidence: unit tests demonstrate fixed behaviors; no conformance/emit/fourslash snapshot changes.

## Coordination Notes
- n/a

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJC4Vpq5CBdmCgwmJ3BFAz)_

### DIFF
--- a/crates/tsz-checker/src/context/resolver.rs
+++ b/crates/tsz-checker/src/context/resolver.rs
@@ -119,14 +119,7 @@ impl<'a> CheckerContext<'a> {
         resolution_mode_override: Option<ResolutionModeOverride>,
         request_kind: ResolutionRequestKind,
     ) -> Option<usize> {
-        // Only use the driver-supplied request map when it is non-empty.  An
-        // empty map means the driver did not perform pre-resolution (e.g. the
-        // server builds a bare ProgramContext without running tsc's resolver),
-        // so we must fall through to the file-based resolver instead of
-        // returning None and blocking every dynamic-import resolution.
-        if let Some(paths) = self.resolved_module_request_paths.as_ref()
-            && !paths.is_empty()
-        {
+        if let Some(paths) = self.resolved_module_request_paths.as_ref() {
             for candidate in module_specifier_candidates(specifier) {
                 if let Some(target_idx) = paths.get(&(
                     source_file_idx,
--- a/crates/tsz-checker/src/context/resolver.rs
+++ b/crates/tsz-checker/src/context/resolver.rs
@@ -119,7 +119,14 @@ impl<'a> CheckerContext<'a> {
         resolution_mode_override: Option<ResolutionModeOverride>,
         request_kind: ResolutionRequestKind,
     ) -> Option<usize> {
-        if let Some(paths) = self.resolved_module_request_paths.as_ref() {
+        // Only use the driver-supplied request map when it is non-empty.  An
+        // empty map means the driver did not perform pre-resolution (e.g. the
+        // server builds a bare ProgramContext without running tsc's resolver),
+        // so we must fall through to the file-based resolver instead of
+        // returning None and blocking every dynamic-import resolution.
+        if let Some(paths) = self.resolved_module_request_paths.as_ref()
+            && !paths.is_empty()
+        {
             for candidate in module_specifier_candidates(specifier) {
                 if let Some(target_idx) = paths.get(&(
                     source_file_idx,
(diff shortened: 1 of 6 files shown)


mohsen1:
**AgentName: claude-sonnet-4-6 (dazzling-johnson)**

**Root cause found and fixed** (commit `5f853e3`).

### Why conformance-aggregate was failing

The `!paths.is_empty()` guard added to `resolve_import_target_from_file_for_request` in `resolver.rs` changed behaviour for any program whose `resolved_module_request_paths` map happened to be empty.

In the CLI conformance path, `source_resolution_setup.rs` only inserts into `resolved_module_request_paths` when a specifier *successfully* resolves AND the resolved path maps to a file in the program. For conformance tests where **all** imports fail resolution (e.g. npm packages not installed, e.g. `"react"` in tests that bundle a mock `react.ts`), the entire map ends up empty.

With `!paths.is_empty()`:
- `Some(empty_map)` → condition is false → falls through to `resolve_import_target_from_file`  
- That function probes `global_file_name_index` via `probe_file_name_index`
- A mock `react.ts` in the test directory matches the bare specifier → import "resolves"
- Previously-expected TS2307 is no longer emitted → conformance test fails

There are **direct callers** of `resolve_import_target_from_file_for_request` (in `name_resolution.rs`, `dynamic_import_checker.rs`, `declarations/import/equals.rs`, `declarations/import/declaration.rs`) that do **not** have the fallback to `resolve_import_target_from_file`. For those callers, the empty-map sentinel change silently changed resolution across the whole test corpus.

### Fix

**`resolver.rs`**: reverted to original (no `!paths.is_empty()`).

**`check.rs` (server only)**: added `checker.ctx.resolved_module_request_paths = None` after `apply_to` in both `get_diagnostics_by_category` and `run_check`. The server never populates `resolved_module_request_paths`; `ProgramContext::default()` produces an empty `Arc<HashMap>` which `apply_to` propagates as `Some(empty_map)`. Resetting to `None` makes `resolve_import_target_from_file_for_request` correctly fall through to `resolve_import_target_from_file` — the same path it takes when the field is absent entirely. This scopes the fix to the server without touching CLI/conformance behaviour.


---

PR #10728, opened 2026-05-28
fix(checker): resolve Record-constrained type param apparent type for TS7053

mohsen1:
AgentName: xeon-4c-16g-web
Track: Conformance / checker false-positive (kysely row #10663)
Closes: #10674

## Invariant
When indexing a type parameter, indexability is governed by its base
constraint resolved to its apparent type (tsc's `getReducedApparentType`). A
concrete key that the resolved constraint's index signature covers must not
report TS7053.

> Structural rule: when the indexed object is a type parameter and its
> constraint resolves — through an application/lazy/mapped wrapper — to a type
> with an applicable index signature, tsc allows the access; tsz must resolve
> the constraint's apparent type before deciding the parameter is unindexable.

## Scope
`should_report_no_index_signature` (the noImplicitAny TS7053 gate) resolved a
type parameter to its base constraint but used the raw constraint. A
`T extends Record<string, V>` constraint is stored as an unevaluated
`Application` of a mapped type; the downstream indexability query
(`classify_element_indexable`) evaluates with a resolver-less pass that cannot
expand the application, so the constraint looked unindexable and tsz emitted a
false TS7053. The fix evaluates the constraint to its apparent type through
the checker environment (`evaluate_type_with_resolver` with `&self.ctx`) before
the index-signature queries run.

One logic-arm change in `crates/tsz-checker/src/types/utilities/core.rs`, plus a
new registered regression test file. `is_element_indexable` and the solver
classifier are intentionally untouched (broad blast radius — see Coordination
Notes).

## Generalization (matrix in the test file)
- Reported repro: `T extends Record<string, unknown>` indexed by `string` -> clean.
- Renamed param (`U`, `Row`) -> clean (rule is structural, not name-based).
- Value-type variant `Record<string, number>` -> clean.
- Nested wrappers `Partial<Record<...>>`, `Readonly<Record<...>>` -> clean.
- User mapped alias `type MyRec<V> = { [P in string]: V }` -> clean.
- Plain alias to `Record` -> clean.
- Negatives preserved: unconstrained `T`, `T extends { a: number }`,
  still-generic `T extends Record<K, V>`, and the genuine `{} | Record<...>`
  union TS7053 (real tsc 6.0.2 reports it too).

## Project Corpus Impact
- Row: kysely-project (#10663)
- Bug family: false TS7053 implicit-any-index on utility/alias/constrained types
- Evidence: differential testing against real `tsc` 6.0.2
  (`--noImplicitAny --strict`): the reproduced false positives
[...]

PR #10744, opened 2026-05-28
fix(checker): resolve intersection constraint members for TS7053 indexability

### LINKED ISSUE #10726: false TS7053 indexing a type param constrained to an intersection with a Record member
**AgentName:** xeon-4c-16g-web

Found while fixing #10674. Real `tsc` 6.0.2 is clean; `tsz` reports a false `TS7053` when a type parameter is constrained to an **intersection that contains a `Record`-style application member** and is indexed by a `string` key:

```ts
// --noImplicitAny --strict
function i1<T extends ({ a: number } & Record<string, number>)>(x: T, k: string) {
  return x[k]; // tsc: clean ; tsz: false TS7053
}
```

Sibling union cases (`{a:number} | {[s:string]:number}`, etc.) are correctly handled — both tsz and tsc report/accept identically. Only the **intersection-with-application-member** shape diverges.

## Root cause

The indexability decision flows through `is_element_indexable` (checker) →
`classify_element_indexable` (solver, `crates/tsz-solver/src/type_queries/extended.rs`).
For an intersection, `is_element_indexable` recurses per member and re-enters
`classify_element_indexable`, which evaluates with a **resolver-less** pass
(`evaluate_type` with `NoopResolver`). A `Record<string, V>` member is an
`Application` of a mapped type whose base resolves only through the checker's
type environment, so the resolver-less evaluation leaves it as an opaque
`Application` → classified `Other` → not indexable → false `TS7053`.

The #10674 fix resolves the *type-parameter constraint* to its apparent type,
but does not drill into intersection/union **members**, because evaluating the
intersection at top level does not deeply re-evaluate each member's application.

## Why it's a separate PR

A complete fix means making the shared indexability classifier resolver-aware
(either threading a resolver into `classify_element_indexable`, or having the
checker's `is_element_indexable` resolve each member through the environment
before classifying). `is_element_indexable` has many callers across
`indexed_access.rs` and `error_reporter/properties.rs`, so the change has a
broad blast radius and needs its own conformance-gated PR rather than being
bundled into the narrow #10674 fix.

## Repro
`tsz --noImplicitAny --strict --noEmit` on the snippet above; compare with
`node typescript/lib/tsc.js --noImplicitAny --strict --noEmit`.

mohsen1:
## Agent
AgentName: claude-sonnet-4-6 (dazzling-johnson)

## Track
Track 5 — Key-Space, Indexed Access, and Property Semantics  
PR type: semantic campaign / false-positive elimination

## Invariant
When the indexed object is a type parameter and its constraint is an intersection type, `tsc` allows the element access if any intersection member—when resolved through the checker's type resolver—has a string or number index signature that covers the key. This PR makes tsz evaluate each intersection member via `evaluate_type_with_resolver` before deciding the intersection is unindexable.

## Scope
- `crates/tsz-checker/src/types/utilities/core.rs` — one logic arm added in `should_report_no_index_signature`, before the final `!is_element_indexable` call
- `crates/tsz-checker/tests/ts7053_intersection_constraint_index_tests.rs` — new 9-test regression suite
- `crates/tsz-checker/Cargo.toml` — register the new integration test

## Root Cause
`should_report_no_index_signature` fed the intersection type directly to `is_element_indexable`, which calls the solver-level `classify_element_indexable`. That classifier uses `evaluate_type` (no `TypeResolver`), so `Application`/`Lazy` members like `Record<string, V>` remained unresolved and fell through to `Other` → false-positive TS7053.

The fix checks whether the unwrapped `check_type` is an intersection; if so, evaluates each member through `evaluate_type_with_resolver` (which has access to the `TypeEnvironment`) and returns `false` (no diagnostic) when any member is element-indexable.

## Adjacent-case test matrix
1. Reported repro: `T extends { a: number } & Record<string, unknown>` indexed by `string` → no TS7053
2. Renamed type parameter (`Row`, `U`) — rule is structural, not name-based → no TS7053
3. Record as **first** intersection member (order-independence) → no TS7053
4. Concrete value type: `Record<string, number> & { b: boolean }` → no TS7053
5. Inline index signature `{ readonly [key: string]: unknown } & { a: number }` (already worked; regression guard) → no TS7053
6. Number index: `{ length: number } & Record<number, string>` indexed by `number` → no TS7053
7. **Negative**: intersection with no index signature in any member → TS7053 still fires
8. **Negative**: unconstrained `T` → TS7053 still fires
9. **Negative**: `T extends { a: number } & Record<K, V>` (K/V generic) → `contains_type_parameters` guard → no TS7053 (deferred to instantiation, correct behavior)

## Project Corpus Impact
Row: kysely-project  
Bug family: false-positive TS7053 implicit-any-index on type-parameter constrained by an intersection including a utility/index type  
Evidence: Closes #10726. Complementary to #10728 (direct `T extends Record<K,V>` path), which was explicitly documented in #10728's Coordination Notes as needing a separate conformance-gated PR.

## Verification
```
cargo test -p tsz-checker --test ts7053_intersection_constraint_index_tests
# → 9 passed; 0 failed

cargo test -p tsz-checker --test ts7053_unconstrained_type_param_index_tests
cargo test -p tsz-checker --test ts7053_template_mapped_tests
# → all pass; no regressions in adjacent suites

cargo clippy -p tsz-checker --lib -- -D warnings
# → clean
```

## Coordination Notes
- Does NOT overlap with #10728 (direct Record constraint path) — the two are complementary: #10728 handles `T extends Record<K,V>` directly; this PR handles `T extends X & Record<K,V>` (intersection constraints).
- Does NOT require changes to `classify_element_indexable` or `is_element_indexable` — the fix is narrowly scoped to the intersection case in `should_report_no_index_signature`, avoiding the "broad blast radius" noted in #10728's Coordination Notes.
- Checked open draft and ready PRs before branching; no overlap found.
- Stack: no dependencies on other in-flight branches.

Signed: claude-sonnet-4-6 (dazzling-johnson)

---
_Generated by [Claude Code](https://claude.ai/code/session_01YYs5oVrvsmtZJMhJTBGVGZ)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -1587,6 +1587,10 @@ name = "ts7053_unconstrained_type_param_index_tests"
 path = "tests/ts7053_unconstrained_type_param_index_tests.rs"
 
 [[test]]
+name = "ts7053_intersection_constraint_index_tests"
+path = "tests/ts7053_intersection_constraint_index_tests.rs"
+
+[[test]]
 name = "tuple_union_contextual_typing_tests"
 path = "tests/tuple_union_contextual_typing_tests.rs"
 
--- a/crates/tsz-checker/src/types/utilities/core.rs
+++ b/crates/tsz-checker/src/types/utilities/core.rs
@@ -2581,6 +2581,31 @@ impl<'a> CheckerState<'a> {
             return true;
         }
 
+        // When the constraint is an intersection, evaluate each member through the
+        // checker's type resolver before checking indexability. Intersection members
+        // may be `Application`/`Lazy` types (e.g. `Record<string, V>`) that
+        // `classify_element_indexable` cannot expand without a resolver — the
+        // solver-level `evaluate_type` it uses has no access to the `TypeEnvironment`.
+        // Structural rule: if any intersection member, when resolved through the
+        // checker's resolver, has a string or number index signature that covers
+        // the key, tsc allows the access and tsz must agree.
+        if let Some(members) =
+            crate::query_boundaries::common::intersection_members(self.ctx.types, unwrapped_type)
+        {
+            let any_member_indexable = members.iter().any(|&member| {
+                let resolved =
+                    crate::query_boundaries::state::type_environment::evaluate_type_with_resolver(
+                        self.ctx.types,
+                        &self.ctx,
+                        member,
+                    );
+                self.is_element_indexable(resolved, wants_string, wants_number)
+            });
+            if any_member_indexable {
+                return false;
+            }
+        }
+
         !self.is_element_indexable(unwrapped_type, wants_string, wants_number)
     }
 
--- a/crates/tsz-checker/tests/ts7053_intersection_constraint_index_tests.rs
+++ b/crates/tsz-checker/tests/ts7053_intersection_constraint_index_tests.rs
@@ -1,0 +1,185 @@
+//! TS7053 — indexing a type parameter constrained to an intersection that
+//! includes a member with an index signature.
+//!
+//! Structural rule: when the indexed object is a type parameter and its
+//! constraint is an intersection type, tsc allows the access if any
+//! intersection member—when resolved through the checker's type resolver—has
+//! a string or number index signature that covers the key. tsz must evaluate
+//! each intersection member through the resolver before deciding the
+//! intersection is unindexable.
+//!
+//! Issue: <https://github.com/mohsen1/tsz/issues/10726>
+//!
+//! Complementary to #10728 (direct `T extends Record<K,V>` constraint), which
+//! fixes the same resolver gap for the non-intersection path.
+
+use tsz_checker::context::{CheckerOptions, ScriptTarget};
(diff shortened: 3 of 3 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Closing as superseded by #10754 per user request to remove redundant PRs.

Evidence:
- Both PRs target the same TS7053 intersection-constraint indexability family (`T extends { ... } & Record<...>` and resolver-aware classification of intersection members).
- #10754 is the stronger active implementation: it threads the checker resolver through `classify_element_indexable`, preserves index-signature-carrying intersection members during simplification, and has `CI Summary` / `conformance-aggregate` green.
- #10744 is draft and currently fails light CI/lint, while touching the same core checker utility path and adding overlapping tests.
- Preserved branch/commit for reference: `claude/dazzling-johnson-Fm1lB` @ `b1ce9518533541116b6e7fd3ed12065a84e9e109`.

Useful work preserved: #10744's focused intersection test matrix is largely represented in #10754's 17-test suite; any missing names/order variants can be added there if needed.

PR #10732, opened 2026-05-28
chore(conformance): drop stale accepted-regression entries

mohsen1:
AgentName: M5Opus

## Track
Conformance hygiene (resolves #10654, #10655, #10350, #10443; trims aggregate #10306).

## Invariant
`conformance-accepted-regressions.txt` is a temporary runway for *failing* tests. Entries whose tests now pass are stale and must be removed (documented "stale-accepted" workflow in `scripts/conformance/link-regression-issues.py`).

## Scope
Remove 10 stale entries that `link-regression-issues.py` classifies as "stale: accepted-regression entry no longer fails" (each is `PASS` in `conformance-baseline.txt`; current snapshot `conformance-detail.json` = 12582/12582, 0 failures, 0 known failures):
declarationsWithRecursiveInternalTypesProduceUniqueTypeParams, errorInfoForRelatedIndexTypesNoConstraintElaboration, reorderProperties, circular2, circular4, declarationFileForHtmlImport, keyofAndIndexedAccess, keyofAndIndexedAccessErrors, mappedTypeRelationships, unionAndIntersectionInference1.

Retained the 3 perf/relation-depth-sensitive entries (deeplyNestedMappedTypes, intersectionsOfLargeUnions, intersectionsOfLargeUnions2) whose runway guards against CI timeout variance — left for a separate perf-focused removal rather than risk timeout flake here.

Since conformance has no deficit (12582 ≥ floor), the allowlist is not consulted by the aggregate gate, so removing passing entries is safe; the ready-CI full conformance suite verifies it.

## Project Corpus Impact
- Row: n/a
- Bug family: conformance accepted-regression staleness (entries that no longer fail)
- Evidence: `link-regression-issues.py` stale-accepted classification + `conformance-detail.json` 12582/12582.

## Verification
- `python3 scripts/conformance/link-regression-issues.py` reports all removed entries as "stale: accepted-regression entry no longer fails".
- Ready-CI conformance suite is the authoritative gate.

## Coordination Notes
Signed M5Opus. Resolves #10654, #10655, #10350, #10443. Complements the recent jsxComplexSignatureHasApplicabilityError removal on main.

PR #10765, opened 2026-05-28
fix(solver): defer large O[T] evaluation when T extends keyof O

### LINKED ISSUE #10654: conformance: related index type diagnostic elaboration accepted regression
AgentName: M1-C

## Summary

Create a narrow owner for the accepted-regression path `errorInfoForRelatedIndexTypesNoConstraintElaboration.ts`, which was previously only covered by aggregate tracker #10306.

## Retained Path

- `TypeScript/tests/cases/compiler/errorInfoForRelatedIndexTypesNoConstraintElaboration.ts`

## Current Evidence

- Aggregate tracker: #10306.
- Current accepted-regression artifact still lists this path in `scripts/conformance/conformance-accepted-regressions.txt`.
- Current checked-in dashboard reports exact diagnostic conformance (`12582/12582`) while accepted-regression strictness still retains this path, so removal needs focused CI/artifact evidence or a structural fix.
- Focused local evidence from this M1-C cycle:
  - `scripts/safe-run.sh --limit 50% -- ./scripts/conformance/conformance.sh run --filter errorInfoForRelatedIndexTypesNoConstraintElaboration --verbose`
  - Result: the harness rebuilt successfully, then the filtered `tsz --project` child ran for more than two minutes without producing a conformance result; M1-C interrupted it to avoid waiting indefinitely. Treat the exact expected/actual diagnostic tuple as still pending artifact work.

## Suspected Owner Layer

Solver relation/indexed-access comparison and structured relation failure reasons, with checker diagnostics rendering the solver-provided reason. The fix should not inspect rendered type strings, source text, file names, or this conformance test name.

## Structural Rule To Prove

When two related indexed-access types are compared and their constraints do not justify constraint elaboration, `tsc` reports the stable relation diagnostic without expanding into misleading constraint details. tsz should choose that diagnostic through solver/query-boundary relation failure facts rather than checker-local display heuristics.

## Adjacent Cases To Cover

- Indexed-access relation checks where both sides share a key constraint but differ in the indexed value.
- Generic indexed-access comparisons where one side has no useful constraint elaboration.
- A positive assignability case that should not emit a relation diagnostic.
- A negative case where constraint elaboration is valid and should still be rendered.

## Next Step

Recover exact expected vs actual diagnostics from a ready-review artifact or a reduced targeted repro, then add solver/query-boundary coverage before removing this path from the accepted-regression list.

mohsen1:
AgentName: vibrant-lovelace-opus47 (remote execution / `claude/vibrant-lovelace-bWX2G`)

Closes #10654.

## Track
Conformance / solver — performance regression cleanup for the indexed-access
relation path (`errorInfoForRelatedIndexTypesNoConstraintElaboration.ts`).

## Invariant
When the index of an `IndexAccess` is a type parameter whose effective
constraint is structurally `keyof O` and `O` has many properties, the
evaluator keeps the access deferred. This matches tsc's `getIndexedAccessType`
semantics for generic indexes and preserves the per-call-site type-parameter
identity that the TS2322 + TS5075 elaboration chain relies on
(`assignability_relation::is_assignable_to`, pre-evaluation key-identity
rejection at lines 456-480).

## Scope
- `crates/tsz-solver/src/evaluation/evaluate_rules/index_access.rs`:
  In the `IndexAccessVisitor`, add the structural check
  `index_is_type_param_constrained_by_keyof_of_this_object` (which walks the
  index's type-parameter constraint, also peering through intersections, to
  recognize a `KeyOf(X)` whose `X` is the object being indexed, modulo
  evaluation). Have `visit_object` and `visit_object_with_index` return
  `None` to defer when the shape has enough properties to make eager
  expansion expensive (`LARGE_OBJECT_DEFERRAL_THRESHOLD = 60`). The threshold
  is documented in-place; below it, eager expansion is preserved so
  downstream property/method lookup on shapes like `Array<O[T]>.push`
  continues to work.
- `crates/tsz-checker/tests/deferred_keyof_index_access_assignability_tests.rs`:
  Add five new unit tests: (a) a JSX.IntrinsicElements-scale synthetic (80
  keys × DetailedHTMLProps-style generic value types) that previously timed
  out >120 s and now completes in milliseconds; (b) the same with three
  different type-parameter spellings (T1/T2, Key1/Key2, Alpha/Beta — §25
  anti-hardcoding); (c) identity `Things[K] -> Things[K]` accepts (negative,
  at 80 keys); (d) at-scale rejection of `{}` against a value type with a
  required prop (negative, at 80 keys — deferral does not over-relax); (e)
  coverage for the `visit_object_with_index` site (object with a string
  index signature, at 80 keys).

## Why this layer
The eager expansion of `O[T]` is a `WHAT` decision: type evaluation. The
deferral lives in the solver's `IndexAccessVisitor`, alongside the existing
`try_mapped_type_param_substitution` precedent at the top of
`evaluate_index_access`. The new structural check reuses existing solver
helpers (`type_param_info`, `intersection_list_id`, `keyof_inner_type`,
`constraints_semantically_match`) instead of duplicating identity rules.

## Adjacent-case matrix (covered by tests)
- Reported repro: TypeScript conformance test
  `errorInfoForRelatedIndexTypesNoConstraintElaboration.ts` with the full
  `react16.d.ts` (`JSX.IntrinsicElements` with ~150 keys). Before: >120 s
  timeout in CI conformance. After: ~3 s end-to-end, emits the expected
  TS2322 + TS5075 elaboration matching tsc's baseline.
- 80-key synthetic shape (above threshold): timing assertion in test body
  ensures the deferral path stays fast.
- 80-key synthetic shape with three different parameter spellings: proves
  the rule is not name-bound.
- Identity `Things[K] -> Things[K]` at 80 keys: still accepts.
- 80-key shape with a single required-prop value: still rejects.
- 80-key object-with-index-signature shape: distinct generic keys still emit
  TS2322 (exercises the second deferral site).
- Below-threshold shapes (existing tests at 2–3 keys): eager expansion
  preserved; property/method lookup on `Array<O[T]>` still works
  (regression check via the pre-existing
  `mapped_type_generic_indexed_access_class_member` test in
  `crates/tsz-checker/tests/ts2322_tests`).

## Project Corpus Impact
- Row: conformance — `errorInfoForRelatedIndexTypesNoConstraintElaboration.ts`
  removed from the accepted-regression runway. PR #10732 previously tried
  to remove this entry as "stale"; the closing comment confirmed it was a
  genuine CI regression. This PR addresses the underlying performance issue.
- Bug family: non-termination on indexed access with generic key constrained
  by `keyof O` for large `O`.
- Evidence: local `dist-fast` tsz on the reduced JSX repro (~150 keys, two
  indexed accesses with distinct generic keys) drops from >120 s
  (terminated) to 3.07 s and emits the expected TS2322 + TS5075 chain
  matching tsc's
  `errorInfoForRelatedIndexTypesNoConstraintElaboration.errors.txt`. The
  full CI conformance suite is the authoritative gate.

## Verification
- `cargo test -p tsz-checker --test deferred_keyof_index_access_assignability_tests`
  → 15/15 pass (10 pre-existing + 5 new).
- Reduced JSX repro on `dist-fast` tsz: ~3 s end-to-end (across three runs)
  vs. >120 s without the fix.
- `cargo test -p tsz-checker --test ts2322_tests`: identical pre-existing
  failure set (2 mapped-application-correlation tests, both failing on
  `origin/main`); no new regressions from this PR. The previously-passing
  `mapped_type_generic_indexed_access_class_member` test still passes (this
  was the regression caught in round 1 of review, fixed by the size gate).
- `cargo test -p tsz-solver --lib`: identical pre-existing failure set (21
  local-env failures unrelated to this PR — `Equal<any>` conditionals,
  template-`any` widening, file-size ceiling); no new regressions.
- `cargo clippy -p tsz-solver --lib --all-features --release`: clean.
- Heavy CI suites (conformance-aggregate, emit, fourslash, WASM) will run on
  ready.

## Coordination Notes
Signed vibrant-lovelace-opus47. Issue #10654 was previously linked to PR
#10732 which tried to drop the accepted-regression entry as "stale"; that
diagnosis was wrong — the test was genuinely failing (timing out) on CI.
This PR addresses the underlying performance issue so the test can pass
without the accepted-regression escape hatch.

The fix went through three rounds of `/simplify` review:
- Round 1 surfaced reuse opportunities (now uses existing `type_param_info`,
  `intersection_list_id`, `keyof_inner_type`, `constraints_semantically_match`
  helpers) and a regression in the mapped-type generic indexed-access path
  (now gated by object size).
- Round 2 surfaced a test-coverage gap (three tests at 30–40 keys did not
  exercise the new gating path — now at 80 keys) and absence of a
  `visit_object_with_index` test (now added).

The size-gating threshold of 60 is documented in-place. Below the threshold
the eager expansion is preserved so property/method lookup on shapes like
`Array<O[T]>.push` still resolves; above it, the deferral matches tsc.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VFFUCMWBK7BjPGhoYQFBXY)_

### DIFF
--- a/crates/tsz-checker/tests/deferred_keyof_index_access_assignability_tests.rs
+++ b/crates/tsz-checker/tests/deferred_keyof_index_access_assignability_tests.rs
@@ -410,16 +410,17 @@ class Holder<T1 extends keyof Things, T2 extends keyof Things> {{
 /// directive §25): the structural rule is "for `O[A] -> O[B]` with A != B both
 /// generic type parameters constrained by `keyof O`, emit TS2322 with the
 /// stable elaboration." A name-bound fix would silently regress when callers
-/// pick a different spelling.
+/// pick a different spelling. Uses 80 keys — above the large-object deferral
+/// threshold — so the deferral path is exercised across name choices.
 #[test]
 fn large_object_with_generic_value_types_two_names() {
     for (p1, p2) in [("T1", "T2"), ("Key1", "Key2"), ("Alpha", "Beta")] {
         let mut things_body = String::new();
-        for i in 0..40 {
+        for i in 0..80 {
             things_body.push_str(&format!("    k{i}: Detailed<Base<E{i}>, E{i}>;\n"));
         }
         let mut elements_body = String::new();
-        for i in 0..40 {
+        for i in 0..80 {
             elements_body.push_str(&format!("interface E{i} {{}}\n"));
         }
         let source = format!(
@@ -455,11 +456,13 @@ class Holder<{p1} extends keyof Things, {p2} extends keyof Things> {{
 
 /// Negative companion: same shape but the source and target use the *same*
 /// type-parameter identity — `Things[T] -> Things[T]` is identity, must not
-/// emit TS2322 regardless of how many keys `Things` has.
+/// emit TS2322 regardless of how many keys `Things` has. Sized above the
+/// large-object deferral threshold to verify the deferral does not reject
+/// the identity case.
 #[test]
 fn large_object_same_key_param_identity_accepts() {
     let mut things_body = String::new();
-    for i in 0..30 {
+    for i in 0..80 {
         things_body.push_str(&format!("    k{i}: {{ id?: string }};\n"));
     }
     let source = format!(
@@ -489,10 +492,11 @@ class Holder<K extends keyof Things> {{
 /// Negative companion: `{} -> Things[K]` must still detect rejection when at
 /// least one key's value type has a required property, even at large scale.
 /// The fix targets evaluation deferral, not the rejection rule itself.
+/// Sized above the large-object deferral threshold.
 #[test]
 fn large_object_with_required_member_still_rejects_empty_object() {
     let mut things_body = String::new();
-    for i in 0..29 {
+    for i in 0..79 {
         things_body.push_str(&format!("    k{i}: {{ a?: string }};\n"));
     }
     // One key with a REQUIRED prop; the rule must still fire.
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: M4-A

Merge-sweep coordination note: exact-head PR checks for `20efdf4d3fbaaa05174e78b74760587712cb800f` are green (`CI Summary` and `GitGuardian Security Checks` passed), the PR is ready/not draft and has no WIP markers. Per the user-requested merge-first pass, I am adding `merge-queue` so the repo-local queue can produce `Queue Tested` and land it.

PR #11655, opened 2026-05-29
fix(dts): guard direct-call return inference against self-recursive function bodies

mohsen1:
AgentName: m5-pro-48g-gpt5

## Original Agent
claude-sonnet-4-6 (remote execution, session dazzling-johnson-CboCN)

## Track
Emit / DTS

## Invariant
`direct_call_primitive_return_type_text` must not cause infinite recursion
when it encounters a self-recursive callee during declaration type inference.

### Root Cause

`direct_call_primitive_return_type_text` has a `depth > 8` guard, but that
guard is silently bypassed because `function_body_preferred_return_type_text`
(called via `.or_else(...)` at the end of the body-analysis branch) resets
depth to **0** when it internally calls
`declaration_summary_primitive_expression_type_text`. For a self-recursive
callee the cycle is:

```
direct_call_primitive_return_type_text(d)
  → declaration_summary_primitive_expression_type_text(d+1)
    → ... infer_fallback_type_text_at ...
      → direct_call_primitive_return_type_text(d+2)
        ...
          // depth exceeds 8, returns None
        → function_body_preferred_return_type_text  ← resets depth to 0!
          → declaration_summary_primitive_expression_type_text(0)
            → direct_call_primitive_return_type_text(1)   ← infinite loop
```

### Fix

Mirror the existing guard already present in `call_expression_source_return_type_text`
(see `type_inference_source_callables.rs` lines 450-455): skip body analysis when
`source_function_body_contains_direct_call_to_name` detects the callee directly calls
itself by name. Depth alone is insufficient when intermediate helpers reset the counter;
the name-based text scan is the correct structural fix.

### Structural rule

> When a function body directly calls itself by name, skip body analysis in
> `direct_call_primitive_return_type_text` to prevent infinite recursion —
> exactly as `call_expression_source_return_type_text` already does.

### Files changed

- `crates/tsz-emitter/src/declaration_emitter/helpers/type_inference_fallback_types.rs`:
  add `!source_function_body_contains_direct_call_to_name(...)` condition to the
  `else if func.body.is_some()` branch in `direct_call_primitive_return_type_text`
- `crates/tsz-emitter/src/declaration_emitter/tests/misc_features.rs`:
  three new regression tests (see below)

## Project Corpus Impact

- Row: n/a
- Bug family: infinite-recursion / stack-overflow in DTS return-type inference
- Evidence: crash-prevention fix in `direct_call_primitive_return_type_text`; all three new unit tests pass; no existing `tsz-emitter` tests regressed

## Verification

Three regression tests covering distinct structural shapes:

1. **Shadowed inner function repro** — outer `pick` has `T extends () => infer R ? R : never`
   return; inner `pick` is self-recursive. Guard fires, `wrap`'s DTS does not contain `infer R`.
2. **Directly self-recursive exported function** — `loop(n)` and `identity(s)` call themselves;
   both appear in DTS without a stack overflow.
3. **Non-recursive callee happy path** — `helper()` returns `"hello"`, `wrapper()` calls it;
   `wrapper` is inferred as `string`, confirming the guard does not regress the
   non-recursive path.

```
test declaration_emitter::tests::misc_features::test_exported_function_returning_shadowed_helper_does_not_borrow_top_level_return_type ... ok
test declaration_emitter::tests::misc_features::test_self_recursive_exported_function_does_not_stack_overflow ... ok
test declaration_emitter::tests::misc_features::test_non_recursive_callee_still_infers_primitive_return ... ok
```

## Coordination Notes

- Independent of PR #11647 (decorated-class constructor lowering, now merged); no shared files.
- `source_function_body_contains_direct_call_to_name` is a text-scan helper
  already used by `call_expression_source_return_type_text`; no new utility added.
- Rebased onto main after #11647 merged.

### DIFF
--- a/crates/tsz-emitter/src/declaration_emitter/helpers/type_inference_fallback_types.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/helpers/type_inference_fallback_types.rs
@@ -782,7 +782,13 @@ impl<'a> DeclarationEmitter<'a> {
                     }
                     other => other,
                 }
-            } else if func.body.is_some() {
+            } else if func.body.is_some()
+                && !self.source_function_body_contains_direct_call_to_name(
+                    self.arena,
+                    func,
+                    &symbol.escaped_name,
+                )
+            {
                 self.function_body_single_return_expression(func.body)
                     .and_then(|return_expr| {
                         self.declaration_summary_primitive_expression_type_text(
--- a/crates/tsz-emitter/src/declaration_emitter/tests/misc_features.rs
+++ b/crates/tsz-emitter/src/declaration_emitter/tests/misc_features.rs
@@ -406,6 +406,69 @@ export function wrap<T>(value: T) {
     );
 }
 
+#[test]
+fn test_self_recursive_exported_function_does_not_stack_overflow() {
+    // A directly self-recursive exported function must not cause infinite
+    // recursion in the return-type inferrer; it should fall back without crashing.
+    let output = emit_dts_with_usage_analysis(
+        r#"
+export function loop(n: number): number {
+    return loop(n - 1);
+}
+export function identity(s: string): string {
+    return identity(s);
+}
+"#,
+    );
+    assert!(
+        output.contains("declare function loop"),
+        "Self-recursive loop must appear in declarations: {output}"
+    );
+    assert!(
+        output.contains("declare function identity"),
+        "Self-recursive identity must appear in declarations: {output}"
+    );
+}
+
+#[test]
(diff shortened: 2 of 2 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Updated the PR body to satisfy the current ready-PR metadata gate (`AgentName:` plus `Project Corpus Impact` bullet fields were verified in the remote body). The first rerun request landed while CI was still active; now that run `26646594870` has completed, I reran failed jobs again.

PR #11679, opened 2026-05-29
fix(checker): stable TypeId for class type params to eliminate false TS2349

mohsen1:
## AgentName
claude-sonnet-4-6-dazzling-johnson

## Track
Checker / Solver correctness — false-positive TS2349 elimination

## Invariant
`{ [K in keyof T]: V }[K]` must evaluate to `V` when `K extends keyof T`, regardless of whether the mapped type appears as a class instance field or a standalone function parameter.

## Scope

**Structural rule being implemented:**
> When the same class type parameter declaration node is processed by multiple independent `push_type_parameters` calls (from `check_class_declaration` and `get_class_instance_type_inner`), tsc produces consistent type identity for that parameter. This change makes tsz do the same by adding a node-keyed cache for type parameters that lack `DefId` registration.

**Root cause:**
Class, method, and interface type parameters are not registered in `semantic_defs` by the binder (`SemanticDefKind` has no `TypeParameter` variant; `bind_type_parameters` does not call `record_semantic_def`). As a result, `intern_type_param_for_decl` had no `DefId`-keyed cache entry for them. It fell back to `fresh_type_param` on every call, minting a new `TypeId` each time.

Two independent `push_type_parameters` calls occur for every class:
1. `check_class_declaration` → creates `T_id_check`
2. `get_class_instance_type_inner` → creates `T_id_instance` (different!)

The mapped type stored `KeyOf(T_id_instance)` as its constraint. The method type parameter `K` stored `KeyOf(T_id_check)` as its constraint. `type_param_constraint_matches` in `visit_mapped` compared these two and found them unequal, so the index access `this.map[key]` was left as a deferred `IndexAccess` rather than evaluating to `() => void`. The call site then emitted a false TS2349.

The standalone function form (`function test(map: {...}, key: K)`) was not affected because both the map parameter type and `K`'s constraint share the same `push_type_parameters` scope in a single call to `check_function_declaration`.

**Fix:**
Added `type_param_node_cache: FxHashMap<(u32, TypeParamInfo), TypeId>` to `CheckerContext`. In `intern_type_param_for_decl`, when no `DefId` is available, this node-keyed cache is consulted before calling `fresh_type_param`. The key is `(name_node.0, info)` — the declaration node's raw u32 paired with the full `TypeParamInfo` so the constraint-refinement second pass still creates a new (constrained) entry, exactly mirroring the def-based path.

**Files changed:**
- `crates/tsz-checker/src/context/mod.rs` — new `type_param_node_cache` field
- `crates/tsz-checker/src/context/constructors.rs` — initialize to `FxHashMap::default()`
- `crates/tsz-checker/src/state/type_analysis/core.rs` — node-cache fallback in `intern_type_param_for_decl`
- `crates/tsz-checker/tests/class_field_mapped_type_indexed_access_tests.rs` — 5 new regression tests
- `crates/tsz-checker/Cargo.toml` — register new test target
- `crates/tsz-solver/tests/index_access_comprehensive_tests.rs` — additional solver-level coverage
- `scripts/conformance/conformance-accepted-regressions.txt` — remove `keyofAndIndexedAccess.ts` (now passes)

## Project Corpus Impact
- Row: checker correctness — `{[K in keyof T]: V}[K]` index access resolution
- Bug family: false TS2349 on mapped-type class field indexed access
- Evidence: All 5 new tests in `class_field_mapped_type_indexed_access_tests.rs` pass locally; `keyofAndIndexedAccess.ts` removed from accepted regressions. Conformance gate runs in CI on ready-for-review.

## Verification
```
cargo check -p tsz-checker  # clean
```
Rebased onto main (d31d452) on 2026-05-29 — no conflicts.

Adjacent shapes covered by the test matrix (per §26 generalization gate):
1. `C1` with `map: {[K in keyof T]: () => void}`, `test` — core repro
2. `Box` with `handlers: {[P in keyof U]: () => void}`, `run<p>` — renamed params
3. Standalone `function test(map: {...}, key: K)` — pre-existing passing form
4. `Strings` with `data: {[K in keyof T]: string}`, cast to `any` — soundness/non-callable
5. `Multi` with `actions: {[K in keyof A]: () => B}` — multi-param class

## Coordination Notes
- No open PRs touch `intern_type_param_for_decl`, `push_type_parameters`, `type_param_node_cache`, or `get_class_instance_type_inner`.
- This does not change `SemanticDefKind` or the binder — that would be a larger structural change. The cache fix is minimal and localized to the checker's context.
- `keyofAndIndexedAccessErrors.ts` remains in accepted regressions — it has a fingerprint-only failure (error codes match, positions/messages differ) unrelated to this TS2349 fix.
- Conformance suite runs in CI when PR is marked ready.

AgentName: claude-sonnet-4-6-dazzling-johnson

https://claude.ai/code/session_013KokZDEMrUPeFKbmPFPha7

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -1591,6 +1591,10 @@ name = "string_mapping_alias_display_tests"
 path = "tests/string_mapping_alias_display_tests.rs"
 
 [[test]]
+name = "class_field_mapped_type_indexed_access_tests"
+path = "tests/class_field_mapped_type_indexed_access_tests.rs"
+
+[[test]]
 name = "class_arrow_property_type_inference_tests"
 path = "tests/class_arrow_property_type_inference_tests.rs"
 
--- a/crates/tsz-checker/src/context/checker_context_lifetimes.toml
+++ b/crates/tsz-checker/src/context/checker_context_lifetimes.toml
@@ -246,3 +246,4 @@ js_body_uses_arguments = { lifetime = "DiagnosticsOnly", capability = "Diagnosti
 emitted_ts2454_errors = { lifetime = "SpeculationScoped", capability = "SpeculationState", reason = "snapshot/rollback-sensitive state mutated during speculative checking" }
 emitted_ts2411_for_iface_prop = { lifetime = "DiagnosticsOnly", capability = "DiagnosticState", reason = "diagnostic emission, dedupe, or suppression state; clear at file-session boundary" }
 type_resolution_fuel = { lifetime = "FileLocalReset", capability = "RelationSessionState", reason = "checker-local mutable state tied to the active file or traversal; reset at file-session boundary" }
+type_param_node_cache = { lifetime = "FileLocalReset", capability = "FileTypeCache", reason = "node-keyed TypeId cache for class/method/interface type params lacking DefId registration; ensures push_type_parameters convergence within a file session" }
--- a/crates/tsz-checker/src/context/constructors.rs
+++ b/crates/tsz-checker/src/context/constructors.rs
@@ -301,6 +301,7 @@ impl<'a> CheckerContext<'a> {
             emitted_ts2411_for_iface_prop: FxHashSet::default(),
             type_resolution_fuel: Cell::new(crate::state::MAX_TYPE_RESOLUTION_OPS),
             typeof_resolution_stack: RefCell::new(FxHashSet::default()),
+            type_param_node_cache: FxHashMap::default(),
         }
     }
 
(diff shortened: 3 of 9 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Rebased this PR onto current `origin/main` and pushed `a8d8ad9129`. The rebase conflict was limited to `scripts/conformance/conformance-accepted-regressions.txt`; I preserved both removals now present across main/this PR (`parserSymbolIndexer5.ts` remains removed from main, and this PR continues removing `keyofAndIndexedAccess.ts`).

Verification after the rebase:
- `python3 scripts/arch/arch_guard.py` passed

Fresh CI is running on the rebased head. Next action: queue only after the new head has a clean complete check picture.

— m5-pro-48g-gpt5

PR #11661, opened 2026-05-29
test(checker): lock generic-default branch independence

mohsen1:
## Summary

Adds regression coverage proving that generic type-parameter **defaults stay
scoped to their own declaration and instantiation** and never "bleed" between
independent branches. This pins the behavior tracked by the `solver-29-20`
"generic defaults bleed between independent branches" family (#11608, with
siblings #11589 / #11487).

Investigation found the reported divergence is **not reproducible** — tsz
already matches `tsc` across every shape I probed (basic defaults, earlier-param
references, `keyof` defaults, shadowed inner type params, forward references,
function generics, class generics). Rather than leave the invariant unguarded,
this PR converts the report into durable regression tests at two altitudes.

**Structural rule under test:** when a generic declaration has a defaulted type
parameter, each instantiation resolves that default against *its own* supplied
arguments only; a default resolved for one instantiation (or for a declaration
that merely shares a type-parameter name) is never reused for an independent
instantiation/declaration.

## Track
Conformance / regression-prevention (no behavior change).

## Invariant
Generic-default resolution is per-declaration and per-instantiation; defaults do
not bleed across independent branches. `TypeSubstitution::from_args` builds a
fresh substitution per call (no shared/cached default state).

## Scope
- `crates/tsz-checker/tests/generic_default_branch_independence_tests.rs` (new):
  7 end-to-end cases over the full AST → binder → solver → checker pipeline —
  independent aliases, earlier-param defaults (`B = A`) per instantiation,
  `keyof` defaults, shadowed inner type params (no capture), function-generic
  defaults per call, class-generic defaults, and a negative guard.
- `crates/tsz-checker/Cargo.toml`: register the new test target (crate uses
  `autotests = false`).
- `crates/tsz-solver/tests/instantiate_tests_parts/part_00.rs`: one low-level
  `from_args` unit test proving `B = A` resolves independently across calls.

Per the anti-hardcoding directive (§25), cases vary type-parameter spellings
(`T`/`K`, `A`/`B`, `P`/`Q`) so a name-keyed implementation would fail.

## Project Corpus Impact
- Row: type-challenges-solutions-project (case solver-29-20)
- Bug family: generic-default branch independence (defaults bleeding across instantiations)
- Evidence: tests-only; no compiler code changed. New tests pass locally and no project-corpus behavior changes (the behavior was already correct; this only locks it in).

## Verification
- `cargo test -p tsz-checker --test generic_default_branch_independence_tests` → 7 passed.
[...]

Issue #11589, opened 2026-05-28
nextjs: solver generic defaults bleed between independent branches (case solver-29-20)


mohsen1:
## Problem

Benchmark row `nextjs` shows a mismatch in `solver` for `solver generic defaults bleed between independent branches` (`solver-29-20`).

## Target surface

- Row: `nextjs`
- File/module: case fixture within `nextjs` project row
- Boundary: AST → binder/solver/checker pipeline

## Reproduction

1. Create a minimal fixture from the affected `nextjs` case.
2. Compare tsz output against TypeScript (`tsc`) for the same fixture.
3. Capture compile invocation and any diagnostic trace.

## Expected

- Type behavior matches TypeScript (`tsc`) for the same row fixture.

## Current tsz behavior

- tsz diverges in diagnostics/acceptance for this case.

## Root-cause hypothesis

- Likely a boundary regression in binder-to-checker handoff or relation-graph memoization for this case family.

## Fix plan

1. Add a reduced benchmark fixture for `nextjs` reproducer.
2. Add targeted checker/syntax/conformance regression under the corresponding family.
3. Validate no spillover regressions into adjacent rows.
4. Verify with the full project-bench row run.

## Acceptance criteria

- [ ] Minimal repro exists for `nextjs`.
- [ ] `tsc` and tsz outputs agree for this case.
- [ ] Conformance/fourslash or solver/bench fixture added.
- [ ] Adjacent row stability preserved.

mohsen1:
Validation note: I reviewed this issue and validated the report shape.

- Current TS target in tsz is scripts/conformance/typescript-versions.json -> 6.0.3, so the comparison point should be TS 6 behavior.
- Case family: solver-29-20.
- Boundary: AST → binder/solver/checker pipeline, target surface: `nextjs`.
- File/module hint: case fixture within `nextjs` project row.
- This issue appears structurally valid and includes enough context for triage.
- Next validation step: reproduce under TS 6.0.3 (tsc + tsz harness for this fixture) and attach concrete divergence before assigning priority.
- If this is now fixed on TS 6.0.3 behavior, post evidence and close as resolved.
Related issue family: nextjs.

mohsen1:
Validation pass (TS6 parity): keeping open for now.

- TS target checked for conformance: 6.0.3 ().
- Family tag: solver-29-20.
- This reads as a potentially valid TS6 parity issue unless we can prove the same behavior already matches.
- For a definitive close, please attach:
[...]

PR #11742, opened 2026-05-29
fix(solver): treat protected members as hierarchical in assignability

### LINKED ISSUE #11631: vite-vanilla-ts-app: binder drops augmentation visibility for private and protected mirrors (case binder-11-20)
## Problem

Benchmark row `vite-vanilla-ts-app` shows a mismatch in `binder` for `binder drops augmentation visibility for private and protected mirrors` (`binder-11-20`).

## Target surface

- Row: `vite-vanilla-ts-app`
- File/module: case fixture within `vite-vanilla-ts-app` project row

## Reproduction

1. Create a minimal fixture from the affected `vite-vanilla-ts-app` case.

## Expected

## Current tsz behavior

## Root-cause hypothesis

## Fix plan

1. Add a reduced benchmark fixture for `vite-vanilla-ts-app` reproducer.

## Acceptance criteria

- [ ] Minimal repro exists for `vite-vanilla-ts-app`.

mohsen1:
## Summary

Fixes a false-positive `TS2322` where tsz rejected assigning a subclass to its
base when the subclass widened an inherited `protected` member to `public`
(legal in TypeScript). Closes #11631 (`binder-11-20` "augmentation drops
private/protected mirror visibility" family).

Verified against `tsc` 6.0.2: tsc accepts the assignment and only reports the
later out-of-class access (`TS2445`); tsz previously emitted a spurious
`TS2322` on the assignment itself.

## Structural rule

> When a target property is `protected`, a source object is assignable as long
> as the source property is declared in the target's protected-declaring class
> *or in a class derived from it*; the source may legally widen the member from
> `protected` to `public`. `private` is unchanged — it still requires the exact
> same declaration (declaration identity). Mirrors tsc's
> `isPropertyInClassDerivedFrom` / `isValidOverrideOf`.

## Track

Conformance / TS2322 parity (checker/solver assignability).

## Invariant

Protected-member assignability is hierarchical (resolved through the nominal
class inheritance graph), not declaration-identical. Private-member
assignability remains strictly nominal. A single helper
`SubtypeChecker::nominal_member_origin_ok(source_parent, target_parent,
target_visibility)` is the one source of truth, shared by every nominal
property-compatibility decision.

## Scope

- `crates/tsz-solver/src/relations/subtype/rules/objects.rs`: add
  `nominal_member_origin_ok`; route the structural property check, the
  fast-fail prepass, and the with-index-signature object path through it.
- `crates/tsz-solver/src/relations/subtype/explain.rs`: route both
  failure-reason decision sites through the same helper so the diagnostic path
  agrees with the relation result.
- `crates/tsz-solver/src/relations/compat_overrides.rs`: route the
  nominal-brand override's target loop through the same helper (this path is
  reached by the checker first and would otherwise short-circuit reject).
- Tests: `crates/tsz-checker/tests/class_protected_widening_assignability_tests.rs`
  (9 end-to-end cases) and two solver-level unit tests in
  `crates/tsz-solver/tests/lawyer_tests.rs` that wire an `InheritanceGraph`.

Per §25 the cases vary member/class spellings (`p`/`kind`/`items`,
`Base`/`Shape`/`Repo`) and include generic, deep-chain, and index-signature
variants; per §26 the matrix covers the repro plus equivalent shapes and
negative/fallback guards (unrelated classes, private strict nominal, narrowing
`public`→`protected`).

## Project Corpus Impact

- Row: `vite-vanilla-ts-app` (case `binder-11-20`), and any row that assigns a
  subclass widening a protected member to its base.
- Bug family: protected-member assignability treated as declaration-identical.
- Evidence: behavior verified against `tsc` 6.0.2; new checker + solver unit
  tests; existing visibility suites (`private_brands`,
  `union_protected_intersection_property_tests`, solver lawyer/subtype
  visibility tests) stay green.

## Verification

- `cargo test -p tsz-checker --test class_protected_widening_assignability_tests` → 9 passed.
- `cargo test -p tsz-solver --lib -- protected private brand nominal visibility derived` → 61 passed.
- `cargo test -p tsz-checker --lib -- protected private accessib visibility ts2445 ts2341 ts2415` → 147 passed.
- `cargo clippy -p tsz-solver --lib --tests -- -D warnings` → clean; `cargo fmt` clean.

## Coordination Notes

AgentName: opus48-web-pensive-wright
#11631 was unowned and not WIP when claimed; marked WIP with a signed comment.
Also closed sibling #11589 as already-resolved (covered by merged #11661) while
triaging this family. `/simplify` run 3×: collapsed two near-duplicate helpers
into one shared `nominal_member_origin_ok`, simplified the test builder via
`PropertyInfo::new`, and (pass 2) found + fixed two additional decision sites
(with-index object path + the explain/diagnostic path) so the rule is applied
everywhere, not just at the reported repro.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TnjMaBCpqWbMAfwMCLJxaK)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -103,6 +103,10 @@ name = "generic_alias_assignability_pollution_tests"
 path = "tests/generic_alias_assignability_pollution_tests.rs"
 
 [[test]]
+name = "class_protected_widening_assignability_tests"
+path = "tests/class_protected_widening_assignability_tests.rs"
+
+[[test]]
 name = "generic_default_branch_independence_tests"
 path = "tests/generic_default_branch_independence_tests.rs"
 
--- a/crates/tsz-checker/tests/class_protected_widening_assignability_tests.rs
+++ b/crates/tsz-checker/tests/class_protected_widening_assignability_tests.rs
@@ -1,0 +1,203 @@
+//! Regression coverage: `protected` member assignability is *hierarchical*,
+//! not strictly nominal like `private`.
+//!
+//! Tracks the `binder-11-20` "augmentation drops private/protected mirror
+//! visibility" family (mohsen1/tsz#11631).
+//!
+//! Structural rule under test:
+//!
+//! > When a target property is `protected`, a source object is assignable as
+//! > long as the source property is declared in the target's
+//! > protected-declaring class *or in a class derived from it*. The source is
+//! > allowed to widen the member from `protected` to `public` in the derived
+//! > class. (`private` is unaffected: it still requires the exact same
+//! > declaration — declaration identity.)
+//!
+//! This mirrors tsc's `isPropertyInClassDerivedFrom` / `isValidOverrideOf`.
+//! Behavior was verified against `tsc` 6.0.2 for every case below.
+//!
+//! Per the anti-hardcoding directive (§25) the cases vary the user-chosen
+//! member and class names (`p`/`kind`/`items`, `Base`/`Shape`/`Repo`) and
+//! include a generic variant, so a fix keyed on a specific spelling would fail
+//! at least one case.
+
+use tsz_checker::test_utils::{check_source_strict, diagnostic_codes, diagnostics_without_codes};
+
+/// Semantic error codes, dropping TS2318 ("cannot find global type") which is
+/// expected noise in the no-stdlib unit harness.
+fn errors(source: &str) -> Vec<u32> {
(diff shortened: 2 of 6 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Queue handoff: #11742 is non-draft, not WIP, mergeable, and exact-head CI is green for `3ebf9d13028a028677d1e55fd8732f35be7f9829`; the only required blocker is `Queue Tested`.

Adding `merge-queue` and waking the repo queue now.

PR #11769, opened 2026-05-30
fix(emit): capture this into _this when an ES5 derived super-call argument references this

mohsen1:
## Agent
AgentName: opus48-emit100

## Track
emit/js — ES5 derived-constructor this-in-super-arg capture (emit→100% campaign, wave 3)

## Invariant
When an ES5-downleveled derived constructor's `super(...)` call has any argument expression containing a `this` keyword reference, tsc captures `this` into `_this` (`var _this = _super.call(this, _this) || this; return _this;`) rather than the inlined tail-super-return form (`return _super.call(this, this) || this;`) — because the inlined form references the pre-super `this`. So the tail-super-return optimization must be DISABLED when any super-call argument contains a `this` reference. Keyed on AST node kind (super-call argument containing `this`, via `contains_this_reference`, lexical-boundary aware) — no identifier/file-name matching, no printer-output inspection. Falling through to the existing full path already emits the arg-captured `_super.call(this, _this)` + `return _this`.

## Scope
- `crates/tsz-emitter/src/transforms/class_es5_ir_constructor.rs` — new `super_call_arguments_reference_this(...)`; `&& !super_args_reference_this` guard on `can_use_tail_super_return`.
- `crates/tsz-emitter/src/emitter/source_file/derived_constructor_tests.rs` (+3 tests).

## Project Corpus Impact
- Row: n/a (ES5 derived-constructor emit fix)
- Bug family: ES5 tail-super-return fast path skips `this`→`_this` capture when a super-call argument references `this`
- Evidence: thisInInvalidContextsExternalModule(target=es5) FAIL→PASS; full --js-only 79→78.

## Verification
- Witness FAIL→PASS (`ClassWithNoInitializer`: now `var _this = _super.call(this, _this) || this; return _this;`). **Full --js-only: 79→78 fail, net +1, ZERO regressions** (newly-fixed = exactly the target; no new failures). DTS unaffected (ES5 class IR JS-only path). The other verbose diffs (`// Error` etc. trailing comments) are comment-only, normalized by the harness ladder. 3 new structural unit tests (no-field super-arg-this capture; renamed-class; negative `super(42)` keeps inline tail form). Clippy clean; arch guard passes.
- §25/§26 compliant.

## Coordination Notes
Rebased onto current main. Root cause: the no-field/super-only fast path (`can_use_tail_super_return` → `emit_super_call_return_ir`) didn't rewrite `this` args to `_this`; the full path (`emit_super_call_ir` with `capture_args_this=true`) already did, which is why the sibling field-bearing case was correct. — opus48-emit100

### DIFF
--- a/crates/tsz-emitter/src/emitter/source_file/derived_constructor_tests.rs
+++ b/crates/tsz-emitter/src/emitter/source_file/derived_constructor_tests.rs
@@ -77,6 +77,64 @@ fn derived_constructor_fields_follow_parenthesized_super_call() {
 }
 
 #[test]
+fn super_arg_this_without_fields_captures_this_es5() {
+    // A `this` reference inside the super-call arguments must materialize the
+    // captured `_this` form even when the constructor has no field initializer
+    // and `super(...)` is its only statement. tsc emits
+    // `var _this = _super.call(this, _this) || this; return _this;` rather than
+    // the inlined `return _super.call(this, this) || this;` tail form.
+    let source = "class Base { constructor(t: any) {} }\nclass OnlySuperArg extends Base {\n    constructor() {\n        super(this);\n    }\n}\n";
+
+    let es5_output = emit(source, ScriptTarget::ES5);
+
+    assert!(
+        es5_output.contains(
+            "function OnlySuperArg() {\n        var _this = _super.call(this, _this) || this;\n        return _this;\n    }"
+        ),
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Exact-head ready-review CI is green for `63b561da1412e7b1819aea1f28c33fb945fc19fe`, and this PR is non-draft with no WIP markers. The `merge-queue` label is already present; I am waking the repo-local queue so `Queue Tested` can synthetic-test and land it.

PR #11773, opened 2026-05-30
fix(emit): rename ES5 block-scoped class on out-of-order enclosing-scope name collision

mohsen1:
## Agent
AgentName: opus48-emit100

## Track
emit/js — ES5 block-scoped class out-of-order collision rename (emit→100% campaign, wave 3)

## Invariant
When an ES5 block-scoped class lowered to an IIFE sits in a NESTED block and its name matches a value binding (`var`/`function`/`class`/`enum`/`namespace`) visible throughout an enclosing function/script scope — REGARDLESS of declaration order — the hoisted `var` renames to `Name_N`. Body-level block-scoped classes still reuse their name. Extends the forward-collision fix (#11761): tsz's single forward walk only saw the collision when the outer declaration PRECEDED the class; tsc resolves it via its binder pre-pass. This fix seeds each scope's visible value-binding names up front (body-level declarations of all kinds + hoisted `var`s collected from nested blocks, stopping at nested function/class boundaries) into the block-scope shadow set. Keyed on AST node kind — no identifier/file-name matching, no printer-output inspection.

## Scope
- NEW `crates/tsz-emitter/src/emitter/block_scope_shadow_seed.rs` (216 lines — collector + seeder).
- `emitter/mod.rs` (register), `emitter/source_file/emit.rs` (seed module scope), `emitter/functions.rs` + `statements/core.rs` (seed function-body scopes).
- `transforms/block_scoping_es5.rs` — consult the shadow set in `register_block_scoped_class`.
- `tests/block_scoping_es5.rs` (+2 tests, names A/T/K and A/P/X).

## Project Corpus Impact
- Row: n/a (ES5 block-scoped class-rename fix)
- Bug family: ES5 block-scoped class not renamed when a same-named value binding appears LATER in an enclosing scope (out-of-order collision)
- Evidence: localTypes1(target=es5) FAIL→PASS; full --js-only 77→76.

## Verification
- Witness FAIL→PASS. **Full --js-only: 77→76 fail, net +1, ZERO new failures** (only localTypes1(es5) flipped). 11 adjacent structural repros match tsc (module/body/nested/sibling-block; var/let/function/class outer bindings; both orders; negative no-collision + function-A-sibling). --dts-only unaffected (ES5-JS-var-renaming only). Emitter unit: 21 pre-existing fails identical on clean HEAD; +2 new tests pass; zero new regressions. Clippy clean; arch guard passes (new code in a fresh file, all files <2000).
- §25/§26 compliant.

## Coordination Notes
Rebased onto current main. STOP-reported the other re-triage candidates (NOT this cluster): jsDeclarationsClasses(es5) (multi-cause: inline `@type {*}` JSDoc-cast dropped [checker-coupled] + CJS export ordering for aliased classes); staticFieldWithInterfaceContext (temp-numbering-broad — class-expr `__setFunctionName` temp renumber, the documented global temp-model track); nestedClassDeclaration + nestedGlobalNamespaceInClass (parser-recovery of invalid source). — opus48-emit100

### DIFF
--- a/crates/tsz-emitter/src/emitter/block_scope_shadow_seed.rs
+++ b/crates/tsz-emitter/src/emitter/block_scope_shadow_seed.rs
@@ -1,0 +1,216 @@
+//! Order-independent seeding of value-binding names that an ES5 block-scoped
+//! declaration nested in a sub-block must not capture when lowered to `var`.
+//!
+//! ## Why this exists
+//!
+//! tsc renames a block-scoped class lowered to ES5 (`var Foo = (function(){…})()`)
+//! to `Foo_1` when its name collides with another *value* binding that is
+//! visible throughout the enclosing function/script scope:
+//!
+//! ```ts
+//! function f(b: boolean) {
+//!     if (b) {
+//!         class A { x: number }   // -> var A_1 = …
+//!         let c = [new A()];      //    new A_1()
+//!     }
+//! }
+//! class A {}                      // module-level value binding named `A`
+//! ```
+//!
+//! tsc reaches this decision through its binder pre-pass, so the colliding
+//! outer declaration may appear *textually after* the block-scoped class.
+//! tsz's emitter is a single forward walk, so it must seed the enclosing
+//! scope's visible value-binding names up front; otherwise the collision is
+//! only detected when the outer declaration happens to precede the class.
+//!
+//! ## What is "visible throughout the scope"
+//!
+//! For a function/script scope, the names visible at any nested block are:
+//!   * every declaration directly at the scope's own body level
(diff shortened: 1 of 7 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Exact-head ready CI is green on ee2f05d55e0c10fac0286b2065ec5c3c197337da; Queue Tested is the only remaining pending context and is queue-owned. I am adding merge-queue.

Verification: GitHub ready CI Summary, conformance aggregate, emit, fourslash aggregate, project compile guard/canary, LSP, lint, unit, dist, and WASM are all successful on the current head.

PR #11774, opened 2026-05-30
fix(emit): emit yield __await for implicit awaits in async-generator for-await downlevel (non-runnable JS)

mohsen1:
## Agent
AgentName: opus48-emit100

## Track
emit/js — async-generator for-await downlevel (correctness, non-runnable JS)

## Invariant
When a `for await...of` (ForOfStatement with await modifier) is lowered inside a function body emitted in async-generator-to-generator mode (`emit_await_as_yield_await` — an `async function*` downleveled below ES2018 to a `function*` wrapped in `__asyncGenerator`), the implicit awaits on `iter.next()`, the `iter.return()` cleanup, and the async-`using` dispose MUST be emitted as `yield __await(<expr>)` (mirroring explicit-`await` lowering), NOT a literal `await` — which is a SyntaxError inside the `function*` body. Mode-keyed (lowering-mode flags), not name-keyed. tsc 6.0.3 emits `yield __await(...)`.

## Scope
- NEW `crates/tsz-emitter/src/emitter/es5/bindings_for_await_keyword.rs` — `emit_for_await_implicit_await_prefix`/`_suffix` (pick `await `/`yield `/`yield __await(...)` from the lowering-mode flags).
- `crates/tsz-emitter/src/emitter/es5/bindings_param_patterns.rs` — the 3 keyword sites (loop `.next()` step, `.return()` cleanup, async `using` dispose) call the helpers (net -20/+32, shrank at call sites).
- `es5/mod.rs` (register) + `tests/for_await_async_generator_keyword_tests.rs` (new, 4 tests).

## Project Corpus Impact
- Row: n/a (emit CORRECTNESS fix — non-runnable JS → runnable)
- Bug family: async-generator for-await downlevel emits literal `await` (invalid in `function*`) instead of `yield __await(...)`
- Evidence: `async function* g(y){ for await (const x of y) {...} }` at es2015/es2017 — `node --check` before = SyntaxError, after = valid (2× `yield __await(...)`); matches tsc 6.0.3 `emitter.forAwait` baselines.

## Verification
- node --check VALID at es2015/es2017 (was invalid); async `using` dispose also `yield __await(result_1)`. **Full --js-only fail-SET IDENTICAL (79/79, net-zero — the emit suite's listed cases don't include this multi-file forAwait source, so no row flips; correctness confirmed by node --check + conformance smoke), ZERO regressions. --dts-only IDENTICAL (24/24).** tsz-emitter async/generator/for-of/await/optional-chain family 309/309 (+4 new tests, name-independent + async-non-generator negative). Conformance smoke: emitter.forAwait 1/1, asyncGenerator 21/21. Clippy clean; arch guard passes.
- §25/§26 compliant. Emitter-only (conformance-safe).

## Coordination Notes
Rebased onto current main. Separate non-runnable bug found, NOT bundled (kept this PR contained): labeled for-await (`outer: for await (...) { continue outer; }`) lowers the label onto the wrapping `try` instead of the inner `for`, so `continue outer` targets a non-loop label (SyntaxError); tsc places `outer:` on the inner `for`. Flagged for follow-up. — opus48-emit100

### DIFF
--- a/crates/tsz-emitter/Cargo.toml
+++ b/crates/tsz-emitter/Cargo.toml
@@ -275,6 +275,10 @@ name = "async_loop_emit"
 path = "tests/async_loop_emit.rs"
 
 [[test]]
+name = "for_await_async_generator_keyword_tests"
+path = "tests/for_await_async_generator_keyword_tests.rs"
+
+[[test]]
 name = "while_loop_body_capture_tests"
 path = "tests/while_loop_body_capture_tests.rs"
 
--- a/crates/tsz-emitter/src/emitter/es5/bindings_for_await_keyword.rs
+++ b/crates/tsz-emitter/src/emitter/es5/bindings_for_await_keyword.rs
@@ -1,0 +1,50 @@
+//! Keyword lowering for the implicit awaits emitted by the downlevel
+//! `for await...of` loop transform.
+//!
+//! A `for await...of` statement lowers to a synchronous `for` loop that
+//! awaits the async iterator protocol calls (`iter.next()`, `iter.return()`)
+//! and any async `using` disposal. The keyword used for those implicit awaits
+//! depends on the lowering mode of the *enclosing* function body, mirroring how
+//! an explicit `await` expression is lowered in the same context:
+//!
+//! * Native async (`async function`, target supports `await`): emit `await x`.
+//! * Async-to-generator via `__awaiter` (ES2015/ES2016 async lowering): emit
+//!   `yield x`.
+//! * Async-generator-to-generator via `__asyncGenerator` (an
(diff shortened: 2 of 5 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5
Exact-head ready CI is green on `b8d47842ddcb3503e50830be0136e1698a5beddf`: CI Summary/GitGuardian plus conformance, emit, fourslash, LSP, project compile guard/canary, wasm, lint, unit, and dist contexts have all passed. `Queue Tested` is the only remaining pending context and is queue-owned, so I am adding `merge-queue` and waking the queue.

PR #10778, opened 2026-05-28
refactor(emitter): shard es_decorators.rs into 7 focused submodules

mohsen1:
## Summary

`es_decorators.rs` had grown to **5755 lines**, violating the §19 hard limit (2000 lines max per hand-authored file) and causing `solver_file_size_ceiling_tests::test_emitter_file_size_ceiling` to fail (baseline 5724, actual 5755).

Split following the established `async_es5_ir_*.rs` flat-file pattern already used in the same transforms directory.

### New files (all under 2000 lines):

| File | Lines | Contents |
|------|-------|----------|
| `es_decorators_application.rs` | 181 | `emit_decorator_application` |
| `es_decorators_class_body.rs` | 1179 | `emit_class_body` + `render_decorated_constructor` |
| `es_decorators_member_emit.rs` | 1117 | auto accessor / static private / member text helpers |
| `es_decorators_node_utils.rs` | 416 | node text utilities |
| `es_decorators_collection.rs` | 564 | collect / analysis / compute helpers |
| `es_decorators_private_members.rs` | 571 | `emit_es_decorate_call` + private member emit |
| `es_decorators_collect_fields.rs` | 842 | field / constructor / auto-accessor collection |

`es_decorators.rs` itself is now **1129** lines — struct definitions, constructor, setters, and `emit_class`.

AgentName: Studio-C

## Track

Refactor / §19 compliance

## Invariant

Behavior-preserving: pure structural split of one `impl` block into multiple `impl` blocks across sibling files. No logic changes. During the 2026-05-30 rebase, the post-base decorated-derived-constructor fixes from #11647/#11750 were preserved in the split `es_decorators_member_emit.rs` module, including the final `super(...arguments)` behavior.

## Scope

`crates/tsz-emitter/src/transforms/es_decorators*.rs` only.

## Project Corpus Impact

- Row: n/a
- Bug family: n/a
- Evidence: pure refactor — no algorithm or diagnostic changes; focused local compile/file-size checks pass; ready CI will run full emit/conformance/fourslash gates.

## Verification

Original draft verification:
- `cargo check -p tsz-emitter` — clean
- `cargo nextest run -p tsz-solver -E 'test(file_size_ceiling)'` — `test_emitter_file_size_ceiling` PASS (was FAIL)

2026-05-30 rebase verification on exact head `abeb85cdeb`:
- `cargo fmt --check`
- `python3 scripts/arch/arch_guard.py`
- `cargo check -p tsz-emitter`
[...]

PR #11776, opened 2026-05-30
fix(emit): scoped-static-super rewrite for hoisted decorated static-field super at pre-ES2022 (non-runnable JS)

mohsen1:
## Agent
AgentName: opus48-emit100

## Track
emit/js — correctness (non-runnable JS): decorated-class hoisted static-field super

## Invariant
When a TC39-decorated class is lowered for a pre-ES2022 target (static fields hoisted out of the class body to `_classThis.<name> = <init>`), and a static field initializer references the `super` keyword, the hoisted value MUST use the scoped-static-super rewrite (`Reflect.get(_classSuper, "x", _classThis)` for reads; an IIFE wrapping `Reflect.set(...)` for writes/updates) — the SAME rewrite already applied to the in-body ES2022 path — instead of raw source text. A raw `super` in `_classThis.<name> = super.x;` is a SyntaxError (`super` is illegal outside the class body). Keyed on AST node kind (super reference in a hoisted static-field initializer) — no identifier/file-name matching.

## Scope
- NEW `crates/tsz-emitter/src/emitter/decorator_static_super_hoist.rs` — seeds the rewritten value via the existing `emit_expression_with_scoped_static_initializer_mode` machinery (gated pre-ES2022); also hosts a combined `seed_tc39_decorator_static_member_and_hoisted` helper to keep transform_dispatch.rs a single call site.
- `crates/tsz-emitter/src/transforms/es_decorators_node_utils.rs` — `plain_static_field_assignment` (relocated here by #10778's sharding) now takes `member_idx` + consults the seeded value map.
- `crates/tsz-emitter/src/transforms/es_decorators.rs` — value map field + setter; `es_decorators_class_body.rs` passes member_idx.
- `crates/tsz-emitter/src/emitter/transform_dispatch.rs` — single combined seed call (2119 ≤ 2124 ratchet, below baseline).
- `emitter/mod.rs` (register) + `source_file/tc39_decorator_tests.rs` (+6 tests).

## Project Corpus Impact
- Row: n/a (emit CORRECTNESS fix — non-runnable JS → runnable)
- Bug family: decorated-class hoisted static-field initializer emits raw `super` (invalid outside class body) at pre-ES2022
- Evidence: `@dec class C extends Base { static a = super.x; }` — `node --check` before = SyntaxError at ES2015/2017/2020/2021, after = valid `_classThis.a = Reflect.get(_classSuper, "x", _classThis)`; matches tsc 6.0.3. Original witness: esDecorators-classDeclaration-classSuper.5.

## Verification
- node --check VALID at ES2015/2017/2020/2021/2022 (was invalid pre-ES2022); reads/writes/prefix-update + `super[...]` index all match tsc byte-for-byte. **Full --js-only fail-SET IDENTICAL (78/78, net-zero — the conformance witness runs only at ES2022 where output was already correct), ZERO regressions. --dts-only IDENTICAL (24/24).** 6 new structural tests (read/write/prefix-update, super[...] index, ES2020, super-free negative). Conformance smoke: classDeclaration/classSuper 7/7, classExpression/classSuper 6/6. Clippy clean; arch guard passes (all ratchets respected; transform_dispatch.rs reduced to 2119).
- §25/§26 compliant. Emitter-only (conformance-safe).

## Coordination Notes
Rebased onto current main AFTER sibling #10778 sharded es_decorators.rs into 8 submodules — re-applied the behavioral change onto the new shards (hooked into es_decorators_node_utils.rs), dropped the now-redundant relocation, consolidated seed calls to respect the transform_dispatch.rs ratchet. — opus48-emit100

### DIFF
--- a/crates/tsz-emitter/src/emitter/decorator_static_super_hoist.rs
+++ b/crates/tsz-emitter/src/emitter/decorator_static_super_hoist.rs
@@ -1,0 +1,81 @@
+//! Hoisted-static-field `super` rewrite for pre-ES2022 decorated classes.
+//!
+//! When a TC39-decorated class is lowered for a target below ES2022, plain
+//! static fields cannot stay as class-body declarations: they are hoisted to
+//! `_classThis.<name> = <init>` statements emitted *after* the class. A bare
+//! `super` token is only legal inside a class/object method body, so an
+//! initializer such as `static a = super.x` becomes `_classThis.a = super.x`,
+//! which Node rejects with `'super' keyword unexpected here`.
+//!
+//! `tsc` rewrites the value to the scoped-static-super form
+//! (`Reflect.get(_classSuper, "x", _classThis)` for reads, an IIFE wrapping
+//! `Reflect.set(...)` for writes) before hoisting. The hoist assembly lives in
+//! the `TC39DecoratorEmitter` transform, which only has raw source text. This
+//! helper renders the rewritten *value* (no name, no modifiers, no trailing
+//! `;`) through the main emitter's scoped-static-super machinery and hands it
+//! to the transform via `set_hoisted_static_field_value_text`.
+
+use super::Printer;
+use tsz_parser::parser::NodeIndex;
+
+impl<'a> Printer<'a> {
+    /// Seed both the in-body static-member text and (for pre-ES2022 lowering)
+    /// the hoisted-assignment value text for a static field whose initializer
+    /// references `super` or `this`. Combining the two seedings keeps the
+    /// dispatch call site to a single statement.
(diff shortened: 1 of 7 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5
Exact-head ready CI is green on `d39c7c25d9db0a233037a6820f3bddc457e27eb3`: CI Summary/GitGuardian plus conformance, emit, fourslash, LSP, project compile guard/canary, wasm, lint, unit, and dist contexts have all passed. `Queue Tested` is queue-owned, so I am adding `merge-queue` and waking the queue.

PR #11783, opened 2026-05-30
fix(emit): drop spurious parens on UMD dynamic-import conditional under yield (+2)

mohsen1:
AgentName: m3-max-64g-opus48

## Track
Emit parity → 100% (JS emit). Dynamic `import()` lowering for AMD/UMD module format.

## Invariant
When lowering a dynamic `import()` for the AMD/UMD module format, the
`__syncRequire ? <cjs> : <amd>` conditional is parenthesized **only** when its
parent expression binds tighter than `?:`. A native (retained) `await` operand
binds tighter and keeps the parens; but when async-to-generator lowering
rewrites that `await` to `yield` (targets below ES2017), or for a source-level
`yield`, the operand binds looser than `?:`, so the bare conditional needs no
parens. Keyed on AST parent kind + the await→yield module/target lowering facts
(`emit_await_as_yield`, `emit_await_as_yield_await`, `needs_async_lowering` +
function-scope depth), not on identifier/file names or printer output.

## Scope
- `crates/tsz-emitter/src/emitter/expressions/call.rs` — `umd_dynamic_import_needs_parens`
  gains a `const fn await_parent_emits_as_yield` predicate (mirrors
  `emit_await_expression`); a captured comma `_a = spec, ...` sequence still
  parenthesizes via the existing `is_sequence` path. Structured paren-emit
  decision, no text surgery.
- `crates/tsz-emitter/tests/dynamic_import_amd_umd_parity_tests.rs` — 4 new tests.

## Inverted-report correction
The internal hand-off claimed tsc *adds* parens and tsz was missing them. The
opposite is true: tsc 6.0.3 emits the **bare** `yield __syncRequire ? ... : ...`
and tsz was *wrongly adding* a paren, which also displaced the trailing import
comment. Removing the spurious paren fixed both the spelling and the comment.

## Project Corpus Impact
- Row: emit (`--js-only`)
- Bug family: AMD/UMD dynamic `import()` lowering parenthesizes the `__syncRequire ? :` conditional even when the `yield` operand binds looser than `?:`
- Evidence: `scripts/emit/run.sh --js-only -j4` before 77 fail / 13453 pass → after 75 fail / 13455 pass (**+2, zero new failures**). Witnesses flipped: `importCallExpressionAsyncES6UMD`, `importCallExpressionAsyncES5UMD(target=es2015)`. `--dts-only` 24 fail unaffected. `node --check` on both witnesses OK. All 85 `importCallExpression*` JS tests pass.

## Verification
- arch guard `Architecture guardrails passed` (fit inside existing fn, no new file).
- 17/17 `dynamic_import_amd_umd_parity_tests` pass post-rebase, incl. a
  renamed-binding + renamed-specifier test proving no hardcoding, an ES2017
  native-await negative (parens preserved), and statement/return-position cases.
- clippy `-D warnings` clean; `node --check` valid on both witnesses.

## Coordination Notes
Emit is OUTPUT — structured precedence-paren decision in the printer, no
semantic validation, no output surgery. Stayed entirely within the
dynamic-import paren decision; the for-await-of `_d` var and System `execute:`
async-marking paths (distinct clusters) were not touched. Two pre-existing
unrelated crate test failures (createRequire emission + a declaration namespace
import-alias) were confirmed present on clean HEAD before this change.

### DIFF
--- a/crates/tsz-emitter/src/emitter/expressions/call.rs
+++ b/crates/tsz-emitter/src/emitter/expressions/call.rs
@@ -854,18 +854,34 @@ impl<'a> Printer<'a> {
         }
     }
 
+    /// Whether the parent `await` keyword is lowered to a `yield` keyword in
+    /// the current emit context, mirroring `emit_await_expression`.
+    ///
+    /// A `yield` operand binds looser than `?:`, so `yield a ? b : c` already
+    /// parses as `yield (a ? b : c)`; a native `await` binds tighter than `?:`,
+    /// so `await a ? b : c` parses as `(await a) ? b : c` and needs parens.
+    /// This decides whether an `await`-parented UMD conditional needs wrapping.
+    const fn await_parent_emits_as_yield(&self) -> bool {
+        self.ctx.emit_await_as_yield
+            || self.ctx.emit_await_as_yield_await
+            || (self.ctx.needs_async_lowering && self.function_scope_depth > 0)
+    }
+
     /// Whether a UMD dynamic-`import()` substitution must be parenthesized for
     /// the parent expression it sits in, matching `tsc`'s parenthesizer.
     ///
     /// Statement-level parents (expression statement, `return`, `throw`, `for`
     /// headers) accept a full expression — including a comma sequence — so
     /// neither form needs parentheses. Parents that bind tighter than `?:`
-    /// (operand of `await`/`yield`/unary/binary, object of a member access,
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Exact-head ready CI is green on `07714c84146f3cc08a2dba9036337c197ba62b77`: PR-head checks are passing/skipped, and `Queue Tested` is queue-owned. I am adding `merge-queue` and waking the queue now.

PR #11793, opened 2026-05-30
fix(emit): erase no-initializer static fields under useDefineForClassFields (+1 staticPropertyNameConflicts)

mohsen1:
AgentName: m3-max-64g-opus48

## Track
Emit parity → 100% (JS emit). Static class-field define-for-class-fields lowering.

## Invariant
When a class field has NO initializer AND carries a `static` modifier, tsc's
`useDefineForClassFields` field-lowering erases it entirely (emits no
`Object.defineProperty(C, …)`). A no-init *instance* field still materializes
with `value: void 0`; an *initialized* static field still emits its define.
Keyed structurally on initializer-presence + static-modifier + define
semantics — NOT on the field name.

## Scope
- `crates/tsz-emitter/src/emitter/declarations/class/static_field_erasure.rs`
  (new, 66 lines): `static_no_init_field_is_erased(no_initializer, use_define_for_class_fields)`
  predicate + 3 unit tests.
- `declarations/class/emit_es6.rs`: call-site at the static-field collection
  branch (kept at its 4191-line ratchet by housing the predicate in the new module).
- `declarations/class/mod.rs`: module registration.
- `crates/tsz-emitter/tests/static_no_init_field_erasure_tests.rs` (new, 193 lines): 7 tests; registered in Cargo.toml.

## Bug
tsz emitted `Object.defineProperty(C, <name>, { … value: })` with an EMPTY
value expression (invalid JS) for no-init static fields. The earlier
"Function own-property name conflict" framing was a red herring — the real tsc
rule erases ALL no-init static fields regardless of name (the test merely used
reserved names because those are the ones that would conflict if defined).

## Project Corpus Impact
- Row: emit (`--js-only`)
- Bug family: no-initializer static field emitted as empty-`value:` Object.defineProperty instead of erased (useDefineForClassFields)
- Evidence: `--js-only -j4` 72→71 (flips `staticPropertyNameConflicts(target=es2015,useDefineForClassFields=true)`, zero new). `--dts-only` 24→24. `node --check` passes; zero empty-`value:` lines remain.

## Verification
- arch guard `Architecture guardrails passed`; clippy `-D warnings` clean.
- 10 new tests: witness (`static name;`), renamed (`static length;`),
  non-conflicting user name (`static notReserved;` — also erased, proving it's
  the static-no-init family not the name), computed-name (define erased, temp
  still hoisted), negatives (initialized static reserved-name kept; no-init
  instance field materialized), es5 erasure.

## Coordination Notes
Emit is OUTPUT — a structural erasure predicate, no name-string matching, no
output surgery, no semantic validation. STOP-scoped: `staticPropertyNameConflicts(target=es5)`
left failing — its remaining diff is computed-name temp PLACEMENT (the broad
static-member capture-phase / temp-model track, Stages A–C), explicitly out of
scope; this change does not regress it. No creep into typeOfThisInStaticMembers/
thisAndSuperInStaticMembers.

### DIFF
--- a/crates/tsz-emitter/Cargo.toml
+++ b/crates/tsz-emitter/Cargo.toml
@@ -91,6 +91,10 @@ name = "computed_name_no_init_define_field_tests"
 path = "tests/computed_name_no_init_define_field_tests.rs"
 
 [[test]]
+name = "static_no_init_field_erasure_tests"
+path = "tests/static_no_init_field_erasure_tests.rs"
+
+[[test]]
 name = "jsx_spread_tests"
 path = "tests/jsx_spread_tests.rs"
 
--- a/crates/tsz-emitter/src/emitter/declarations/class/emit_es6.rs
+++ b/crates/tsz-emitter/src/emitter/declarations/class/emit_es6.rs
@@ -1,6 +1,7 @@
 use super::super::super::core::PropertyNameEmit;
 use super::super::super::{Printer, ScriptTarget};
 use super::replace_identifier;
+use super::static_field_erasure::static_no_init_field_is_erased;
 use super::{AutoAccessorEmitOptions, AutoAccessorInfo, StaticFieldInit};
 use crate::emitter::core::{PrivateMemberInfo, PrivateMethodDef};
 use crate::transforms::private_fields_es5::{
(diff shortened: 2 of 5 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Exact-head ready CI is green on `aeaa4888d0cbb2a97d340e5619541e30c31ad957`: PR-head checks are passing/skipped, and `Queue Tested` is queue-owned. The `merge-queue` label was already present; I normalized canonical labels (`emit`, `agent:Studio-C`) and am waking the queue now.

PR #11800, opened 2026-05-30
fix(emit): descend into template substitutions in ES5 block-scoping closure capture (latent correctness)

mohsen1:
AgentName: m3-max-64g-opus48

## Track
Emit correctness (bugs-on-top) — ES5 block-scoping closure-capture analysis.

## Invariant
When the ES5 block-scoping capture analyzer walks an expression's children,
template-literal substitution expressions (`TEMPLATE_EXPRESSION` head +
`TEMPLATE_SPAN` expressions, `TAGGED_TEMPLATE_EXPRESSION` tag + template) — and
the other unary/new/spread/labeled child positions — are real child nodes and
must be visited, so a loop binding referenced only inside `${...}` is detected
and gets correct per-iteration `_loop_N` IIFE capture. Keyed on AST node KIND.

## Scope
- `crates/tsz-emitter/src/transforms/block_scoping_es5.rs` (+65): replace the
  `_ => {}` catch-all in `visit_children` with real arms for
  TEMPLATE_EXPRESSION / TEMPLATE_SPAN / TAGGED_TEMPLATE_EXPRESSION /
  NEW_EXPRESSION / PREFIX+POSTFIX_UNARY / AWAIT+YIELD+NON_NULL+SPREAD /
  LABELED_STATEMENT (each verified genuinely unvisited before adding).
- `crates/tsz-emitter/tests/loop_capture_template_substitution_tests.rs` (new, 5 tests); Cargo.toml registration.

## Bug
The ES5 block-scoping closure-capture child-walk swallowed template-substitution
(and several other) node kinds via a `_ => {}` catch-all, so a loop variable
referenced only inside a `${...}` captured by a closure was never detected → a
plain `for` loop emitted instead of the correct `_loop_N` per-iteration IIFE
(wrong runtime capture semantics at ES5).

## Project Corpus Impact
- Row: n/a (emit CORRECTNESS fix — latent ES5 closure-capture bug; well-formed inputs were mis-lowered)
- Bug family: ES5 block-scoping capture analysis does not descend into template-literal substitutions / other child positions
- Evidence: `--js-only -j4` 72→72, `--dts-only` 24→24 (net-zero, zero regressions). The originally-clustered `labeledStatementDeclarationListInLoopNoCrash3/4` witnesses do NOT flip — their residual diff is an INDEPENDENT malformed-unterminated-template parser/scanner error-recovery confound (the es2015 variant, which has no block-scoping transform, fails identically), so those are re-routed to the parser-recovery lane. This fix is proven on WELL-FORMED inputs: 5 new unit tests + CLI ES5 emit producing correct `_loop_1(fontType)` / tagged-template `__makeTemplateObject` capture; `node --check` OK.

## Verification
- arch guard `Architecture guardrails passed`; clippy `-D warnings` clean; file 1027 lines.
- 5 new tests (for-of capture, while multi-span, tagged template, renamed binding proving no name-keying, negative no-capture) pass; full `cargo nextest -p tsz-emitter` = only the 13 pre-existing unrelated failures (confirmed on origin/main).

## Coordination Notes
Emit is OUTPUT — pure traversal completeness fix, no semantic validation, no
output surgery. Net-zero on the metric but fixes a real latent ES5 capture bug.
Re-routes the `labeledStatementDeclarationListInLoopNoCrash` cluster to
parser-error-recovery (conformance-gated) in the endgame plan — it was
mis-clustered as block-scoping-traversal.

### DIFF
--- a/crates/tsz-emitter/Cargo.toml
+++ b/crates/tsz-emitter/Cargo.toml
@@ -207,6 +207,10 @@ name = "loop_capture_labeled_var_recovery_tests"
 path = "tests/loop_capture_labeled_var_recovery_tests.rs"
 
 [[test]]
+name = "loop_capture_template_substitution_tests"
+path = "tests/loop_capture_template_substitution_tests.rs"
+
+[[test]]
 name = "loop_capture_continue_control_tests"
 path = "tests/loop_capture_continue_control_tests.rs"
 
--- a/crates/tsz-emitter/src/transforms/block_scoping_es5.rs
+++ b/crates/tsz-emitter/src/transforms/block_scoping_es5.rs
@@ -926,6 +926,71 @@ fn visit_children<F: FnMut(NodeIndex)>(arena: &NodeArena, node: &Node, mut visit
                 visitor(ret.expression);
             }
         }
+        // A labeled statement (`label: stmt`) is transparent to capture
+        // analysis: the labeled body can be a loop or block whose closures
+        // capture an outer loop variable, so we must descend into it.
+        k if k == syntax_kind_ext::LABELED_STATEMENT => {
+            if let Some(labeled) = arena.get_labeled_statement(node) {
+                visitor(labeled.statement);
+            }
+        }
+        // `new Expr(args)` shares the call-expression payload. A captured loop
+        // variable can appear in the callee or any argument
+        // (e.g. `new Foo(() => x)`), so descend like a call expression.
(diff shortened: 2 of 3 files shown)


mohsen1:
AgentName: m5-pro-48g-gpt5

Exact-head ready CI is green on `2cb38b9a5c77f22ef0305f363d401dcd177979d2`: PR-head checks are passing/skipped, and `Queue Tested` is queue-owned. I am adding `merge-queue` and waking the queue now.

PR #11822, opened 2026-05-30
fix(module-resolver): resolve path mapping targets with explicit extensions

mohsen1:
AgentName: claude-sonnet-4-6

## Track
Module resolver / `tsconfig` paths-mapping parity

## Invariant
When a tsconfig `paths` target is an explicit-extension path (e.g. `"./foo.d.ts"`, `"./stubs/*.d.ts"`), TypeScript probes the substituted path relative to `baseUrl` with its normal file/directory lookup. tsz was skipping all such targets silently.

### Bug fixed
`has_path_mapping_target_extension` checked whether a substituted target had a known TypeScript/JavaScript extension (`.ts`, `.d.ts`, `.js`, `.tsx`, …). If it did, the target was `continue`d — never probed — causing all fixed-extension `paths` targets to silently fail and fall through to bare-specifier resolution (TS2307 or wrong resolution).

This broke every tsconfig that uses the nextjs guard pattern:
```jsonc
"paths": {
  "next/dist/compiled/*": ["./external.d.ts"],   // ← was skipped
  "next/dist/*":          ["./src/*"],            // ← worked
  "*":                    ["./external.d.ts"]     // ← was skipped
}
```

### Additional improvements (all within the same file cluster)
- **Remove redundant per-call sort**: `build_path_mappings` already sorts by specificity at construction time; the O(n log n) + Vec alloc on every resolution call was wasted.
- **Use shared `apply_wildcard_substitution`**: replaced inline `contains('*')` + `replace` with the existing crate helper that uses `replacen` (first `*` only, per spec).
- **Pass `base: &Path` to `try_path_mappings`**: the call site already guards `base_url.is_some()`; passing the path makes the precondition part of the type signature instead of a distant-invariant `expect`.
- **10 new unit tests** using `TempFixture` (parallel-safe, RAII cleanup), covering: exact `.d.ts` target, wildcard `.d.ts`/`.ts` suffix targets, catch-all `"*"` → fixed file, specificity ordering, fallback targets, and the full nextjs guard config pattern.
- **Updated stale comment** in `lookup_integration.rs` that referenced the now-removed extension-filter behaviour.
- **Pre-existing clippy fixes** in `resolution/helpers.rs`: `map_or` → `is_none_or`, complex-type alias `BestExportPatternEntry`.

## Project Corpus Impact
- Row: nextjs-guard
- Bug family: paths-mapping silent drop of explicit-extension targets → TS2307 divergence on any specifier matched by a fixed-.d.ts target
- Evidence: 10 new unit tests reproduce the exact nextjs guard pattern and all pass locally; `cargo nextest run -p tsz-core -E 'test(module_resolver)'` → 217/217 pass; `cargo clippy -p tsz-core --all-targets --all-features -- -D warnings` → clean.

## Verification
```
cargo clippy -p tsz-core --all-targets --all-features -- -D warnings  # clean
cargo nextest run -p tsz-core -E 'test(module_resolver)'               # 217/217
```

## Coordination Notes
- Claimed from open issue pool. No overlapping open PR targets the `paths`-mapping extension-skip bug.
- The fix is structural (removes the bad guard entirely) rather than a special-case patch. Tests cover exact targets, wildcard suffix targets, catch-all patterns, and specificity ordering — not just the reported fixture.
- Ran `/simplify` 3× before opening; each round had actionable findings that were applied.

### DIFF
--- a/crates/tsz-core/src/module_resolver/mod.rs
+++ b/crates/tsz-core/src/module_resolver/mod.rs
@@ -553,8 +553,10 @@ impl ModuleResolver {
         // TypeScript treats `paths` mappings as requiring `baseUrl` to avoid surprising
         // absolute lookups that behave like relative resolution.
         let mut path_mapping_attempted = false;
-        if self.base_url.is_some() && !self.path_mappings.is_empty() {
-            let attempt = self.try_path_mappings(specifier, containing_dir);
+        if let Some(base_url) = self.base_url.as_deref()
+            && !self.path_mappings.is_empty()
+        {
+            let attempt = self.try_path_mappings(specifier, base_url);
             if let Some(resolved) = attempt.resolved {
                 return (Ok(resolved), path_mapping_attempted);
             }
--- a/crates/tsz-core/src/module_resolver/path_mapping.rs
+++ b/crates/tsz-core/src/module_resolver/path_mapping.rs
@@ -1,7 +1,7 @@
 //! Path mapping resolution from tsconfig `paths` and `baseUrl`.
 
 use super::{ModuleExtension, ModuleResolver, ResolvedModule};
-use crate::module_resolver_helpers::split_path_extension;
+use crate::resolution::helpers::apply_wildcard_substitution;
 use std::path::Path;
 
 pub(super) struct PathMappingAttempt {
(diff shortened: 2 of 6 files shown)


mohsen1:
`/retest`

PR body updated — `project-corpus-pr-body` should now pass. Fixed field formats:
- `AgentName: claude-sonnet-4-6` (inline, not a section header)
- `- Row: nextjs-guard` (list item)
- `- Bug family: ...` (list item)

All other checks (conformance, fourslash, unit, emit, wasm) passed on the first run.

---

PR #11860, opened 2026-05-31
perf(checker): avoid alias post-validation relowering

mohsen1:
## Agent
AgentName: Studio-B

## Track
Green-row performance / checker alias validation

## Invariant
Preserve TypeScript diagnostic parity while avoiding duplicate type-alias body lowering after `check_type_node` has already validated reference-covered subtrees. Unsupported alias body shapes still fall back to the existing `check_type_for_missing_names` path.

## Scope
- Route type-alias post-validation through an alias-body walker instead of re-lowering the whole body whenever the initial type-node pass did not cover every missing-name case.
- Reuse existing name-resolution/checker diagnostics for type references, conditional infer scopes, mapped-type template checks, readonly operators, template literal spans, and fallback shapes.
- Preserve generic type-reference argument validation (`TS2344`/arity) for resolved references in the optimized post-validation walker.
- No solver relation semantics, emit, parser, binder, LSP, WASM, or roadmap changes.

## Project Corpus Impact
- Row: `ts-essentials-project`
- Bug family: duplicate alias body missing-name validation/re-lowering in utility-type mapped/conditional/indexed alias bodies.
- Evidence: green `ts-essentials-project` row, 95 files reached, no diagnostics in either compiler. Current-main quick benchmark before patch: `/private/tmp/studio-b-ts-essentials-before-alias-postwalk-currentmain-20260531.json` and `.tsgo-winners.json` (`tsz=188.26 ms`, `tsgo=154.63 ms`, `tsgo` wins `1.22x`, noisy σ). After patch: `/private/tmp/studio-b-ts-essentials-after-alias-postwalk-currentmain2-20260531.json` and `.tsgo-winners.json` (`tsz=79.86 ms`, `tsgo=84.53 ms`, `tsz` wins `1.06x`). 2x target gap remains open: `target_gap_factor=1.8895`.

## Verification
- `cargo fmt --all`
- `git diff --check`
- `CARGO_TARGET_DIR=/Users/mohsen/code/tsz/.target cargo check -p tsz-checker`
- `CARGO_TARGET_DIR=/Users/mohsen/code/tsz/.target cargo nextest run -p tsz-checker --test ts2304_tests`
- `CARGO_TARGET_DIR=/Users/mohsen/code/tsz/.target cargo nextest run -p tsz-checker --lib missing_name_pass_constrained_type_param_tests`
[...]

Issue #7574, opened 2026-05-16
perf(solver): unbounded recursion crashes tsz on large-ts-repo at 1.59M LOC (stack overflow, both reuse modes)


mohsen1:
## Summary

Discovered in PR #7521 experiments (`docs/architecture/LSP_PERF_EXPERIMENTS_2026-05-16.md` E7): tsz crashes with `thread '<unknown>' has overflowed its stack` (exit 134, SIGABRT) when checking `large-ts-repo` (~10,820 .ts files / 1.59 M LOC) with the default flat tsconfig (`tsconfig.flat.bench.json`).

The crash happens in **both reuse modes** and survives `RUST_MIN_STACK=4294967296` (4 GB worker thread stack) plus `ulimit -s hard`. This is **not** a stack-tuning issue — it's unbounded recursion in the checker on a specific type shape in `large-ts-repo`.

## Repro

```bash
git clone --depth 1 https://github.com/mohsen1/large-ts-repo.git ~/code/large-ts-repo
cd ~/code/large-ts-repo && pnpm install --frozen-lockfile
cargo build --release --bin tsz
RUST_MIN_STACK=536870912 ./target/release/tsz --extendedDiagnostics --noEmit \
    -p ~/code/large-ts-repo/tsconfig.flat.bench.json
```

Expected: completes with diagnostics + extendedDiagnostics summary.
Actual: ~60 seconds in (reuse off) or ~7 minutes in (reuse on), prints:
\`\`\`
thread '<unknown>' (...) has overflowed its stack
fatal runtime error: stack overflow, aborting
\`\`\`
Exit code 134.

## Why this matters

- Blocks tsz on the LSP-target 1M-LOC stress fixture.
- The previously reported "exit 137 OOM at 6.2 GB on 32 GB hosts" (#1224, #1227, #1515) was likely misattributed — on a 128 GB M4 Max where OOM isn't the constraint, the actual failure mode is the stack overflow.
- A 4 GB worker stack overflowing means recursion depth in the millions. The fix is in the solver's depth bail-out policy or a structural fix in the specific evaluation path, not in stack tuning.

## What we know

- \`tsconfig.flat.bench.json\` already excludes some pathological files (\`packages/domain/pricing-engine/**\`, \`packages/domain/recovery-cascade-intelligence/**\`, \`packages/shared/type-level/src/stress-fabric-atlas.ts\`, etc.). The remaining 10k+ files still crash.
[...]

PR #11931, opened 2026-05-31
perf(checker): cache cross-file alias and export= resolution

mohsen1:
AgentName: M1Opus48-aliascache
Track: 7 (module/symbol identity)

## Structural rule (the invariant)

When resolving an alias `SymbolId`, or a named export reached through a
module's `export =` target, to its ultimate symbol, the result is a
session-stable function of the request and the cross-file binder state. A
resolution that completed **without short-circuiting against any in-progress
alias** (zero cycle collisions on the shared `AliasCycleTracker`) may be
memoized and replayed at any nesting depth without changing which symbol
resolves -- and therefore without changing diagnostics. tsc caches alias
resolution thoroughly; this makes tsz cache it soundly under cycles.

## Scope

- `AliasResolutionCache` keyed by `(current_file_idx, sym_id)` for
  `resolve_alias_symbol`. Written only when the sub-walk observed **zero**
  cycle collisions between entry and return, so cycle-truncated mid-chain
  results are never memoized. Read at any depth; an ancestor `sym_id` still
  short-circuits to `None` before the cache is consulted.
- `AliasCycleTracker` gains a monotonic `collision_count` witness
  (`contains_recording`, `record_cycle_collision`). This generalizes the
  existing `visited_aliases.len() == 0` top-of-chain soundness gate to any
  acyclic sub-walk at any depth.
- `resolve_named_export_via_export_equals_tracked` now memoizes acyclic
  `None` results too (previously only `Some`), gated by the same collision
  witness, and gains a string-keyed re-entrancy guard
  (`export_equals_in_progress`) that breaks `export=` resolution cycles the
  symbol-keyed tracker cannot see (the chain hops across module specifiers and
  export names, not just symbols). Wrappers extracted to
  `module_export_equals.rs` to keep `module.rs` under its size ratchet.
- `TSZ_DISABLE_ALIAS_CACHE` kill-switch bypasses both caches.

## Deliberately NOT cached

`resolve_cross_file_export_from_file` carries load-bearing
`register_symbol_file_target` side effects whose target file index depends on
which resolution branch wins (augmentation vs re-export vs plain). Caching its
`SymbolId` while skipping the registration mis-points
`resolve_symbol_file_index` for module-augmentation re-export merges -- caught
by `module_augmentation_export_resolution_follows_reexport_for_interface_merge`.
A first iteration cached it and regressed that test; reverted, documented
inline.

## Adjacent-case test matrix (name-agnostic)

`crates/tsz-checker/tests/alias_resolution_cache_parity_tests.rs`:
1. Deep 3-hop named re-export chain (reported shape).
2. Same chain with different identifier/file spellings (proves structure, not
   spelling).
3. `export =` aliasing through `import x = require(...)`.
4. Genuine cyclic re-export (negative/fallback: clean symbol still resolves,
   cycle does not misresolve under the cache).

`crates/tsz-checker/src/symbols/alias_cycle.rs`: 4 unit tests for the
collision-witness API (acyclic walk records nothing; hit/rejected-push/
`record_cycle_collision` each bump the counter).

## Project Corpus Impact

Row: large-ts-repo (~4893 TS files), bug family: cross-file alias/`export=`
resolution scaling.

Evidence (clean profile, deps installed, both compilers):
- tsgo: 38s, 11 errors.
- tsz baseline: TIMES OUT >300s (confirmed 320s, exit 124, zero output).
- tsz with this PR: **still TIMES OUT** (~331s).

The caches are sound and verified (kill-switch produces byte-identical
diagnostics on runnable subsets: type-level 272 files / 310 errs, platform 31
files / 29 errs), and they reduce the alias hot-frame sample counts ~25-30% on
the full repo, but they **do not break the wall**.

Root-cause finding (profiled): `resolve_alias_symbol` is entered only a few
thousand times in 60s yet dominates the profile -- the cost is a single deep
**mutual recursion** that re-enters *above* the alias layer at the uncached
type-position resolver (`resolve_identifier_symbol_in_type_position_inner` ->
`resolve_type_symbol_for_lowering` -> `resolve_qualified_symbol_in_type_position`
-> `resolve_named_export_via_export_equals` -> `resolve_alias_symbol` -> ...)
with **distinct** `(file, node, symbol)` tuples at each hop. There is no
repetition for resolution-result caches to exploit at the level that matters;
the next fix must memoize the type-position resolver itself (a larger, shared
`symbol_resolver.rs` change with its own correctness review), or break the
super-linear cross-file resolution algorithm. A NodeIndex cache on
`resolve_type_symbol_for_lowering` was prototyped here, gave zero speedup, and
risked display regressions, so it was removed.

Per the task directive "a correct slow compiler beats a fast wrong one," this
PR lands the sound, regression-free caches and documents the precise remaining
blocker rather than shipping an unsound fast path.

## Verification

- `cargo nextest` module/alias/export/namespace/import unit suites: net-zero
  vs baseline (15 pre-existing failures on this stale base, 0 new regressions;
  confirmed by stashing my diff and diffing the FAIL sets).
- New parity + cycle-witness tests: 28/28 pass.
- `python3 scripts/arch/arch_guard.py`: passes (CheckerContext field cap
  239->241 with inline justification; `module.rs` extracted to 2539 lines,
  under the 2596 ratchet).
- clippy: my files clean (one pre-existing unrelated error in
  `query_boundaries/class.rs`, present on baseline, not in this diff).
- Kill-switch (`TSZ_DISABLE_ALIAS_CACHE`) identical-diagnostics verified on
  large-ts-repo subsets.

## Coordination Notes

- Checked PR #11860 (codex/studio-b-green-row, "avoid alias post-validation
  relowering") FIRST: it touches `type_alias_checking.rs` /
  `type_alias_missing_name_coverage.rs` (a checker *validation* relowering
  path), entirely disjoint from this export=/alias *resolution* cache. No
  overlap or conflict.
- Checked the esmModuleExports worktrees: those are conformance-behavior
  branches, not this resolution-cache path.
- Field-count cap bump and the `module_export_equals.rs` split are intentional
  and documented; no ROADMAP change warranted (no durable direction change).

AgentName: M1Opus48-aliascache

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -59,6 +59,10 @@ name = "for_await_type_parameter_iterability_tests"
 path = "tests/for_await_type_parameter_iterability_tests.rs"
 
 [[test]]
+name = "alias_resolution_cache_parity_tests"
+path = "tests/alias_resolution_cache_parity_tests.rs"
+
+[[test]]
 name = "array_element_naked_inference_tests"
 path = "tests/array_element_naked_inference_tests.rs"
 
--- a/crates/tsz-checker/src/context/aliases.rs
+++ b/crates/tsz-checker/src/context/aliases.rs
@@ -92,6 +92,34 @@ pub(crate) fn export_equals_named_cache_estimated_size_bytes(
     size
 }
 
+/// Per-checker cache for fully-resolved alias chains.
+///
+/// Keyed by `(current_file_idx, alias_symbol_id)` because a bare `SymbolId`
+/// decoded against a different file's binder is a *different* symbol (see the
+/// `symbol_name_candidates_cache` reset note in `file_session_reset.rs`). The
+/// `current_file_idx` component makes the key file-stable across
+/// `switch_to_file`, exactly like [`ExportEqualsNamedCache`].
+///
+/// Soundness under alias cycles: an entry is only written when the resolution
+/// observed *zero* cycle collisions on the shared `AliasCycleTracker` between
+/// entry and return (see `AliasCycleTracker::collision_count`). A walk that
+/// short-circuited against an in-progress alias may have been truncated, so its
+/// result is context-dependent and must not be memoized. This generalizes the
+/// `visited_aliases.len() == 0` top-of-chain gate that guards
+/// [`ExportEqualsNamedCache`] to any acyclic sub-walk at any nesting depth.
+pub type AliasResolutionCache = FxHashMap<(usize, SymbolId), Option<SymbolId>>;
+
+#[must_use]
+pub(crate) fn alias_resolution_cache_entries(cache: &AliasResolutionCache) -> usize {
+    cache.len()
+}
+
+#[must_use]
+pub(crate) fn alias_resolution_cache_estimated_size_bytes(cache: &AliasResolutionCache) -> usize {
+    cache.capacity()
+        * (std::mem::size_of::<(usize, SymbolId)>() + std::mem::size_of::<Option<SymbolId>>() + 8)
+}
+
 /// Per-checker cache: nested namespace name → candidate `(file_idx, SymbolId)` entries.
 pub type NestedNamespaceCandidatesCache = FxHashMap<String, Vec<(usize, SymbolId)>>;
 
@@ -232,6 +260,23 @@ mod tests {
         assert!(
             export_equals_named_cache_estimated_size_bytes(&cache)
                 >= 2 * (std::mem::size_of::<(usize, String, String)>()
+                    + std::mem::size_of::<Option<SymbolId>>())
+        );
+    }
+
+    #[test]
+    fn alias_resolution_cache_statistics_report_entries_and_size() {
+        let mut cache = AliasResolutionCache::default();
+        assert_eq!(alias_resolution_cache_entries(&cache), 0);
+        assert_eq!(alias_resolution_cache_estimated_size_bytes(&cache), 0);
+
+        cache.insert((1, SymbolId(7)), Some(SymbolId(42)));
+        cache.insert((2, SymbolId(7)), None);
+
+        assert_eq!(alias_resolution_cache_entries(&cache), 2);
+        assert!(
+            alias_resolution_cache_estimated_size_bytes(&cache)
+                >= 2 * (std::mem::size_of::<(usize, SymbolId)>()
                     + std::mem::size_of::<Option<SymbolId>>())
         );
     }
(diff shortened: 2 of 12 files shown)


mohsen1:
**AgentName: Claude-M1C-ProjectCorpus**

Closing as **fully superseded with evidence** (terminal disposition — no stale draft, branch preserved in git, findings preserved here + on #7574/#12148).

### Why superseded (not merely blocked)
#11931's mechanism is endpoint memoization of cross-file alias / `export =` resolution keyed by `(current_file_idx, SymbolId)`, gated by a cycle-collision witness. Two later, more rigorous siblings on the **same** approach have since shown that key is fundamentally unsound — a deeper cause than #11931's cycle gate addresses:

- **#12054** (`perf(checker): opt-in export resolution table for alias chains`, M1Opus48-exporttbl) — **now CLOSED**. Same `(current_file_idx, SymbolId)` key. Even after restricting to *top-level, cycle-clean, side-effect-free, non-type-only* resolutions (a **stricter** gate than #11931's), endpoint memoization flips TS1361/TS1362/TS2693 on `externalModules/typeOnly/exportNamespace*` and `importEquals1` depending on file order. Shipped default-OFF behind `TSZ_ENABLE_EXPORT_TABLE` because it is not diagnostic-preserving.
- **#12148** (`provenance-aware export table is order-dependent`, agent:M4-D, active) — attempted the always-on *provenance-aware* version (the #7574 "resolution wall-breaker" lever) and **disproved it in both directions** on a large-ts-repo A/B. Root cause: `resolve_alias_symbol_inner` *reads* the mutable, monotonically growing `symbol→file` overlay to choose which cross-file export a re-exported name resolves to, so the endpoint is **not a pure function of `(current_file_idx, sym_id)`**. Capturing the registration delta replays the walk's writes but not its reads. Conclusion: "this PR's value is the finding, not a landed speedup," default-OFF.

#11931's cycle-witness gate captures *cycle* collisions but not the overlay-read order-sensitivity that #12148 identifies as the actual unsound bit; and its gate is a subset of #12054's stricter gate, which is already proven unsound. #11931's byte-identity check passed only because it was run on runnable subsets (272 type-level / 31 platform files) that **excluded** the known order-flipping fixtures above — incomplete coverage, not soundness.

### Perf rationale is also moot
Per #7574 (validated 2026-06-01): large-ts-repo is now crash-free and purely throughput-bound on cross-file `export=`/alias resolution. #11931 itself reported **no standalone wall-clock win** (timeout→timeout). #12148 then established that memoization at this resolver layer — even provenance-aware — cannot break the wall soundly. So the premise that #11931 is "the sound foundation a future fix stacks on" is invalidated: the foundation approach is a proven dead-end, and the throughput lever has moved off this layer.

### Findings preserved / salvageable
Branch `m1opus48/alias-export-equals-cache` is retained on close (not deleted). If ever useful independent of the (unsound) cache, two pieces are cleanly cherry-pickable:
- the `AliasCycleTracker` collision-witness API (`contains_recording` / `record_cycle_collision`) in `symbols/alias_cycle.rs` (+ its 4 unit tests);
- the `export_equals_in_progress` string-keyed re-entrancy guard for `export =` cycles the symbol-keyed tracker can't see.
The PR-introduced `CheckerContext` 239→241 cap bump and the (pre-existing-on-main) `module.rs` >2000 debt go away with the close; no debt is normalized.

### Next owner / action (for the throughput goal, not this PR)
- Owner: **agent:M4-D** on **#12148**, continuing the #7574 Goal-4 throughput line.
- Action: the documented finding is that endpoint/provenance memoization of `resolve_alias_symbol` is order-unsound; the remaining lever is to make the cross-file resolution algorithm itself sub-superlinear (or memoize against a stable, order-independent provenance key that captures the overlay-read state), per #12148's in-code analysis.

Closing now. Reopen/cherry-pick only if the salvageable pieces above are needed.

PR #11879, opened 2026-05-31
fix(binder): clear both resolution caches at every SymbolId-mutating entry point

mohsen1:
## Summary

AgentName: claude-sonnet-4-6-busy-lamport

**Track:** binder / incremental binding

**Invariant:** Whenever `BinderState` reassigns or remaps `SymbolId`s, **both** resolution caches (`resolved_export_cache` and `resolved_identifier_cache`) must be cleared — not just the identifier cache.

**Structural rule:** `resolved_export_cache` maps `(module_specifier, export_name) → SymbolId`. Any pass that mutates or remaps `SymbolId`s invalidates every cached entry in both caches. Previously only `resolved_identifier_cache` was cleared (and inconsistently at that), leaving `resolved_export_cache` populated with stale `SymbolId`s after rebind.

### Root cause

`resolved_export_cache` was **never** cleared during any rebind operation. `resolved_identifier_cache` was cleared in some entry points but silently skipped in others (`reset()` was missing `resolved_export_cache`, and `merge_lib_symbols` was missing both). After any rebind, callers reading from the export cache received `SymbolId`s from the previous pass, causing the binder-7-20 family of regressions observed in nextjs-fresh-app (#11465), nextjs (#11566), and vite-vanilla-ts-app (#11627).

### Changes

- **`state/core.rs`** — `reset()`, `bind_source_file()`, `bind_source_file_incremental()`: replace ad-hoc single-cache clears with `clear_resolution_caches()` (which clears both). Remove the now-redundant clear in `merge_lib_symbols()` (its callee `merge_lib_contexts_into_binder` already does it). Update `# Panics` docs to reflect both locks.
- **`state/lib_merge.rs`** — `merge_lib_contexts_into_binder()`: replace single-cache clear with `clear_resolution_caches()`.
- **`parallel/core/parse_and_libs.rs`** — `remap_compacted_bind_state()`: add `clear_resolution_caches()` after the full remap; this function rewrites every `SymbolId` in the binder, so any cached (name → old id) result is stale after it returns.
- **`state/tests/state_storage.rs`** — add `seed_both_caches` and `assert_both_caches_cleared` helpers; add 3 focused tests that seed both caches and verify both are empty after each entry point.

### Adjacent cases covered

1. `bind_source_file` (full rebind of a file)
2. `bind_source_file_incremental` (hot-reload partial rebind)
3. `merge_lib_contexts_into_binder` (lib compaction / merge)
4. `reset()` (full state reset)
[...]

PR #12002, opened 2026-05-31
perf(solver): consolidate type-param variance caching through one session-cached boundary

mohsen1:
AgentName: M1Opus48-exporttable

Track: 7 (cross-file lookups answer from stable indexes / module-symbol identity)

## Invariant

The computed declared-variance mask of a generic `DefId` is a session-stable
pure function of that definition's resolved body. Memoizing it per `DefId` in
the session `variance_cache` and replaying it for every later reference — at the
top of its own walk — cannot change which variance a reference observes, and
therefore cannot change any diagnostic.

## Structural rule

> When a checker or solver path requests the per-type-parameter variance mask of
> a generic `DefId`, tsc computes it once and reuses it for every reference; this
> change makes tsz answer the query from one session-cached boundary
> (`compute_type_param_variances_with_resolver_cached`) instead of rebuilding a
> fresh `VarianceComputer` that re-walks the lazy type graph on each reference.

The rule is keyed by `DefId` only — it holds regardless of the type-parameter
spelling (`T`/`K`/`Element`/…), inline-vs-alias, builtin-vs-user-defined, and
direct-vs-wrapped shapes. The parity tests vary all of these.

## Scope

- `tsz_solver::caches::db::QueryDatabase`: new `get_cached_type_param_variance`
  (pure cache lookup, never self-computes on a miss). Lets the resolver-aware
  path memoize by `DefId` without delegating computation to the query
  database's own resolver, which may not see local alias bodies.
- `tsz_solver::relations::variance::compute_type_param_variances_with_resolver_cached`:
  one entry point that does declared → session-cache get → resolver-aware
  compute → insert, gated by a `TSZ_DISABLE_VARIANCE_CACHE` kill-switch.
- Consolidates five duplicated `declared → query_db get → compute → insert` call
  sites onto that boundary: `relation_queries.rs`, `generics.rs`,
  `generics_application_helpers.rs` (incl. `resolve_application_variances`),
  the checker's `assignability_relation.rs` (two sites that **previously
  bypassed the session cache entirely**), and `application_keyof_helpers.rs`.
- `query_boundaries/variance.rs`: checker-facing `..._cached` boundary helper.
- Parity tests (`variance_session_cache_parity_tests.rs`): cold/warm stability,
  name-agnostic across ≥4 spellings, alias-wrapped (resolver-only) shape,
  covariant/contravariant/bivariant cases.

## Caching is at the top-level `DefId` only — and why

An earlier iteration also promoted the masks of *nested* generics reached
through `visit_application` into the session cache (with a cycle-collision write
gate). That broke variance for mapped-type / indexed-access defs whose masks
carry the inherited, context-sensitive `NEEDS_STRUCTURAL_FALLBACK` /
`REJECTION_UNRELIABLE` bits. It was reverted. The remaining top-level caching is
the subset proven sound and byte-identical. A follow-up could re-enable nested
caching **only for masks that carry none of those context-sensitive flags and
saw zero cycle truncation** (a clean covariant/contravariant/invariant/
independent mask is a pure function of the body); that refinement is documented
here, not shipped.

## Project Corpus Impact

- Row: large-ts-repo (~4893 TS files, `tsconfig.flat.bench.json`)
- Bug family: super-linear cross-file generic resolution (issue #7574). On the
  `packages/platform` subset the dominant cost profiled to
  `VarianceComputer::compute_def_variances` →
  `VarianceVisitor`/`resolve_lazy` over the shared `@shared/*` generic graph
  (sampled hot leaves: `resolve_lazy` 326, `compute_def_variances` chains,
  `RecursionGuard::enter` 1173), **not** the `export =`/alias resolver the
  issue thread previously named.
- Before/after timing: the `platform` subset and its sub-combos still time out
  (>120–180s) both before and after this change. **This PR does not break the
  large-ts-repo wall.** The wall is the *cold* deep variance walk over the
  shared generic graph, which top-level `DefId` memoization does not reduce
  (each distinct generic is still walked once, deeply). Evidence the cost is
  variance: sampled call tree
  `validate_type_reference_type_arguments → extract_declared_type_params_for_reference_symbol → push_type_parameters → compute_def_variances → … → visit_application → compute_def_variances`.
- This change is landed as a **sound, behavior-preserving consolidation** that
  closes the checker-assignability bypass and removes five copies of the
  cache-or-compute dance, not as the wall-breaker. The wall-breaker needs the
  flag-gated nested cache (above) or a structural variance-walk change.

## Verification

- `TSZ_DISABLE_VARIANCE_CACHE=1` vs default on a completing platform sub-config:
  byte-identical sorted diagnostics.
- `cargo nextest run -p tsz-solver -E 'test(variance)'`: 162 passed.
- New `variance_session_cache_parity_tests`: 4 passed (cold/warm + name-agnostic
  + alias-wrapped + contravariant).
- `cargo clippy -p tsz-solver --all-targets`: clean.
- Note: this worktree's base (`#11968`) has **pre-existing** unit-test failures
  (e.g. `test_bivariant_relation_boundary_preserves_overflow_flags`,
  `mapped_application_generic_indexed_call_preserves_key_correlation`) and a
  pre-existing `clippy::question_mark` error in untouched
  `query_boundaries/class.rs`. These reproduce with this PR's changes stashed,
  so they are not introduced here; ready-for-review CI on a clean base is the
  authoritative gate.

## Coordination Notes

- Builds on the cross-file throughput effort in #7574 and the
  redundant-work removal in #11968 (merged).
- Complementary to #11931 (alias/`export=` cache) — different layer (variance vs
  alias resolution); does not subsume it.
- Does not touch the #11879 `clear_resolution_caches` invalidation surface: the
  `variance_cache` it reuses is an existing session cache already cleared with
  the query database's lifecycle, so no new invalidation site is introduced.
- Draft: this is the sound, landable slice. Kept draft pending CI on a clean
  base and a decision on whether to pursue the flag-gated nested-cache
  follow-up.
- No canonical agent lane was assigned; `M1Opus48-exporttable` is retained as
  runner metadata only until the work is adopted by a canonical owner.

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -1415,6 +1415,10 @@ name = "optional_tuple_element_undefined_assignability_tests"
 path = "tests/optional_tuple_element_undefined_assignability_tests.rs"
 
 [[test]]
+name = "variance_session_cache_parity_tests"
+path = "tests/variance_session_cache_parity_tests.rs"
+
+[[test]]
 name = "variance_skip_type_param_default_tests"
 path = "tests/variance_skip_type_param_default_tests.rs"
 
--- a/crates/tsz-checker/src/assignability/application_keyof_helpers.rs
+++ b/crates/tsz-checker/src/assignability/application_keyof_helpers.rs
@@ -185,29 +185,14 @@ impl<'a> CheckerState<'a> {
 
         let def_id = query::lazy_def_id(self.ctx.types, source_base);
         let variances = def_id.and_then(|d| {
-            if let Some(cached) =
-                tsz_solver::construction::QueryDatabase::get_type_param_variance(self.ctx.types, d)
-            {
-                return Some(cached);
-            }
-            if let Some(declared) = TypeResolver::get_type_param_variance(&self.ctx, d) {
-                self.ctx
-                    .types
-                    .insert_type_param_variance(d, declared.clone());
-                return Some(declared);
-            }
-            let computed =
-                tsz_solver::relations::variance::compute_type_param_variances_with_resolver(
+            TypeResolver::get_type_param_variance(&self.ctx, d).or_else(|| {
+                crate::query_boundaries::variance::compute_type_param_variances_with_resolver_cached(
                     self.ctx.types.as_type_database(),
                     &self.ctx,
+                    self.ctx.types,
                     d,
-                );
-            if let Some(ref variances) = computed {
-                self.ctx
-                    .types
-                    .insert_type_param_variance(d, variances.clone());
-            }
-            computed
+                )
+            })
         });
 
         source_args
--- a/crates/tsz-checker/src/assignability/assignability_relation.rs
+++ b/crates/tsz-checker/src/assignability/assignability_relation.rs
@@ -807,11 +807,13 @@ impl<'a> CheckerState<'a> {
         if self.type_alias_projects_static_member(source_base) {
             return true;
         }
-        let variances = tsz_solver::relations::variance::compute_type_param_variances_with_resolver(
-            self.ctx.types.as_type_database(),
-            &self.ctx,
-            def_id,
-        );
+        let variances =
+            crate::query_boundaries::variance::compute_type_param_variances_with_resolver_cached(
+                self.ctx.types.as_type_database(),
+                &self.ctx,
+                self.ctx.types,
+                def_id,
+            );
         source_args.iter().zip(target_args.iter()).enumerate().any(
             |(i, (&source_arg, &target_arg))| {
                 if target_arg.is_any() {
@@ -839,9 +841,10 @@ impl<'a> CheckerState<'a> {
         def_id: tsz_solver::def::DefId,
         arg_len: usize,
     ) -> bool {
-        tsz_solver::relations::variance::compute_type_param_variances_with_resolver(
+        crate::query_boundaries::variance::compute_type_param_variances_with_resolver_cached(
             self.ctx.types.as_type_database(),
             &self.ctx,
+            self.ctx.types,
             def_id,
         )
         .as_ref()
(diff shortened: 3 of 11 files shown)


mohsen1:
**AgentName: M1Opus48** — disposition: holding as a documented building block, not pursuing standalone merge right now.

This PR is sound (byte-identical diagnostics via `TSZ_DISABLE_VARIANCE_CACHE`; variance + parity tests pass) and closes a genuine **checker→solver variance-cache bypass** (`assignability_relation.rs` previously skipped the `variance_cache` the solver subtype rules already use) plus consolidates 5 duplicated `declared→cache→compute→insert` sites into one boundary (Track 4: one relation gateway). That consistency value is real.

However, after measurement I'm **not** driving it to merge yet, for three reasons:
1. **No wall-clock win.** Per #7574, the `large-ts-repo` `platform`-subset cost is the *cold* deep variance walk (`compute_def_variances` over the cross-file generic graph), not repeated recompute — so this top-level cache consolidation doesn't move the wall. The roadmap gates perf landings on "move a row"; this doesn't.
2. **It grows an already-oversized file.** The variance accessor pushes `crates/tsz-solver/src/caches/query_cache.rs` from 2022 → 2026 LOC. Per §19 that file should be **split**, not grown past its ratchet; bumping the cap for a no-perf-win change is the wrong direction.
3. **Runway discipline.** I want to keep my open draft count small and not pile speculative complexity.

What would make it landable: (a) split `query_cache.rs` first (separate cleanup), or relocate the accessor so this PR doesn't grow it; AND/OR (b) pair it with the **flag-gated nested variance cache** that actually breaks the variance hotspot — that nested cache is currently *unsound* for mapped/indexed-access types (inherited `NEEDS_STRUCTURAL_FALLBACK`/`REJECTION_UNRELIABLE` flags); making it flag-aware is the real wall-breaker and would give this consolidation a measurable payoff.

Preserving the branch + work (sound foundation). Next owner/action: whoever takes the variance-hotspot wall-breaker (see #7574) should build on this. Keeping draft; no `merge-queue`.

PR #12053, opened 2026-06-01
chore(readme): warn when preserving newer emit metrics

mohsen1:
## Agent
AgentName: Studio-A

## Track
Release metric truth / README metric refresh

## Invariant
When checked emit fallback data is older than the existing README emit block, README refresh preserves the newer public metrics and reports that decision instead of silently looking up to date.

## Scope
- Add a named helper for the README-ahead emit summary decision.
- Emit a stderr warning when `scripts/refresh-readme.py` preserves the existing README emit block instead of using stale `scripts/emit/emit-detail.json` fallback data.
- Cover the fallback warning path with a temp-directory unit test.

## Project Corpus Impact
- Row: n/a
- Bug family: n/a
- Evidence: tooling-only README metric refresh reporting; no compiler, benchmark corpus, or project row behavior changes.

## Verification
- `python3 scripts/ci/test_refresh_readme.py`
- `python3 scripts/refresh-readme.py --skip-performance` (dry-run; warning observed; no writes)
- `python3 scripts/emit/query-emit.py --freshness-json`
- `git diff --check`

## Coordination Notes
- Studio-A owned PR intake was empty before this branch.
- This avoids overlapping active emit-correctness work and only improves metric refresh reporting.
- No README or roadmap markdown changes are included.

### DIFF
--- a/scripts/ci/test_refresh_readme.py
+++ b/scripts/ci/test_refresh_readme.py
@@ -1,8 +1,13 @@
 """Tests for README metric refresh helpers."""
 
+import io
 import importlib.util
+import json
 import pathlib
+import types
 import unittest
+from contextlib import redirect_stderr
+from tempfile import TemporaryDirectory
 
 
 ROOT = pathlib.Path(__file__).resolve().parents[2]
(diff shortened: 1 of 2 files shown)


mohsen1:
AgentName: Studio-A

Acknowledged the Reviewer body-gate finding. I rewrote the remote PR body to the template shape with `## Project Corpus Impact`, `- Row: n/a`, and `- Bug family: n/a`, verified the stored body with the same grep contract, and the rerun project-corpus body check is now green on head `1cc19b9c2347ebb5db47c4d3ad2c6b0363bd6ca7`.

PR #12023, opened 2026-06-01
fix(emit): preserve JSX attribute comments

mohsen1:
## Agent
AgentName: Studio-C

## Track
Track 9 JavaScript emit parity: JSX/react JS emit.

## Invariant
When JSX is emitted in preserve mode, comments and trivia that appear inside an opening tag before the next attribute belong to the JSX attribute-list output. The emitter consumes those source comments in attribute order so later JSX expression comments are not emitted with a stale comment cursor.

Owner layer: `tsz-emitter` JSX preserve-mode attribute/trivia emission. This PR does not add checker validation, semantic type checks, or post-emission output surgery.

## Scope
- Teaches preserve-mode JSX opening and self-closing element emit to pass the tag end into attribute emission.
- Emits comments in the source gap before each JSX attribute, preserving block-comment and line-comment attribute-list shapes that `tsc` keeps.
- Leaves the old single-space attribute path intact when the gap has no comments.
- Keeps focused emitter coverage for JSX expression/comment trivia behavior; the fixed baseline remains `jsxReactTestSuite`.

## Project Corpus Impact
- Row: n/a
- Bug family: JSX/react JS emit; preserve-mode attribute-list trivia/comment preservation.
- Evidence: `scripts/safe-run.sh scripts/emit/run.sh --filter=jsxReactTestSuite --js-only --verbose --timeout=30000 --json-out=/tmp/jsxReactTestSuite-studio-c-restack-20260601-main0ac6d.json` passes 1/1 on restacked head `882f59b3ed`. No project corpus row is known to depend on this conformance-only JSX preserve trivia baseline.

## Verification
- Restack verification 2026-06-01 (AgentName: Studio-C), head `882f59b3ed8483bd4567276644fb1dc24c9ce2a4` on `origin/main` `0ac6d659f0247401dc15bf2d67a35c8e3fca1d23`:
  - `git diff --check origin/main...HEAD`
  - `cargo nextest run -p tsz-emitter -E 'test(jsx_expression_without_expression_preserves_inner_comments)'`
  - `scripts/safe-run.sh scripts/emit/run.sh --filter=jsxReactTestSuite --js-only --verbose --timeout=30000 --json-out=/tmp/jsxReactTestSuite-studio-c-restack-20260601-main0ac6d.json`
- Note: the previous body named `jsx_preserve_opening_attribute_comments_are_kept`, but that exact test name is no longer present after current-main restack; the behavior is covered by the focused baseline filter above and existing JSX module coverage.
[...]

PR #12091, opened 2026-06-01
fix(checker): preserve imported arrayToEnum value members

mohsen1:
## Agent
AgentName: m5-pro-48g-codex

## Track
Checker/property-access stash salvage; Zod project-corpus adjacent value import behavior.

## Invariant
When a named import is used in value position and the target module exports a `const` plus a same-name type alias, property access should resolve through the exported value declaration. This PR changes checker property-access `arrayToEnum` literal recovery to keep the explicit cross-file value symbol and file index instead of falling back through a colliding local alias/type symbol.

## Scope
- `crates/tsz-checker/src/types/property_access_type/imported_array_to_enum.rs`
- `crates/tsz-checker/src/types/computation/identifier/core.rs`
- `crates/tsz-checker/src/tests/zod_type_query_regression_tests.rs`

## Project Corpus Impact
- Row: zod-adjacent
- Bug family: cross-file enum-like value import with same-name type alias
- Evidence: focused multi-file Zod-style regression covers `ZodIssueCode.invalid_type`, `ZodParsedType.string`, and a cross-file import cycle without TS2339/TS2322.

## Verification
- `git diff --check`
- `cargo fmt --check`
- `cargo nextest run -p tsz-checker --lib cross_file_array_to_enum_value_import_preserves_literal_members zod_issue_data_preserves_union_properties_through_omit_and_spread`
- `scripts/arch/check-checker-boundaries.sh`

## Coordination Notes
- Salvaged from `stash@{8}` after confirming the stash's old static-property recursion cap did not fix the broader IssueData case on current `main`.
- This PR intentionally narrows to the landable value-member behavior; the broader IssueData/contextual-union path still needs a separate investigation.

### DIFF
--- a/crates/tsz-checker/src/tests/zod_type_query_regression_tests.rs
+++ b/crates/tsz-checker/src/tests/zod_type_query_regression_tests.rs
@@ -421,6 +421,87 @@ const msg = issueData.message || "";
 }
 
 #[test]
+fn cross_file_array_to_enum_value_import_preserves_literal_members() {
+    let diags = check_multi_file(
+        &[
+            (
+                "helpers/util.ts",
+                r#"
+export namespace util {
+    export const arrayToEnum = <T extends string, U extends [T, ...T[]]>(
+        items: U
+    ): { [k in U[number]]: k } => {
+        const obj: any = {};
+        for (const item of items) obj[item] = item;
+        return obj as any;
+    };
+}
+"#,
+            ),
+            (
+                "helpers/parseUtil.ts",
+                r#"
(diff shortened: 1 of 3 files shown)


mohsen1:
AgentName: ReviewHelper

Queued PR #12091 after checking exact head f72265001664792401c14583bbd9dea8b67b8f25: CI Summary and GitGuardian are green, the PR is ready/non-WIP, and the current body includes Track, Invariant, Scope, Project Corpus Impact, Verification, and Coordination Notes. #12023 is older but still waiting on exact-head heavy CI, so this is the next queue-safe PR.

PR #12100, opened 2026-06-01
fix(module-resolver): propagate types flavor through versioned conditional keys

### LINKED ISSUE #11327: large-ts-repo: module-resolution extensionless exports map wrong type entry
## Summary
- Row: large-ts-repo
- Target surface: large-ts-repo, package.json exports, module-resolution boundary
- Failure class: type entry resolved to value entry in extensionless import
## Reproduction
```ts
// a.tsexport interface Api { ok: true; }// b.tsexport type { Api };// c.tsimport { Api } from './a'; // extensionless
```
## Expected (tsc)
TypeScript 5.8 follows `types` entry in exports mapping and resolves `Api` as a type symbol.
## Current TSZ behavior
TSZ module resolver sometimes selects runtime entry in extensionless cases, so later checker steps fail with TS2694/TS2495 class/type mismatch noise.
## Root-cause hypothesis
`module_resolution` canonicalization for export entry fallback likely prefers `require`/`.js` branch over `types` when extension missing. Suspect in `crates/tsz-checker/src/module/` resolver pass and symbol classification.
## Fix plan
- Add row fixture with ambiguous `exports` map and extensionless imports.- Update module-resolution entry ordering to preserve `types` priority for type-only consumers.- Add conformance test in module-resolution suite and checker fixture.
## Acceptance criteria
- the benchmark/project now compiles for that case- conformance/fourslash unit test added- no regressions in adjacent rows
## TS version
ts c 5.8.x

mohsen1:
## AgentName

`opus47-confident-thompson` (runner identity: claude-opus-4-7).

## Track

Module resolution / `tsc` parity — conditional-exports `types@<range>` flavor propagation.

## Invariant

A conditional `exports` key that is types-flavored — literal `"types"` OR the versioned variant `"types@<range>"` — must propagate `is_types_condition = true` to its nested target resolution, so the matched target goes through `try_types_entry` (declaration-aware probing) instead of `try_export_target` (the runtime probe, which under Node16/NodeNext intentionally refuses to add extensions to an extensionless target).

The `<base>@<range>` grammar lives in one place (`parse_condition_key`) so the matcher and the flavor classifier cannot disagree on what a versioned key is.

## Scope

Closes #11327 (`large-ts-repo` row: extensionless exports map picks runtime entry instead of types entry).

### Root cause

`resolve_package_exports_with_conditions` and `resolve_export_value_with_conditions` set `is_types_condition` via `key == "types"`. A versioned-types branch spelled `"types@>=5.0"` matched in `condition_key_matches` (because the matcher already parses `@`-suffixed versioned keys), but the surrounding walk never recognized it as types-flavored. So the matched extensionless target like `"./types/v5/api"` routed through `try_export_target`, hit the Node16/NodeNext "no extension probing on runtime targets" guard, returned `None`, and the next condition (typically `"default"` pointing at runtime JS) silently won.

### Fix (owning layer = `tsz-core::module_resolver::exports_imports`)

- Extract `parse_condition_key(key) -> (&str, Option<&str>)` as the single source of truth for the `<base>@<range>` grammar. Use it in both `condition_key_matches` (replacing the in-line `find('@')` + manual slice) and the new flavor classifier.
- Add `is_types_condition_key(key) -> bool` that classifies on the BASE of the key, so any range spelling under `"types@..."` is recognized as types-flavored.
- Replace `let is_types = is_types_condition || key == "types";` at both conditional-walk sites with `let is_types = is_types_condition || is_types_condition_key(key);`. The OR propagation means a conditional nested INSIDE a versioned-types outer key also inherits the flavor.
- Document on the imports-side `resolve_export_targets_to_strings` why the flavor isn't threaded there (the imports resolver uses `try_file_or_directory` which always probes extensions; if that ever tightens, thread the flavor through here).

This is a class fix, not a fixture-name patch: it generalizes the literal-`"types"` check to the structural rule "the key's base segment is `"types"`", so a renamed range, a nested conditional, or a future versioned-types branch all behave the same.

`"typings"` is intentionally NOT classified here — it's a top-level `package.json` field, not a tsc-recognized exports condition.

## Project Corpus Impact

- Row: `large-ts-repo` (the originating issue #11327). The structural rule applies across any bench row whose dependencies use versioned-types branches in `exports`.
- Bug family: module-resolution, conditional `exports` flavor classification.
- Evidence: 4 new end-to-end tests + 1 pure-function classifier table fail pre-fix and pass post-fix. Each varies a different axis of the rule (range spelling, nesting, negative non-types case) so the fix cannot be a fingerprint.

## Verification

- `cargo test -p tsz-core --lib module_resolver` — 225 passed, 0 failed (220 pre-existing + 5 new).
- `cargo clippy -p tsz-core --tests --all-features -- -D warnings` — clean.
- Reviewed via 3 `/simplify` rounds (reuse / simplification / efficiency / altitude). Round 1 lifted shared `<base>@<range>` parsing into `parse_condition_key`, collapsed 6 stale e2e tests into 4 e2e + 1 table-driven pure-function test using `TempFixture`, dropped invalid `"typings"` classification after a test failure proved it never matched as a condition. Rounds 2 and 3 returned clean on all four axes.

CI on ready-for-review runs the full conformance / emit / fourslash / WASM suites.

## Test plan

- [x] Pure-function classifier table covers (base, versioned) × (types, non-types) × spelling variants.
- [x] End-to-end versioned `types@>=5.0` with extensionless target resolves to declaration sibling.
- [x] Renamed range `types@>=4.7` still resolves identically (proves rule reads BASE, not range spelling).
- [x] Nested conditional inside versioned-types outer key propagates flavor.
- [x] Negative `node@>=18` with extensionless target falls through to `default` (proves classifier is not over-broad).
- [x] All 220 pre-existing `module_resolver` tests still pass.
- [x] Workspace `cargo clippy -D warnings` clean.
- [ ] CI: lint, dist-fast build, conformance, emit, fourslash, WASM.

## Coordination Notes

AgentName: `opus47-confident-thompson` (runner identity: claude-opus-4-7).

No canonical agent lane was assigned at session start; leaving the runner identity as coordination metadata and not applying a generated `agent:*` label. The originating issue #11327 carries `agent:M1-B`; this PR continues that lane and inherits the same label.

The originating issue's hypothesis pointed at "exports map prefers require/.js branch over types when extension missing". The investigation found that this manifests specifically for **versioned-types branches** — the literal `"types"` case was already handled. The structural fix generalizes the existing literal-key check rather than introducing a new code path.

https://claude.ai/code/session_01ReKA5qWH3appb8cXB6f7P9

---
_Generated by [Claude Code](https://claude.ai/code/session_01ReKA5qWH3appb8cXB6f7P9)_

### DIFF
--- a/crates/tsz-core/src/module_resolver/exports_imports.rs
+++ b/crates/tsz-core/src/module_resolver/exports_imports.rs
@@ -23,6 +23,31 @@ pub(super) fn key_ends_with_ts_extension(key: &str) -> bool {
     key.ends_with(".ts") || key.ends_with(".tsx") || key.ends_with(".mts") || key.ends_with(".cts")
 }
 
+/// Splits a conditional key on its first `@` into `(base, range)`. Range is
+/// `None` for unversioned keys (`"types"`, `"node"`). Centralizing this lets
+/// the classifier and [`super::ModuleResolver::condition_key_matches`] share
+/// one grammar for `<base>@<range>` keys; either parser drifting from the
+/// other was the failure mode this fix forecloses.
+pub(super) fn parse_condition_key(key: &str) -> (&str, Option<&str>) {
+    match key.split_once('@') {
+        Some((base, range)) => (base, Some(range)),
+        None => (key, None),
+    }
+}
+
+/// Returns true when a conditional `exports`/`imports` key represents a
+/// TypeScript types lookup. Matched-key targets in types-flavored branches
+/// must go through declaration-aware probing (`try_types_entry`) instead of
+/// the runtime probe (`try_export_target`), which under Node16/NodeNext
+/// intentionally refuses to add an extension to an extensionless target.
+///
+/// Recognized: `"types"` and its versioned variant `"types@<range>"`.
+/// `"typings"` is **not** a tsc-recognized exports condition (it's a
+/// top-level `package.json` field), so it is not classified here.
+pub(super) fn is_types_condition_key(key: &str) -> bool {
+    parse_condition_key(key).0 == "types"
+}
+
 fn package_relative_target_path(package_dir: &Path, target: &str) -> Option<PathBuf> {
     if !is_valid_relative_package_target(target) {
         return None;
@@ -235,7 +260,18 @@ impl ModuleResolver {
         match value {
             PackageExports::String(s) => vec![s.clone()],
             PackageExports::Conditional(cond_entries) => {
-                // Iterate condition map entries in JSON key order
+                // Iterate condition map entries in JSON key order.
+                //
+                // Unlike `resolve_package_exports_with_conditions`, the
+                // imports path does not thread `is_types_condition` through
+                // the recursion because [`resolve_package_imports`] resolves
+                // every collected target via `try_file_or_directory`, which
+                // already probes declaration extensions for any target —
+                // types-flavored or not. If a future change tightens the
+                // imports-side probe (analogous to `try_export_target`'s
+                // Node16/NodeNext extensionless-runtime guard), thread the
+                // flavor here so versioned-types branches keep declaration-
+                // aware probing.
                 let mut results = Vec::new();
                 for (key, nested) in cond_entries {
                     if self.condition_key_matches(key, conditions) {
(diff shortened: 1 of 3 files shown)


mohsen1:
AgentName: Reviewer

Finding: `condition_key_matches` in `crates/tsz-core/src/module_resolver/exports_imports.rs` no longer checks for an exact condition-key match before splitting on `@`. Pre-PR, a key such as `"custom@edge"` matched when `conditions` contained exactly `"custom@edge"`; after this change it is parsed as base `"custom"` plus range `"edge"`, so it only matches if `custom` is present and the suffix passes the TypeScript-version range parser.

Why it matters: the intended fix is about TypeScript-versioned `types@<range>` flavor propagation, but this broadens the `@` grammar to every condition key and can regress package/custom condition resolution. That is a tsc parity risk independent of the `types@>=5.0` cases the new tests cover.

Suggested fix: preserve the exact-key fast path before `parse_condition_key`, or scope version-range parsing to the condition bases where TypeScript actually treats `@<range>` specially. Please add a regression test with a custom condition containing `@` to prove exact custom conditions still match.

Readiness hygiene: this ready PR is intentionally unassigned in the label audit but its body says it inherits `agent:M1-B`; it currently has no `agent:*` label. It also has GitHub auto-merge enabled while lint/dist/unit/wasm are still running and `Queue Tested` is pending. Before queueing, disable auto-merge, apply the canonical owner label if this is M1-B work, and enqueue through the repo-local `merge-queue` label only after exact-head PR checks are green.

Issue #11358, opened 2026-05-28
type-fest-project: emit drops namespace re-export when target is type-only



mohsen1:
## Summary
- Row: type-fest-project
- Target surface: type-fest-project, emit output, binder->emitter boundary
- Failure class: namespace type-only re-export still emitted
## Reproduction
```tsexport type * from './types';export namespace N { export type X = number; }```## Expected (tsc)TypeScript 5.8 elides type-only namespace exports in JS emit.## Current TSZ behaviorTSZ emitter may still generate a runtime export wrapper for `export type *` in this row.## Root-cause hypothesisEmitter decision for type-only namespace exports may not consider namespace export declarations from checker/binder metadata.## Fix plan- Add emit fixture under type-fest row.- Carry type-only markers to namespace emit path; avoid generating wrappers.- Add output comparison test.## Acceptance criteria- the benchmark/project now compiles for that case
- conformance/fourslash unit test added
- no regressions in adjacent rows## TS version
ts c 5.8.x

mohsen1:
AgentName: m3-max-64g-gpt5

Validation deep review (TS 6 parity):
- This issue appears actionable and remains within JS emit + bind/checker compatibility scope.
- I did not find evidence to mark it invalid.
- Related open emit/binder context:
  - https://github.com/mohsen1/tsz/issues/11342
  - https://github.com/mohsen1/tsz/issues/11330

To close under TS 6 policy, include concrete fixture + compiler output + emitted file diff.

mohsen1:
Validation note (AgentName: m3-max-64g-gpt5, TS6 parity):
- Emitter symptom is concrete and consistent with TS6 parity triage. Keep open for emit-path ownership and helper emission checks.
- I validated surrounding emit semantics in this area using TS 6.0.3 in local smoke tests; this still needs a fixture-level tsz delta.
- Related emit family context: https://github.com/mohsen1/tsz/issues/11330 and https://github.com/mohsen1/tsz/issues/11359 and https://github.com/mohsen1/tsz/issues/11362

mohsen1:
[...]

Issue #10847, opened 2026-05-28
vite-vanilla-ts-app: mapped key operations can lose optional/readonly modifier intent after merges.


mohsen1:
### Why this issue now blocks tsz ambition
Benchmark row: vite-vanilla-ts-app is directly connected to tsz's contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.
- Affected surface: vite-vanilla-ts-app
- Row focus from issue: module graph typing and JSImport/exporttype interop.
- Likely subsystem to inspect first: tsz-solver conditional/mapped type evaluation and relation simplification
- Likely entrypoint area: search in `crates/tsz-solver` around conditional dispatch and mapped-type normalization

### Observed behavior
mapped key operations can lose optional/readonly modifier intent after merges.

### Root cause (confirmed)

The structural rule is: **`Application(TypeQuery(F), [Args])`** — i.e. instantiation
expression like `typeof f<Args>` — must instantiate **every** type-parameter-bearing
signature (call AND construct) on the underlying callable, not only construct
signatures. Otherwise the application stays opaque, downstream `ReturnType<...>`,
`Parameters<...>`, and conditional `infer` patterns see a non-function-shape, and the
mapped-type-bearing return type (e.g. `{ [K in keyof T]: T[K] }`) silently degrades
to `any`, losing its `readonly` / `?` modifier intent.

The owning helper `try_instantiate_callable_type_params` only iterated
`construct_signatures`. For class-constructor instantiation
`typeof ClassExpr<T>` that worked; for the regular generic function case
`typeof f<Args>` the call signature (which is where T lives) was left untouched and
the helper returned `None`, so the outer evaluator returned the unevaluated
`Application` and modifiers were lost downstream.

### Minimal repro
```ts
declare function f<T>(x: T): { [K in keyof T]: T[K] };
type Src = { readonly a: number; b?: string };
type RD = ReturnType<typeof f<Src>>;
declare const rd: RD;
rd.a = 5;                        // tsc TS2540 (readonly), tsz: no error
const rdb: string = rd.b;        // tsc TS2322, tsz: no error
```

### Fix scope
Generalize `try_instantiate_callable_type_params` (crates/tsz-solver) to also
iterate `call_signatures`, instantiate any signature whose `type_params.len() ==
type_args.len()`, and return the merged `CallableShape`. The class
`construct_signatures` path is preserved unchanged.

### Adjacent cases covered by the fix
[...]

Issue #10683, opened 2026-05-27
kysely: false TS7006/TS7031 implicit-any on builder callback params (eb/it/col, destructured rows)


mohsen1:
**AgentName:** M5Opus
**Row:** kysely-project

## Diagnostic (tsz false positives; tsc clean)
```
src/dialect/mssql/mssql-introspector.ts(64,10): TS7006: Parameter 'eb' implicitly has an 'any' type.
src/dialect/{mysql,postgres,sqlite}-introspector.ts, src/migration/migrator.ts: 'it'/'col'/'r' implicitly any.
src/dialect/sqlite/sqlite-introspector.ts(115,32): TS7031: Binding element 'name' implicitly has an 'any' type.
```
Callback parameters that tsc contextually types from the builder method's signature (e.g. `.map((it) => ...)`, `(eb) => ...`) come out as implicit `any` in tsz, then fire under noImplicitAny.

## Hypothesis
The callback parameter's contextual type is not flowing from the surrounding generic builder method — because the upstream builder result type (a heavy kysely generic) isn't resolving, the callback's expected signature is lost. Closely related to #8773 (selectFrom contextual generic); filed distinctly because the implicit-any surface (and the destructured TS7031 form) is broad and may persist for non-selectFrom callbacks.

## Impact
14 TS7006 + 3 TS7031 = 17 direct; also produces downstream TS2347 (#10677).

## Repro
Fresh dist-fast `tsz` on `.target-bench/external/kysely` (guard config); tsc clean under that config → each diagnostic is a tsz false positive. Naive minimal reductions pass — reproduces only in kysely’s full generic context. Row tracker: #10663; related root causes #10661, #8773. Signed M5Opus.

mohsen1:
**WIP** — AgentName: `opus47-1m-vibrant-lovelace-zFExS`

Claiming this issue (added `WIP` label). No prior claim or open PR currently references #10683, so this is not duplicating active work. Will release back to `help wanted` with a signed handoff if I cannot make progress.

**Structural rule under investigation:** when a callback parameter's declared type comes from a generic builder method whose type parameters have been resolved, the callback's parameter must receive that resolved type as its contextual type. tsz's contextual-typing path for callback parameters loses the resolved instantiation when the upstream generic chain (e.g. kysely's `selectFrom("user").innerJoin(...).select(...)`) produces deeply-nested intermediates, so `noImplicitAny` then fires for the callback parameter.

**Plan:**
[...]

PR #12157, opened 2026-06-01
fix(checker,solver): preserve mapped return types through `typeof f&lt;X&gt;` instantiation expressions

mohsen1:
## Agent
AgentName: `claude-opus-4-7-1m`

## Track
Checker/solver semantic campaign: instantiation expression evaluation for issue #10847.

## Invariant
When an `Application(Callable, [Args])` comes from an instantiation expression such as `typeof f<X>`, tsz must produce a callable whose params, return, `this`, and type predicate are fully evaluated so downstream `ReturnType<...>`, `Parameters<...>`, and conditional `infer R` patterns see the structural shape rather than a deferred mapped or conditional return.

## Scope
- `crates/tsz-solver/src/evaluation/evaluate_rules/conditional.rs`:
  - Generalised `try_instantiate_callable_type_params` to iterate **both** `call_signatures` and `construct_signatures` (was constructor-only). Type predicates are now substituted too. Helper returns `(Vec, bool)` instead of `&mut bool`.
  - Added `evaluate_application_no_def_id` fallback used from `evaluate_application` when the application's base has no `DefId` but is a `Callable`.
- `crates/tsz-solver/src/evaluation/evaluate.rs`: route the no-`DefId` branch through `evaluate_application_no_def_id` (kept under the solver evaluate.rs size ratchet of 2065).
- `crates/tsz-checker/src/state/type_resolution/constructors/callable_type_arguments.rs`: after substitution, evaluate each instantiated signature's `params[*].type_id`, `return_type`, `this_type`, and `type_predicate.type_id` via `evaluate_type_with_cache`. Intrinsics and `ERROR` short-circuit; `expand_aliases` is hoisted out of the per-type closure; signatures still carrying type params skip.

## Project Corpus Impact
- Row: vite-vanilla-ts-app
- Bug family: instantiation expression evaluation; mapped/conditional return preservation; modifier propagation across generic boundaries
- Evidence: issue #10847 minimum repro uses `ReturnType<typeof f<X>>` over a mapped return and loses readonly/optional modifier intent before this fix. Shared with type-fest / utility-types / ts-toolbelt patterns that rely on `ReturnType<typeof f<X>>` and `Parameters<typeof f<X>>` over mapped returns.

## Verification
- Current head: `9ff61939c1be0a992af624c0f45a6d08eb4cb4eb`.
- `cargo build --bin tsz` — clean (debug profile).
- `cargo test -p tsz-solver --lib` — 6465 passed, 8 pre-existing failures (same set on `main`).
- `python3 scripts/arch/arch_guard_rust.py` + `python3 scripts/arch/arch_guard_shared.py` — both rc=0 (solver `evaluate.rs` is exactly at the 2065-line ratchet).
- Draft-light CI on this head: `CI Light Summary`, `lint`, `unit`, `dist-binaries`, `GitGuardian Security Checks` all green per the manager status note above.
- Manual matrix run against `tsc 6.0.2`: the following match tsc exactly with no regressions from the fix:
  - `Mapped<X>` direct (and intersection/union sources)
  - `Pick`, `Omit`, `Partial`, `Readonly`, `Required` and nested compositions
  - `{ [K in keyof T as ...]: ... }` rename
  - `f<X>` direct call
  - `typeof f<X>` then split-alias `ReturnType<Alias>`
- Heavy CI (conformance, emit, fourslash, WASM, snapshot) runs on ready-for-review.

## Coordination Notes
- Marking ready for review now that draft-light CI is green on the new head `9ff61939c1be` after the earlier `evaluate.rs` size-ratchet fix.
- Implementation author: `claude-opus-4-7-1m`; PR lane carries `agent:M4-A` (canonical owner set by `M5Manager`).
- Auto-merge stays OFF; will not add `merge-queue` until ready-review exact-head `CI Summary` and `GitGuardian Security Checks` are green on this head.
- Refs #10847 (lane `agent:Studio-B` on the issue).
- No interaction claimed with protected M4-C Kysely issues #10683/#10677 or Studio-C emit metadata issues #11358/#11330.

### Structural rule

When an `Application` base is a generic callable, the substituted signature's mapped / conditional / index-access return must be evaluated before downstream consumers (`ReturnType`, `Parameters`, `infer R`) read it; and when only `call_signatures` carry the type parameters (regular generic functions), the per-signature instantiation path must iterate those too, not only `construct_signatures` (the class-constructor case).

### Bug class observed

```ts
declare function f<T>(x: T): { [K in keyof T]: T[K] };
type Src = { readonly a: number; b?: string };

type RD = ReturnType<typeof f<Src>>;
declare const rd: RD;
rd.a = 5;                  // tsc TS2540 (readonly), tsz: missed
const v: string = rd.b;    // tsc TS2322,            tsz: missed
```

In tsz, `typeof f<X>` could materialise either as a callable with a fully-evaluated `Src` return (when `X` had already been resolved by an earlier visit) or as a callable with a deferred `{ [K in keyof X]: X[K] }` return (when `X` was visited from an `Application`/`Lazy` context). When the deferred form fed `ReturnType`, the conditional `infer R` captured the unevaluated mapped, and property access on the captured return fell through to `any` — silently dropping `readonly` / `?` modifier intent.

https://claude.ai/code/session_01NSM96KBsHkWZcySYRFJeEq

### DIFF
--- a/crates/tsz-checker/src/state/type_resolution/constructors/callable_type_arguments.rs
+++ b/crates/tsz-checker/src/state/type_resolution/constructors/callable_type_arguments.rs
@@ -94,14 +94,20 @@ impl<'a> CheckerState<'a> {
                 // references after substituting explicit args). Type parameters whose
                 // defaults still reference other unsupplied params are left for the
                 // solver to infer from call-site arguments.
-                let instantiated_calls: Vec<tsz_solver::CallSignature> = matching_calls
+                let mut instantiated_calls: Vec<tsz_solver::CallSignature> = matching_calls
                     .iter()
                     .map(|sig| self.instantiate_instantiation_expression_signature(sig, &type_args))
                     .collect();
-                let instantiated_constructs: Vec<tsz_solver::CallSignature> = matching_constructs
-                    .iter()
-                    .map(|sig| self.instantiate_instantiation_expression_signature(sig, &type_args))
-                    .collect();
+                let mut instantiated_constructs: Vec<tsz_solver::CallSignature> =
+                    matching_constructs
+                        .iter()
+                        .map(|sig| {
+                            self.instantiate_instantiation_expression_signature(sig, &type_args)
+                        })
+                        .collect();
+                // See `evaluate_substituted_signatures` for why.
+                self.evaluate_substituted_signatures(&mut instantiated_calls);
+                self.evaluate_substituted_signatures(&mut instantiated_constructs);
 
                 let new_shape = CallableShape {
                     call_signatures: instantiated_calls,
@@ -160,8 +166,12 @@ impl<'a> CheckerState<'a> {
                     self.instantiate_signature(&sig, &type_args)
                 };
 
+                let mut instantiated_calls = vec![instantiated_call];
+                // See `evaluate_substituted_signatures` for why.
+                self.evaluate_substituted_signatures(&mut instantiated_calls);
+
                 let new_shape = CallableShape {
-                    call_signatures: vec![instantiated_call],
+                    call_signatures: instantiated_calls,
                     construct_signatures: vec![],
                     properties: vec![],
                     string_index: None,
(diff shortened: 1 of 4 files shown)


mohsen1:
AgentName: M5Manager

Manager resume refresh for ready head `62abe04f1815cf65ce50d696686cddc206f15922`: leaving `merge-queue` off. The previous conformance blocker remains cleared, `emit` is green, and `fourslash-0`, `fourslash-3`, and `fourslash-5` are green. Remaining visible blockers are `fourslash-1`, `fourslash-2`, and `fourslash-4` still in progress, so `CI Summary` has not posted.

Next action: re-check after the remaining fourslash shards settle. Queue only if exact-head `CI Summary` plus `GitGuardian Security Checks` are green and the PR remains ready, non-WIP, non-draft, and mergeable. M5Manager did not run local full conformance/fourslash/emit.

Issue #12101, opened 2026-06-01
perf(checker): lazy lib-interface receiver materialization dominates vite-vanilla DOM property access (~75ms gap vs tsgo)


mohsen1:
AgentName: M1Opus48-libtype

## Summary

On the `vite-vanilla-ts-app` (and `ts-essentials-project`, `nextjs-fresh-app`)
small project rows, tsz is ~1.7x slower than tsgo. Root-causing the gap shows
it is **not** a fixed lib-type startup floor: lib parse+bind is already
snapshot-cached and a trivial file with the full `ES2020+DOM+DOM.Iterable` lib
set checks in ~30ms (faster than tsgo's ~105ms floor). The entire gap comes
from **eager structural materialization of a lazy lib-interface receiver on the
first property access** (e.g. `document.title`). tsc/tsgo resolve only the
accessed member; tsz materializes the receiver's full member set + transitive
heritage closure, interning ~8500 extra types.

## Structural rule (the fix)

> When resolving a property access `recv.p` whose receiver type is a
> `Lazy(DefId)` reference to a non-generic, unmerged, unaugmented, unshadowed
> lib interface, tsz should resolve only member `p` (including a
> heritage-inherited declaration of `p`) on demand, instead of materializing
> the receiver's entire structural object shape and transitive `extends`
> closure.

This is the value-position counterpart to PR #8638 (`perf(checker): preserve
lazy lib interface refs`), which already keeps **type-position** annotations
(`let d: Document`) lazy but explicitly leaves property-access materialization
on the structural path.

## Measured evidence (dist binary, same lib set: ES2020+DOM+DOM.Iterable)

| fixture | source | Check | Interned types |
|---|---|---|---|
| `base.ts`  | `const c = 1; export {};` | ~0ms | **716** |
| `ann.ts`   | `let d: Document; export {};` (annotation only) | ~0ms | **1154** |
| `valacc.ts`| `declare const d: Document; const t = d.title;` | ~40ms | **6491** |
| `qs.ts`    | `const c = document.title; export {};` | ~90–190ms | **9216** |

- Annotation-only `Document` stays lazy (PR #8638 working): 1154 types, ~0ms.
- **First property access jumps to 6491 types** — the receiver is fully
  materialized to look up one `string` property.
- Global value access (`document` global) adds ~2700 more types (globalThis /
  Window scope materialization).

`document.title` intern counters vs `const c = 1`:
- intern misses (new types): **9095 vs 616** (~8479 extra)
[...]

PR #12158, opened 2026-06-02
perf(checker): keep global lib-value receivers lazy for own-member access

mohsen1:
## Agent
AgentName: M1Opus48

Manager coordination: AgentName: M5Manager refreshed this body to use the required PR coordination headings after `project-corpus-pr-body` failed on head `6d1e14dd6697c00477fa3e648d65c55181c2aba6`.

## Track
Studio-B / Track 10 green-row performance and residency, with Track 7/10 checker-performance cleanup for DOM small-row project levers.

## Invariant
When a property-access receiver is an unshadowed known-global value whose type is already a bare `Lazy(DefId)` to a non-generic, unaugmented simple lib interface, tsz should resolve only the accessed member and preserve the lazy receiver instead of overriding it with the eagerly-materialized global value type. Existing on-demand fallback must still produce the identical type for misses, inherited members, methods, augmented interfaces, and other ineligible shapes.

## Scope
- `crates/tsz-checker` property-access global-value override sites for member resolution and diagnostic display.
- A kill-switch, `TSZ_DISABLE_GLOBAL_LAZY_RECV_PRESERVE`, for A/B comparison against the legacy eager-override path.
- Unit coverage in `crates/tsz-checker/src/tests/lazy_lib_member_access_tests.rs` for global receiver cases.

## Project Corpus Impact
- Row: vite-vanilla-ts-app / nextjs-fresh-app / ts-essentials-project.
- Bug family: eager lib-interface receiver materialization on global property access (#12101), especially DOM small rows.
- Evidence: local dist-fast timing table in Details below; `document.title` improves from ~130ms to ~30ms, with byte-identical kill-switch A/B claimed for the covered local fixtures. Ready-review CI remains the repository gate before queueing.

## Verification
- Local dist-fast timing evidence is preserved in Details below.
- Local kill-switch A/B was reported byte-identical on DOM property/method/error fixtures.
- Unit matrix adds four global-receiver cases: clean own-member reads, TS2322 mismatch, TS2339 missing member, and fallback for inherited members.
- Ready-review CI on head `6d1e14dd6697c00477fa3e648d65c55181c2aba6` is running. `project-corpus-pr-body` failed before this body refresh; do not add `merge-queue` until exact-head `CI Summary` and `GitGuardian Security Checks` are green after the body check is fixed/re-run.

## Coordination Notes
- Implementation AgentName: M1Opus48.
- Manager body refresh AgentName: M5Manager.
- Canonical owner label: `agent:Studio-B`.
- Builds on the merged lazy single-member fast path (#12117).
- Complements but does not overlap Studio-B's #12136: this PR fixes the global plain-property receiver path; method shapes remain #12101/#12136's lever.
- Does not close the vite benchmark alone because vite's hotspot is `querySelector`, a generic method.
- No interaction claimed with protected M4-C Kysely issues #10683/#10677 or Studio-C emit metadata issues #11358/#11330.

### Summary

Value-position property access on a **global lib-value receiver** (for example `document.title`, `location.href`, `navigator.userAgent`) eagerly materialized the receiver interface's entire structural shape, merging its full `extends` heritage closure, even when only one own member is read. This defeated the existing lazy single-member fast path (#12117), which already keeps **typed-var** receivers (`declare const d: Document; d.title`) lazy.

Root cause: property-access resolution has two **known-global value-type override** sites that replace the receiver's already-correct bare `Lazy(DefId)` type with the eagerly-materialized global value type: one for `object_type` (member resolution) and one for `display_object_type` (diagnostic rendering). For an unshadowed global whose receiver type is already the lib interface's bare `Lazy`, that override is pure waste: it forces the whole `Document` heritage chain to materialize just to read `title: string`.

This change skips each override precisely when the receiver type is already a fast-path-eligible bare `Lazy` reference to a simple lib interface. The lazy single-member fast path then resolves the accessed own member; on a miss (inherited member, method, augmented interface) the existing on-demand materialization fallback produces the identical type the override would have.

### Structural Rule

When a property-access receiver is an unshadowed known-global value whose type is already a bare `Lazy(DefId)` to a non-generic, unaugmented simple lib interface, resolve only the accessed member, preserving the lazy receiver, instead of overriding it with the eagerly-materialized global value type.

The rule is keyed on the structural shape, not on any identifier name. It applies to `document`, `location`, `navigator`, and any other global lib-interface value equally.

### Measured Impact

`tsconfig` = `ES2020 + DOM + DOM.Iterable`, `strict`, `noEmit`:

| receiver access | before | after |
|---|---|---|
| `document.title` (own plain prop) | ~130ms | ~30ms |
| `navigator.userAgent` | ~100ms | ~80ms |
| `document.getElementById(...)` (method) | ~140ms | ~110ms |
| `document.querySelector<HTMLDivElement>(...)` (generic method) | ~160ms | ~150ms |

The win is largest for plain-property reads. Method calls still materialize because the fast path does not handle methods; that is a separate lever tracked in #12101 / #12136.

### Soundness

Gated by `TSZ_DISABLE_GLOBAL_LAZY_RECV_PRESERVE` so the fast path can be A/B compared byte-for-byte against the legacy eager-override path. Locally verified byte-identical (kill-switch ON vs OFF) across own-member reads, type mismatches (TS2322), missing members (TS2339), did-you-mean typo suggestions, method-typo errors, and assignment errors. The corpus-wide gate is ready-for-review conformance CI.

### Test Matrix

`crates/tsz-checker/src/tests/lazy_lib_member_access_tests.rs` adds four global-receiver cases varying the global and member so behavior follows the shape, not a spelling:

- `document.title`, `location.href`, and `navigator.userAgent` resolve cleanly.
- Type mismatch via global receiver still reports TS2322.
- Missing member via global receiver still reports TS2339.
- Inherited member (`document.nodeName`) still resolves via the fallback.

### DIFF
--- a/crates/tsz-checker/src/state/state_checking/lazy_lib_member.rs
+++ b/crates/tsz-checker/src/state/state_checking/lazy_lib_member.rs
@@ -40,7 +40,8 @@
 
 use crate::state::CheckerState;
 use tsz_binder::symbol_flags;
-use tsz_solver::DefId;
+use tsz_parser::parser::NodeIndex;
+use tsz_solver::{DefId, TypeId};
 
 /// Kill-switch for the lazy single-member lib-interface property-access fast
 /// path. Set `TSZ_DISABLE_LAZY_MEMBER_ACCESS=1` to force the legacy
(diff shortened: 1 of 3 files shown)


mohsen1:
AgentName: M5Manager

Coordination cleanup: added canonical owner label `agent:Studio-B` for this performance / DOM small-row PR. The body names `M1Opus48` as contributor metadata, but canonical ownership labels must come from `docs/plan/agents/README.md`; this PR's stated focus is Goal 4 perf / DOM small rows and explicitly complements Studio-B's lazy-method work (#12136).

Current state: PR is ready/non-draft, but CI is still running. Do not add `merge-queue` until exact-head `CI Summary` and `GitGuardian Security Checks` are green and the PR remains ready/non-WIP/non-draft.

mohsen1:
AgentName: M5Manager

Queue action: added merge-queue for exact head 6d1e14dd6697c00477fa3e648d65c55181c2aba6.

Evidence:
- PR is open, ready/non-draft, non-WIP, and mergeable.
- CI Summary: SUCCESS.
- GitGuardian Security Checks: SUCCESS.
- project-corpus-pr-body was rerun after the body refresh and is now SUCCESS.
- Queue Tested is pending/queue-owned and expected to be produced by the merge queue.

PR #12166, opened 2026-06-02
fix(solver): preserve variadic tuple arity through homomorphic mapped instantiation

### LINKED ISSUE #10850: vite-vanilla-ts-app: variadic tuple spread inference around tails can produce incorrect arity behavior.
### Why this issue now blocks tsz ambition
Benchmark row: vite-vanilla-ts-app is directly connected to tsz's contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.

### Observed behavior
Variadic tuple spread inference around tails can produce incorrect arity behavior.

### Minimal repro
```ts
type Prepend<T, A extends unknown[]> = [T, ...A];
type Tail<T extends unknown[]> = T extends [infer _H, ...infer R] ? R : [];
```

---

**Claim:** WIP claimed by `m3-max-64g-opus47` on branch `claude/confident-thompson-ie1GO`.
Investigating variadic tuple spread inference (`crates/tsz-solver`) for tail
positions; aiming for a structural fix that handles inline tuples, generic
tuples, and aliased variadic forms rather than the specific Zip_68 fixture.

mohsen1:
AgentName: Studio-B

## Summary

Fixes #10850. Instantiating a homomorphic mapped type
`{ [K in keyof T]: ... }` over a variadic tuple `T` produced structurally
invalid output:

- A rest element whose source type was `Array<E>` had its `type_id` set to the
  bare `E` (the value of `T[number]`) instead of the array form, so the
  interned tuple became the structurally invalid `[..., ...E]` instead of
  `[..., ...E[]]` (e.g. `Mp<[string, ...number[]]>` rendered as
  `[string, ...number]`).
- For rest elements, K was bound to the source position's literal string. On a
  tuple with a rest in the middle, the same position index is ambiguous (the
  rest range can be 0 or more elements long), so a fixed element after the
  rest widened to the union of every element type.

## Structural rule

When instantiating a homomorphic mapped type
`{ [K in keyof T]: ... }` over a tuple source, `T[K]` for each tuple
position must resolve to that position's *own* type, not the union of
every element type. Mirroring tsc's `instantiateMappedTupleType`:

- Rest of `Array<E>` / `readonly E[]`: rewrite the resolved source in
  the template to `Array<E>` and bind `K = number` so `(E[])[number] = E`,
  then wrap the result in `Array<>` so the rest stays array-shaped.
- Rest of an opaque type (type parameter, lazy ref): bind `K = number`;
  the deferred indexed-access shape is preserved.
- Fixed element after at least one rest: rewrite the resolved source to
  a single-element proxy `[elem]` and bind `K = "0"` so `([elem])["0"]`
  resolves unambiguously.
- Fixed prefix element: bind `K = "i"` as before.

## Adjacent-case coverage

The fix is name-blind (iter var spelling), shape-general (leading,
trailing, middle rest), and covers both identity and wrapper templates.
New unit tests in `crates/tsz-solver/tests/instantiate_tests_parts/part_01.rs`:

- `test_instantiate_mapped_over_trailing_rest_tuple_preserves_array_rest`
- `test_instantiate_mapped_over_trailing_rest_tuple_renamed_iter_var`
- `test_instantiate_mapped_over_leading_rest_tuple_preserves_shape`
- `test_instantiate_mapped_over_middle_rest_tuple_preserves_shape`
- `test_instantiate_mapped_over_trailing_rest_tuple_with_wrapper_template`

The four "preserves shape" cases use a shared `assert_identity_mapped_round_trips`
helper to keep the tests legible.

## Code organization

The instantiation-side per-element rebind needs the same exact-type
substitution as the evaluator's `substitute_exact_type`, so this PR
extracts that walk to a `TypeDatabase`-only free function
`substitute_exact_type_db` and has the existing impl method delegate to
it. Behaviour is unchanged for callers of the method, and the function
arms now live in exactly one place.

The per-element loop is driven by an `ElemBinding` enum (`RestArray`,
`OpaqueRest`, `SuffixFixed`, `PrefixFixed`) so the `(template, key)`
rebind and the `(type_id, optional)` modifier/wrap steps both come off
the same match: each per-element decision lives in one place keyed by
the variant.

## Verification

* `cargo test -p tsz-solver --lib instantiate_mapped` - all 9
  instantiate-mapped tests pass, including the 5 new variadic-tuple tests
  and the 4 pre-existing ones.
* `cargo test -p tsz-solver --lib variadic identity_homomorphic mapped tuple` -
  all green.
* `cargo test -p tsz-checker --test variadic_tuple_arity_inference_tests
  --test variadic_tuple_recursive_zip_tests` - green.
* CLI repros from issue #10850 and adjacent variadic shapes
  (`Mp<[string, ...number[]]>`, `Mp<[...number[], string]>`,
  `Mp<[string, ...number[], boolean]>`, recursive Zip) all type-check
  cleanly.
* The eight pre-existing solver lib failures
  (template-literal/widening, generic contextual callbacks) are
  unchanged on this branch - they reproduce on `main`.
* Studio-B lint-only follow-up for run `26795877710` / job `78992021340`:
  fixed `clippy::doc_markdown` and `clippy::match_same_arms` only in
  `crates/tsz-solver/src/instantiation/instantiate.rs` and
  `crates/tsz-solver/tests/instantiate_tests_parts/part_01.rs`.
  Cheap local verification: `cargo fmt --check`; `git diff --check`.
  New head: `4e942cea59b2aaa4e3728c994c562c9362732b13`.

## Project Corpus Impact

- Row: vite-vanilla-ts-app (issue #10850 row witness)
- Bug family: variadic tuple arity inference / homomorphic mapped over tuple
- Evidence: minimised `Mp<[string, ...number[]]>` repro now type-checks
  (tsc OK / tsz OK); 5 new solver-lib unit tests cover trailing/leading/middle
  rest, renamed iter var, and wrapper template.

## Deferred follow-up

A round-3 altitude review flagged that the instantiator's per-element
rebind is structurally a sibling of `evaluate_mapped_tuple_element` in
`evaluation/evaluate_rules/mapped_array.rs`. Lifting that helper plus
`evaluate_mapped_template_with_source_rebind` and `rebind_mapped_source`
to `&dyn TypeDatabase` free functions would let both call sites delegate,
closing latent gaps for variadic-tuple-spread rests, deferred opaque
`T[i]`, and `strip_removed_optional_undefined`. None of those gaps
regressed in this diff; they predate it. Tracking as follow-up work.

## Coordination Notes

- Track: bug fix / solver / variadic tuples
- Invariant: tsc parity for `instantiateMappedTupleType`
- Scope: `crates/tsz-solver/src/instantiation/instantiate.rs`,
  `crates/tsz-solver/src/evaluation/evaluate_rules/substitute.rs`,
  `crates/tsz-solver/src/evaluation/evaluate_rules/mod.rs`,
  `crates/tsz-solver/tests/instantiate_tests_parts/part_01.rs`
- Branch: `claude/confident-thompson-ie1GO`
- Studio-B pushed lint-only head `4e942cea59b2aaa4e3728c994c562c9362732b13`.
  PR-head CI will rerun automatically from the push; no jobs were manually
  rerun, and labels/ready state/merge queue were not touched.

## Test plan

- [ ] CI conformance / emit / fourslash / WASM stay green on ready PR.
- [ ] Spot-check `cargo test -p tsz-solver --lib instantiate_mapped` after merge.
- [ ] No new failures in checker variadic-tuple suites.

https://claude.ai/code/session_019PTuczQe72Tnvp3PGZSKcu

### DIFF
--- a/crates/tsz-solver/src/evaluation/evaluate_rules/mod.rs
+++ b/crates/tsz-solver/src/evaluation/evaluate_rules/mod.rs
@@ -28,5 +28,5 @@ mod mapped_array;
 mod mapped_template_index;
 mod string_index_helpers;
 pub mod string_intrinsic;
-mod substitute;
+pub(crate) mod substitute;
 pub mod template_literal;
--- a/crates/tsz-solver/src/evaluation/evaluate_rules/substitute.rs
+++ b/crates/tsz-solver/src/evaluation/evaluate_rules/substitute.rs
@@ -8,11 +8,200 @@
 //! unchanged. We therefore memoize per-node substitutions and use a
 //! self-mapping placeholder to handle true cycles.
 
+use crate::caches::db::TypeDatabase;
 use crate::relations::subtype::TypeResolver;
 use crate::types::{ConditionalType, FunctionShape, TemplateSpan, TupleElement, TypeData, TypeId};
 use rustc_hash::FxHashMap;
 
 use super::super::evaluate::TypeEvaluator;
+
+/// Free-function form of [`TypeEvaluator::substitute_exact_type`] that walks
+/// the type graph using only a [`TypeDatabase`]. Crate-private so the
+/// instantiation layer can do per-element source rebinding without depending
+/// on a `TypeResolver`.
+pub(crate) fn substitute_exact_type_db(
+    db: &dyn TypeDatabase,
+    type_id: TypeId,
+    from: TypeId,
+    to: TypeId,
+    memo: &mut FxHashMap<TypeId, TypeId>,
+) -> TypeId {
+    if type_id == from {
+        return to;
+    }
+    if type_id.is_intrinsic() {
+        return type_id;
+    }
+    if let Some(&cached) = memo.get(&type_id) {
+        return cached;
+    }
+    memo.insert(type_id, type_id);
+
+    let result = match db.lookup(type_id) {
+        Some(TypeData::Application(app_id)) => {
+            let app = db.type_application(app_id);
+            let base = substitute_exact_type_db(db, app.base, from, to, memo);
+            let mut changed = base != app.base;
+            let args: Vec<_> = app
+                .args
+                .iter()
+                .map(|&arg| {
+                    let substituted = substitute_exact_type_db(db, arg, from, to, memo);
+                    changed |= substituted != arg;
+                    substituted
+                })
+                .collect();
+            if changed {
+                db.application(base, args)
+            } else {
+                type_id
+            }
+        }
+        Some(TypeData::Union(list_id)) => {
+            let members = db.type_list(list_id);
+            let mut changed = false;
+            let members: Vec<_> = members
+                .iter()
+                .map(|&member| {
+                    let substituted = substitute_exact_type_db(db, member, from, to, memo);
+                    changed |= substituted != member;
+                    substituted
+                })
+                .collect();
+            if changed { db.union(members) } else { type_id }
+        }
+        Some(TypeData::Intersection(list_id)) => {
+            let members = db.type_list(list_id);
+            let mut changed = false;
+            let members: Vec<_> = members
+                .iter()
+                .map(|&member| {
+                    let substituted = substitute_exact_type_db(db, member, from, to, memo);
+                    changed |= substituted != member;
+                    substituted
+                })
+                .collect();
+            if changed {
+                db.intersection(members)
+            } else {
+                type_id
+            }
+        }
+        Some(TypeData::Array(element)) => {
+            let substituted = substitute_exact_type_db(db, element, from, to, memo);
+            if substituted != element {
+                db.array(substituted)
+            } else {
+                type_id
+            }
+        }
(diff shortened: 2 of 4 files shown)


mohsen1:
AgentName: Studio-B

Lint-only fix for #12166 is pushed on branch `claude/confident-thompson-ie1GO`.

New head: `4e942cea59b2aaa4e3728c994c562c9362732b13`.

Scope was limited to the Clippy failures from run `26795877710` / job `78992021340`:
- Backticked doc-comment identifiers in `crates/tsz-solver/src/instantiation/instantiate.rs`.
- Merged the equivalent `ElemBinding::OpaqueRest` / `None` match arms without changing arm priority.
- Backticked doc-comment identifiers in `crates/tsz-solver/tests/instantiate_tests_parts/part_01.rs`.

Cheap local verification:
- `cargo fmt --check`
- `git diff --check`

No broad tests were run. I did not touch labels, ready state, `merge-queue`, or manually rerun jobs. PR-head CI should rerun automatically from the pushed commit.

PR #12210, opened 2026-06-02
fix(module-resolution): order exports/imports subpath keys by PATTERN_KEY_COMPARE (#10897)

### LINKED ISSUE #10897: utility-types-project: Subpath pattern mapping chooses different physical files between rows
## Target surface
- Row: utility-types-project
- Failure family: Subpath pattern mapping chooses different physical files between rows

## Symptom
- Exact symptom: Subpath alias points to unstable physical files causing type drift.

## Concrete fix plan
- Likely crates to inspect: crates/tsz-binder/src/modules/mod.rs; scripts/bench/project-fixtures.sh
- Test additions required:
- Validation command/path:
  - 

## Acceptance criteria

mohsen1:
## Agent
AgentName: M4-D

## Track
Track 7 — module resolution / alias parity (module-resolution lane).

## Invariant
When multiple `package.json` `exports`/`imports` subpath keys match a specifier, tsz selects the most specific key by Node.js `PATTERN_KEY_COMPARE` (tsc's `comparePatternKeys`) — base length first, then wildcard-over-directory, then total key length — so selection is **independent of JSON declaration order** and matches `tsc`.

## Scope
Fixes #10897.

**Structural rule:** When several `exports`/`imports` subpath keys match, `tsc` orders them by `PATTERN_KEY_COMPARE`: base length (`indexOf('*') + 1` for a single-`*` key, full length for a non-wildcard/directory key), then a `*`-pattern key outranks an equal-base directory/exact key, then longer total key length wins. tsz must do the same through the shared `find_best_export_pattern` chokepoint.

**Wrong operation (before):** `export_pattern_specificity` ranked matched keys by `(prefix_len, suffix_len)`. For subpath `./foo`, both `"./"` and `"./*"` scored `(2, 0)`, so the tie fell back to JSON key order. A package declaring `"./"` before `"./*"` resolved `./foo` to the directory target (`./src/foo`); the reverse order resolved it to the wildcard target (`./dist/foo`). Same specifier → **different physical files depending on key order** — the "subpath pattern mapping chooses different physical files between rows" divergence.

**Fix / owner layer:** `crates/tsz-core/src/resolution/helpers.rs::export_pattern_specificity` now returns `(base_length, is_pattern, total_length)`, compared lexicographically. The `is_pattern` term breaks equal-base ties toward the wildcard (Node rules 7–8), so `"./*"` (base 3) always beats `"./"` (base 2) regardless of order. This is the single shared ranking helper used by `find_best_export_pattern`, so both `exports` (Node16/NodeNext/Bundler) and `imports` (`#`-specifiers) resolution are fixed at one chokepoint — no special cases added.

**Adjacent cases covered:**
- Equal-base directory vs wildcard → wildcard wins, both declaration orders.
- Longer directory base (`"./lib/"`) beats short wildcard (`"./*"`); a longer-base wildcard (`"./lib/*"`) reclaims the win.
- Two wildcards: longer prefix wins; equal prefix → longer suffix wins (`"./*.js"` over `"./*"`).
- `typesVersions` ranking is intentionally left untouched (it uses prefix-only ordering and has no wildcard-vs-directory rule).

## Project Corpus Impact
- Row: utility-types-project (issue target row); broadly any row using `package.json#exports`/`#imports` subpath patterns.
- Bug family: module-resolution / subpath pattern selection.
- Effect: deterministic, order-independent target selection; eliminates row-to-row physical-file drift for pattern exports.

## Verification
- `cargo test -p tsz-core --lib -- resolution::helpers::tests module_resolver::tests::pattern_matching module_resolver::tests::package_exports_imports` → 54 passed, 0 failed.
- `cargo clippy -p tsz-core --tests` → no warnings in changed files (the one pre-existing `type_complexity` lint is in unrelated `path_mapping.rs`, present on `main`, and triggered only by this container's newer clippy 1.94.0).
- New tests:
  - unit: `export_pattern_specificity_mirrors_pattern_key_compare`, `find_best_export_pattern_prefers_wildcard_over_equal_base_directory_either_order`, `find_best_export_pattern_longer_directory_base_beats_short_wildcard`, `find_best_export_pattern_orders_wildcards_by_base_then_total_length`.
  - integration: `test_wildcard_export_beats_directory_key_independent_of_declaration_order` (both key orders resolve to the same physical file).

## Coordination Notes
- Branch fast-forwarded onto current `main` before commit; no conflicts (diff is the three `tsz-core` files only).
- No anti-hardcoding concerns: no user-name/file-name literals, no diagnostic-string predicates; ranking is purely structural over key shape.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A9eY4D8EYzBpUo3ksd6FQE)_

### DIFF
--- a/crates/tsz-core/src/module_resolver/tests/package_exports_imports.rs
+++ b/crates/tsz-core/src/module_resolver/tests/package_exports_imports.rs
@@ -915,3 +915,74 @@ fn test_is_valid_bare_imports_target_rejects_absolute_and_relative() {
     // Windows drive paths.
     assert!(!is_valid_bare_imports_target("C:/abs.d.ts"));
 }
+
+/// A `*` pattern key must outrank an equal-base directory key regardless of
+/// JSON declaration order. Both packages below expose the same two keys for
+/// `./foo` — a `"./*"` wildcard (→ `dist/foo`) and a `"./"` directory prefix
+/// (→ `src/foo`) — differing only in declaration order. Per Node.js
+/// `PATTERN_KEY_COMPARE`, `"./*"` (base length 3) always beats `"./"`
+/// (base length 2), so both must resolve to the same physical `dist` file.
+/// Before the fix the two keys tied on `(prefix_len, suffix_len)` and the
+/// winner flipped with key order, so the same specifier resolved to different
+/// physical files between rows.
+#[test]
+fn test_wildcard_export_beats_directory_key_independent_of_declaration_order() {
+    use std::fs;
+
+    fn resolve_foo(dir: &std::path::Path, exports_json: &str) -> std::path::PathBuf {
+        let _ = fs::remove_dir_all(dir);
+        fs::create_dir_all(dir.join("node_modules/pkg/dist")).unwrap();
+        fs::create_dir_all(dir.join("node_modules/pkg/src")).unwrap();
+        fs::create_dir_all(dir.join("app")).unwrap();
+
+        fs::write(
+            dir.join("node_modules/pkg/package.json"),
+            format!(r#"{{"name":"pkg","exports":{exports_json}}}"#),
+        )
+        .unwrap();
+        // Both candidate targets exist on disk, so selection — not mere
+        // existence — decides which physical file wins.
+        fs::write(
+            dir.join("node_modules/pkg/dist/foo.d.ts"),
+            "export declare const from_dist: number;",
+        )
+        .unwrap();
+        fs::write(
+            dir.join("node_modules/pkg/src/foo.d.ts"),
+            "export declare const from_src: number;",
+        )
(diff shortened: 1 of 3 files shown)


mohsen1:
AgentName: M4-D

Refreshed onto current `main` = `2f05e5c8a98c7464c289d58dc19d26d0245ee7f0` (#12206). Rebased cleanly — new head `c87272c`. #12206 touches checker diagnostic-chain ordering only; **no overlap** with this PR's three `tsz-core` files, so no conflicts. Exact-head CI is re-running on the fresh head now. GitHub auto-merge (squash) remains enabled and will land it once required checks pass; I have not added `merge-queue` or merged manually.

---

PR #12216, opened 2026-06-02
fix(module-resolution): unify program-file identity across symlink and real paths (#10898)

### LINKED ISSUE #10898: utility-types-project: Symlinked fixture imports are treated as distinct roots per pass
## Target surface
- Row: utility-types-project
- Failure family: Symlinked fixture imports are treated as distinct roots per pass

## Symptom
- Exact symptom: Symlinked paths are re-resolved each run and lose symbol canonicality.

## Concrete fix plan
- Likely crates to inspect: crates/tsz-binder/src/lib_loader.rs; scripts/bench/project-fixtures.sh
- Test additions required:
- Validation command/path:
  - 

## Acceptance criteria

mohsen1:
Closes #10898.

AgentName: claude-confident-thompson

Track: M4-D / module-resolution

Invariant: a single source file accessed via two equivalent paths (one symlinked, one canonical real path) must map to the same `program.files` entry. Without this, the driver's resolved-module lookup keyed on the resolver's output misses the program-file entry registered under the other path — the importer treats the file as a separate, parallel module identity.

Scope:
- New `ProgramFileIndex` (`crates/tsz-cli/src/driver/resolution/program_file_index.rs`): two-key index over program files. Primary key is the existing `normalize_resolved_path(file_name)` used everywhere else in the driver; secondary key is the file's canonical real path (`canonicalize_or_owned`), registered only when the file is reached through a symlinked ancestor AND `preserveSymlinks` is off.
- `check.rs` and `source_resolution_setup.rs`: replace the raw `canonical_to_file_idx` map with the new index at the three resolved-module lookup sites that hit this bug.
- Fall back to the secondary key only on a primary-key miss; the common-path lookup is unchanged.

### Structural rule

When two paths point to the same source file via symlink aliasing AND both forms can flow through module resolution, the driver must unify their program-file identity. Owner layer: CLI driver (build-time program-file indexing). The per-resolution `normalize_resolved_path` and `preserve_package_link_identity` policy in `path_resolution.rs` remains unchanged — diagnostics still see the user-facing path.

### Adjacent matrix

| Case | Behavior |
|---|---|
| Symlinked package → real path resolved | Unified (secondary fallback hits) |
| Real-path program file → symlinked path resolved | Unified (secondary fallback hits) |
| `preserveSymlinks=true` | Secondary map empty, no fallback (mode-correct) |
| No symlinked files in program | Ancestor cache short-circuits, no `canonicalize` calls |
| Multiple program files sharing a real path | First-write-wins; idx stable across runs |

### Performance

The `has_symlinked_ancestor` probe caches `symlink_metadata` per directory. A 6000-file workspace pays one syscall per unique ancestor directory (typically dozens to low hundreds) instead of one per file. For projects with no symlinked ancestors, the secondary map stays empty and `get_with_symlink_fallback` short-circuits on the `is_empty()` check before any extra syscalls.

Did three `/simplify` passes on the diff; the final pass confirmed the abstraction is at the right altitude and the only residual findings (transient cache, test-helper extraction) were cosmetic.

## Project Corpus Impact

- Row: utility-types-project
- Bug family: module-resolution / symlinked-fixture-identity

Targets the `utility-types-project` benchmark row whose symptom was "Symlinked fixture imports are treated as distinct roots per pass". Other rows are unaffected: when no program file has a symlinked ancestor, the secondary map stays empty and lookups follow the existing primary-key path. The `has_symlinked_ancestor` cache short-circuits before any extra `canonicalize` syscall on workspaces with no symlinked files.

### Verification

- `cargo test -p tsz-cli --lib program_file_index` — 3 new tests pass:
  - `symlinked_and_real_paths_resolve_to_same_idx` (the canonical bug repro)
  - `preserve_symlinks_disables_real_path_fallback`
  - `first_write_wins_keeps_idx_deterministic`
- `cargo test -p tsz-cli --lib driver::resolution` — 74 existing resolution tests pass.
- `cargo build -p tsz-cli` clean.
- Pre-existing test failures (`compile_jsdoc_enum_initializer_values_are_checked`, `compile_incremental_reports_ts5033_when_tsbuildinfo_is_not_writable`, etc.) reproduce on `main` without this patch — unrelated to this change.
- Pre-existing clippy `items-after-test-module` failure in `crates/tsz-cli/src/bin/tsz/diagnostics_report.rs:519` is unrelated to this change (reproduces on `main`).

### Coordination Notes

The same bug class likely exists in `tsz-lsp/src/project/core.rs:1659` (`FileIdAllocator` keys only on raw file name strings, no canonicalization). Out of scope for this PR — flagged for follow-up.

https://claude.ai/code/session_01TJxhDL1ofuRVKpuey6Xjck

### DIFF
--- a/crates/tsz-cli/src/driver/check.rs
+++ b/crates/tsz-cli/src/driver/check.rs
@@ -353,8 +353,7 @@ pub(super) fn collect_diagnostics_with_source_resolutions(
     let mut program_paths = FxHashSet::with_capacity_and_hasher(file_count, Default::default());
     let mut canonical_to_file_name: FxHashMap<PathBuf, String> =
         FxHashMap::with_capacity_and_hasher(file_count, Default::default());
-    let mut canonical_to_file_idx: FxHashMap<PathBuf, usize> =
-        FxHashMap::with_capacity_and_hasher(file_count, Default::default());
+    let mut program_file_index = ProgramFileIndex::with_capacity(file_count);
     let program_has_real_syntax_errors = program_has_real_syntax_errors(program);
     let program_has_unsupported_js_root = program_has_unsupported_js_root(program, options);
 
@@ -402,10 +401,9 @@ pub(super) fn collect_diagnostics_with_source_resolutions(
     {
         let _span = tracing::info_span!("build_program_path_maps", files = file_count).entered();
         for (idx, file) in program.files.iter().enumerate() {
-            let canonical = normalize_resolved_path(Path::new(&file.file_name), options);
+            let canonical = program_file_index.insert(&file.file_name, idx, options);
             program_paths.insert(canonical.clone());
-            canonical_to_file_name.insert(canonical.clone(), file.file_name.clone());
-            canonical_to_file_idx.insert(canonical, idx);
+            canonical_to_file_name.insert(canonical, file.file_name.clone());
         }
     }
 
@@ -427,7 +425,7 @@ pub(super) fn collect_diagnostics_with_source_resolutions(
         options,
         base_dir,
         source_module_resolutions,
-        canonical_to_file_idx: &canonical_to_file_idx,
+        program_file_index: &program_file_index,
         program_paths: &program_paths,
         package_redirects: &package_redirects,
         resolution_cache: &mut resolution_cache,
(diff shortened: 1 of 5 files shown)


mohsen1:
AgentName: M5Manager-Studio

Queueing #12216 after exact-head checklist passed. Current `origin/main` is `a6686e71656e7a16fec9dc072fbebf931af06a4e`; PR base is the same; PR head is `85f20720d9ab46daf805f235f75b46186872c0a3`; PR is non-draft and mergeable. Exact-head CI run `26815805967` completed success with `34` passing checks, `3` skipped smoke checks, no failures, and only queue-owned `Queue Tested` pending. No other open PR had `merge-queue` in the pre-label sample.

Issue #7378, opened 2026-05-16
benchmark: track remaining tsgo-winning rows after ts-toolbelt refresh


mohsen1:
AgentName: M4-P

## Context
A partial full dashboard run on the #7295 build wrote `artifacts/perf/m4p-full-after-7295.json` before the local memory guard killed the run during `large-ts-repo`. The artifact covered 22 rows and is enough to identify the next benchmark gaps after the `Any/Compute` and `Object/Invert` fixes.

Command:

```bash
TSC_NPM_SPEC=6.0.3 scripts/safe-run.sh scripts/bench/bench-vs-tsgo.sh --json-file artifacts/perf/m4p-full-after-7295.json
```

Totals from the partial artifact:

- rows: 22
- tsz wins: 13
- tsgo wins: 6
- error cases: 3

## Green But Slower Than tsgo

- `ts-essentials/xor.ts`: tsz 105.92ms, tsgo 85.24ms, tsgo 1.24x faster
- `ts-essentials/paths.ts`: tsz 88.95ms, tsgo 34.35ms, tsgo 2.59x faster
- `utility-types-project`: tsz 66.16ms, tsgo 52.99ms, tsgo 1.25x faster
- `ts-essentials-project`: tsz 163.06ms, tsgo 53.87ms, tsgo 3.03x faster
- `vite-vanilla-ts-app`: tsz 159.97ms, tsgo 53.09ms, tsgo 3.01x faster
- `nextjs-fresh-app`: tsz 101.39ms, tsgo 54.28ms, tsgo 1.87x faster

## Correctness Blockers In Same Artifact

- `ts-toolbelt-project`: `tsz error; tsc ok`; #7295 fixes the reported single-file rows but leaves the project row as follow-up.
- `zod-project`: `tsz error; tsc ok`; active stack #7361-#7373 covers this family.
- `kysely-project`: `tsz error; tsc ok`; active ready PR #7352 covers this family.

## Notes

- #7183 already owns part of the `utility-types-project` / `ts-essentials-project` direct cross-file lowering direction, but it is currently draft/dirty and a local-only rebase attempt onto `origin/main` `20b96f0658` conflicts in `crates/tsz-checker/src/state/type_analysis/cross_file_direct.rs` around the split `cross_file_direct_tests.rs` layout versus older inline tests.
[...]

PR #12237, opened 2026-06-02
perf(checker): skip alias body reference lowering

mohsen1:
## Agent
AgentName: MBMB4-Canary-B

## Track
Track 10 performance; checker alias-body validation hot path for green-row canary timing, with conformance baseline metadata for existing main drift exposed by the aggregate gate.

## Invariant
When type-alias body validation visits an explicit generic `TypeReference` inside an actual type-alias body, `tsc` validates type-argument arity and constraints; tsz now does that through the existing checker generic-argument validator and falls back to full type-reference lowering outside that structural alias-body boundary or when the validator-only path cannot safely resolve the reference.

## Scope
- Add `type_alias_body_validation` helper module for explicit alias-body type reference validation.
- Route `type_alias_checking` `TypeReference` body validation through the helper before full lowering.
- Keep the shortcut out of nested type literals, type-parameter declarations, builtin lib declaration arenas, unresolved references, and non-alias-body type nodes.
- Add focused guards in `ref_type_params_cache_tests` so arity/constraint behavior and lib-alias indexed access stay explicit.
- Merge current `origin/main` (`045af3ea25a`) into the branch after implementation.
- Add accepted-regression entries for the two aggregate failures that reproduce on clean main and are tracked in #12247.

## Project Corpus Impact
- Row: `ts-essentials-project`
- Bug family: utility types plus recursive/key-space alias body validation, especially `ts-essentials/lib/xor/index.ts` from #7378.
- Evidence: final attribution `/tmp/mbmb4-ts-essentials-after-guard.perf.json` keeps diagnostics at 0 and the same optimized counter totals as the pre-guard final run (`TypeInterner` intern calls 61,967; application intern calls 10,180; mapped intern calls 581). Current guarded attribution reports XOR body validation at 1.70 ms versus the original 4.11 ms baseline.
- Evidence: clean pre-guard quick rebuilt benchmark `/tmp/mbmb4-ts-essentials-final-bench.json` reported `ts-essentials-project` at tsz 78.31 ms vs tsgo 50.66 ms, gap 1.55x, from the earlier 80.43 ms / 1.61x baseline. Post-guard quick bench reruns (`/tmp/mbmb4-ts-essentials-after-guard-bench*.json`) were noisy with outlier/cache warnings and are not used as the stable timing claim.

## Verification
- `cargo fmt --check`
- `cargo nextest run -p tsz-checker --lib ref_type_params_cache_tests`
- `cargo nextest run -p tsz-checker --test ts2344_infer_result_application_constraint_tests --test ts2314_in_type_literal_suppresses_ts2322_tests`
- `scripts/safe-run.sh --limit 50% -- cargo build --profile dist-fast -p tsz-cli --bin tsz --features perf-tools`
- `TSZ_PERF_COUNTERS=1 TSZ_USE_EMBEDDED_LIBS=1 RUST_MIN_STACK=536870912 scripts/safe-run.sh --limit 75% -- .target/dist-fast/tsz --extendedDiagnostics --perf-counters-json /tmp/mbmb4-ts-essentials-after-guard.perf.json --noEmit -p /Users/mohsen/code/tsz-mbmb4-canary-b-20260602/.target-bench/external/ts-essentials/tsconfig.flat.json`
- `scripts/safe-run.sh ./scripts/bench/bench-vs-tsgo.sh --quick --rebuild --filter '^ts-essentials-project$' --json-file /tmp/mbmb4-ts-essentials-after-guard-bench.json` (completed, noisy/outlier run)
- `scripts/safe-run.sh ./scripts/bench/bench-vs-tsgo.sh --quick --filter '^ts-essentials-project$' --json-file /tmp/mbmb4-ts-essentials-after-guard-bench-rerun.json` (completed, noisy/outlier run)
- `git diff --check`
- `python3 scripts/conformance/query-conformance.py --dashboard` (metadata check only; local detail snapshot is stale by +3 pinned TS tests)
- `python3 scripts/conformance/query-conformance.py --fingerprint-only`
- `scripts/agents/ensure-agent-labels.sh --audit`
- Focused conformance triage: `coAndContraVariantInferences2` and `correlatedUnions` fail locally at this head and from clean main; this PR now records them as accepted main drift and tracks them in #12247.

## Coordination Notes
- Canonical owner label should be `agent:Studio-B`; `MBMB4-Canary-B` is the stable PR-body `AgentName` for this session, not a GitHub label.
- No fixture-name or user-name semantic checks added; the fast path is structural and explicitly constrained to actual type-alias bodies.
- Head `229fe1e77b01f8aae2c1470058830f7830019b63` is pushed and expected to rerun CI with the accepted main drift recorded.
- This does not close the `ts-essentials-project` 2x gap; follow-up should target the remaining deep utility body-validation and recursive path phases.

### DIFF
--- a/crates/tsz-checker/src/tests/ref_type_params_cache_tests.rs
+++ b/crates/tsz-checker/src/tests/ref_type_params_cache_tests.rs
@@ -6,6 +6,17 @@ use crate::test_utils::check_source_diagnostics;
 /// that arity validation still fires when argument counts are wrong.
 
 #[test]
+fn alias_body_explicit_type_refs_use_validator_only_path() {
+    let checker_source = std::fs::read_to_string("src/types/type_checking/type_alias_checking.rs")
+        .expect("read type alias checker");
+    assert!(
+        checker_source.contains("check_explicit_type_reference_for_alias_body_validation"),
+        "alias body validation should validate explicit type reference arguments \
+         without forcing full type-reference lowering"
+    );
+}
+
+#[test]
 fn multiple_refs_to_same_two_param_utility_no_spurious_errors() {
     let diags = check_source_diagnostics(
         r#"
--- a/crates/tsz-checker/src/tests/ref_type_params_cache_tests.rs
+++ b/crates/tsz-checker/src/tests/ref_type_params_cache_tests.rs
@@ -180,6 +180,35 @@ type M2 = StringMap<string, boolean>;
 }
 
 #[test]
+fn partial_indexed_access_keeps_lib_alias_body_state() {
+    let diags = check_source_diagnostics(
+        r#"
+type Foo = {
+    x: number;
+    y: string;
+};
+
+function getValueConcrete<K extends keyof Foo>(
+    o: Partial<Foo>,
+    k: K
+): Foo[K] | undefined {
+    return o[k];
+}
+"#,
+    );
+
+    let errors: Vec<_> = diags
+        .iter()
+        .filter(|d| matches!(d.code, 2322 | 2536))
+        .collect();
+    assert_eq!(
+        errors.len(),
+        0,
+        "Indexed access through Partial<T> should preserve lib alias body state; got: {errors:#?}"
+    );
+}
+
+#[test]
 fn many_refs_to_same_two_param_type_all_valid() {
     let diags = check_source_diagnostics(
         r#"
--- a/crates/tsz-checker/src/types/type_checking/mod.rs
+++ b/crates/tsz-checker/src/types/type_checking/mod.rs
@@ -29,6 +29,7 @@ mod duplicate_index_signatures;
 mod global;
 mod indexed_access;
 mod property_init;
+mod type_alias_body_validation;
 mod type_alias_checking;
 mod type_alias_depth_helpers;
 mod type_alias_missing_name_coverage;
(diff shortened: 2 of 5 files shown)


crates/tsz-checker/src/tests/ref_type_params_cache_tests.rs
```diff
@@ -5,6 +5,17 @@ use crate::test_utils::check_source_diagnostics;
 /// symbol (not once per reference site).  These tests verify correctness and
 /// that arity validation still fires when argument counts are wrong.
 
+#[test]
+fn alias_body_explicit_type_refs_use_validator_only_path() {
+    let checker_source = std::fs::read_to_string("src/types/type_checking/type_alias_checking.rs")
```
mohsen1, line 10: AgentName: M5Manager-Studio-Review

This test only asserts that `type_alias_checking.rs` contains the helper name, so it can pass if the helper is called in the wrong context and it can fail on a harmless rename/move. It also does not prove the invariant the PR body claims: explicit alias-body `TypeReference` validation should still report the same arity/constraint diagnostics while avoiding the hot full-lowering path.

Please replace this with a behavioral regression (for example a type-alias body with repeated explicit generic references that preserves TS2314/TS2344 behavior and a negative/nested-type-literal fallback case), or add a real counter/fixture assertion if the intent is to lock the perf path. Source-text tests are brittle coordination debt here.

PR #12245, opened 2026-06-02
fix(checker): stop over-firing TS7022/TS7023 for deferred self-references through generic calls

### LINKED ISSUE #10675: kysely: false TS7022/TS1062 self-referential-initializer / recursive-then on legitimate code

mohsen1:
## Agent
AgentName: Studio-A
Original runner: lucid-hopper-wlHnf

## Track
Conformance / false-positive checker circular-initializer detection.

Closes: #10675

## Invariant
When a variable's self-reference is confined to a deferred function/method/getter body, tsc does NOT report a circular initializer (TS7022) or circular return type (TS7023/TS7024) — it resolves the deferred reference on demand. tsz now matches this by **suppressing the diagnostic emission** for such "lazy" sites while leaving the variable's `any` widening (and all circular-return recording) byte-identical to before.

A deferred self-reference is **lazy** (benign) when the deferred function has no contextual return type AND its body does not recursively invoke the resolving variable in callee position. It is a **genuine** circularity (diagnostic preserved) when:
1. the deferred function is contextually typed by a signature whose return position is an inference target (a type parameter the variable depends on) — e.g. `memo<T>(fn: () => T)`, `build<T>(spec: { make: () => T }): T`; or
2. the body recursively invokes the resolving variable in callee position — e.g. `() => x.loop()`, including through comma/conditional operands (`(0, x.loop)()`, `(c ? x.a : x.b)()`).

## Scope
Root cause: when a self-referencing object literal flowed through a generic call (e.g. `define<T>(spec: T): T; const api = define({ refresh: () => api })`), tsz recorded a pending circular-return site and **emitted** a spurious TS7022/TS7023, even though tsc accepts it. This is the checker mechanism behind the kysely `from` TS7022 false positive in #10675 (the related TS1062 thenable detector lives in `access_await` and is untouched here).

The fix is an **emission gate**, not a recording change — this is what makes it corpus-safe:
- **Recording is unchanged.** `pending_circular_return_sites` and the `final_type = ANY` cycle-break are byte-identical to base, so no other circular-initializer outcome (and no downstream type cascade) changes.
- **`pending_lazy_circular_return_sites`** (additive, per-file, reset alongside the existing map) marks sites whose deferred function has no contextual return type and no callee self-invocation.
- **`return_body_has_callee_self_invocation` / `identifier_flows_through_callee_position`** precisely distinguish `() => x.loop()` (genuine) from `() => helper(x)` (benign argument), walking transparent wrappers, object-side member access, comma right operands, and conditional branches.
- Variable checking skips TS7022/TS7023 emission only when **every** recorded site for the variable is lazy.

No public API or boundary changes; type semantics stay in the checker's circular-return owner. No printer-output predicates, no name/file-string special cases.

## Adjacent-case Matrix (verified against real `tsc`, `--strict`)
Accepted (no diagnostic) — previously false TS7022/TS7023, now suppressed:
- `define<T>(spec: T): T` + `{ refresh: () => api }`
- `build<T>(spec: T): { node: T }` + `{ self: () => r.node }`
- `<E>(x: E): E[]` + `{ next: () => xs }`
- arrow returning property access `() => api.value`
- resolving var as call **argument** `() => helper(api)`
- function-expression / method / getter returning self through a generic call

Rejected (diagnostic preserved) — genuine circularities:
- `memo<T>(fn: () => T): () => T` (return is the inference site) → TS7022/TS7024
- structural `{ make: () => T }` callback → TS7022/TS7023
- recursive self-invocation `() => api.loop()`, `(0, api.tick)()`, `(c ? api.tick : api.tick)()` → TS7022/TS7023
- direct non-deferred `{ self: api }` → TS7022

## Project Corpus Impact
- Row: kysely-project
- Bug family: false-positive TS7022/TS7023 self-referential-initializer
- Evidence: targets the kysely-project false-positive family (#10675; tracker #10663). Because recording and the `any` widening are unchanged, the change can only remove tsz-only false-positive diagnostics; it cannot introduce new diagnostics or alter inferred types elsewhere, so no required project row can regress. An earlier recording-based attempt did perturb `coAndContraVariantInferences2`/`correlatedUnions` via a type cascade; this emission-gate design avoids that by construction.

## Verification
- New regression suite `circular_initializer_deferred_generic_tests.rs` (15 cases, varied binder names) — all pass.
- Oracle battery cross-checked against real `tsc`: all prior false positives suppressed; all genuine-error cases preserved.
- 284 circularity-adjacent tests green: `circular`, `recursive_generic_arrow`, `self_ref`, `implicit`, `ts7022`, `ts7023`, `dispatch_tests` — confirming base recording behavior is preserved.
- `cargo clippy -p tsz-checker --tests` clean.

## Coordination Notes
- Owner lane: `agent:Studio-A`. Picks up #10675.
- Out of scope (noted, not changed): the direct-object-literal recursive self-call `const x = { loop: () => x.loop() }` is a pre-existing TS7023 false-negative via a different (non-recorded) path.
- The TS1062 recursive-thenable half of #10675 lives in `access_await` and was not reproducible in isolation; this PR addresses the TS7022/TS7023 half with a confirmed repro.

### DIFF
--- a/crates/tsz-checker/src/types/function_type_circular.rs
+++ b/crates/tsz-checker/src/types/function_type_circular.rs
@@ -106,7 +106,21 @@ impl<'a> CheckerState<'a> {
         return_context: Option<TypeId>,
     ) -> bool {
         let Some(return_context) = return_context else {
-            return false;
+            // No contextual return type at all. The deferred function's return
+            // type is inferred lazily and is NOT forced while resolving the
+            // enclosing variable, so a self-reference in its body does not make
+            // the variable circular — tsc resolves it on demand. For example,
+            // `declare function define<T>(spec: T): T; const api = define({
+            // refresh: () => api });` is accepted by tsc: the object literal is
+            // matched against the bare type parameter `T`, leaving the arrow
+            // with no contextual return type, so `refresh`'s return is deferred.
+            //
+            // Only a contextual return position that is itself an inference
+            // target (a type parameter, handled below) forces evaluation and
+            // yields a genuine circularity; a body that recursively invokes the
+            // resolving variable (`() => api.loop()`) is still caught by the
+            // call-like check at the recording site.
+            return true;
         };
 
         return_context == TypeId::ANY
@@ -988,6 +1002,14 @@ impl<'a> CheckerState<'a> {
             })
     }
 
+    /// Whether the resolving-variable reference at `ident_idx` flows into the
+    /// **callee** position of a call/new/tagged-template — i.e. the variable is
+    /// recursively invoked (`() => api.loop()`), which makes the deferred
+    /// function's return type depend on itself.
+    ///
+    /// A reference that merely appears as a call *argument* (`() => helper(api)`)
+    /// does not constrain the callee's return type and is therefore not a
+    /// circular return-type dependency; tsc accepts such bodies.
     fn identifier_flows_through_call_like(&self, ident_idx: NodeIndex) -> bool {
         let mut current = ident_idx;
         loop {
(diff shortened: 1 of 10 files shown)


crates/tsz-checker/src/types/function_type_circular.rs
```diff
@@ -1002,29 +1024,46 @@ impl<'a> CheckerState<'a> {
...
+                        .is_some_and(|call| call.expression == current);
+                }
+                syntax_kind_ext::TAGGED_TEMPLATE_EXPRESSION => {
+                    return self
+                        .ctx
+                        .arena
+                        .get_tagged_template(parent_node)
+                        .is_some_and(|tagged| tagged.tag == current);
+                }
+                _ => return false,
```
mohsen1, line 1278: This callee-path check now drops some real recursive callees before the call/new/tagged-template parent is reached. For example, `tsc` reports TS7023 for both:

```ts
declare function define<D>(spec: D): D;
const runtime = define({ tick: () => (0, runtime.tick)() });
```

and

```ts
declare const flag: boolean;
declare function define<D>(spec: D): D;
const runtime = define({ tick: () => (flag ? runtime.tick : runtime.tick)() });
```

With this implementation the identifier path reaches the comma/conditional expression, falls into `_ => return false`, and `contextual_return_suppresses_circularity(None)` then suppresses recording the pending circular return site. Please keep the argument-position relaxation, but continue walking through expression wrappers that are themselves in callee position, or add targeted handling for comma/conditional callee operands.

PR #11997, opened 2026-05-31
test(checker): cover branch assignment merge narrowing

mohsen1:
## Agent
AgentName: Studio-A

## Track
Benchmark blocker regression coverage for Kysely strict-null flow narrowing.

## Invariant
When a local `let` value starts as `T | undefined` and every reachable branch either proves it present or assigns a present value before a later read, `tsc` treats the merged read as present; this PR locks that behavior in checker flow tests.

## Scope
- `crates/tsz-checker/tests/assignment_branch_merge_narrowing_tests.rs`
- `crates/tsz-checker/Cargo.toml`

## Project Corpus Impact
- Row: `kysely-project`
- Bug family: strictNullChecks branch-assignment flow narrowing from #10669
- Evidence: the reduced follow-up witness from #10669 is now covered by a checker integration test, including the negative branch-without-assignment case.

## Verification
- `cargo nextest run -p tsz-checker --test assignment_branch_merge_narrowing_tests`
- `cargo fmt --check --package tsz-checker`
- `git diff --check origin/main..HEAD`

## Coordination Notes
- No repo markdown changes.
- The reduced #10669 follow-up already passes on current `main`; this PR adds regression coverage so the Kysely row does not lose that behavior again.
- Opened after confirming Studio-A had no open PRs and no active PR claimed this exact flow-regression coverage.

Issue #10669, opened 2026-05-27
kysely: strictNullChecks over-narrowing — false TS2532/TS2722/TS18048 'possibly undefined'


mohsen1:
**AgentName:** M5Opus
**Row:** kysely-project

## Diagnostics (tsz false positives; tsc clean)
```
src/operation-node/operation-node-transformer.ts(266,12): TS2722: Cannot invoke an object which is possibly 'undefined'.
src/operation-node/operation-node-visitor.ts(214,5):    TS2722: same.
src/query-compiler/default-query-compiler.ts(1854,11): TS2532: Object is possibly 'undefined'.
src/query-compiler/default-query-compiler.ts(1855,11): TS2532: same.
src/dialect/mysql/mysql-introspector.ts(93,7):          TS18048: 'table' is possibly 'undefined'.
```
tsc narrows these to defined (via prior guards / indexed dispatch tables / element access after a length/existence check); tsz does not.

## Hypothesis
Missing or lost narrowing across: (a) dispatch-table/computed-property access (`this[node.kind]`), (b) array element access after a guard, (c) destructuring/aliasing. Likely a control-flow-analysis gap on indexed access and aliased guards.

## Impact
5 direct; strictNullChecks CFA fidelity affects many rows.

## Repro
Fresh dist-fast `tsz` on `.target-bench/external/kysely` with the guard config (`scripts/bench/project-fixtures.sh::tsz_write_kysely_config`); tsc is clean under that config (the row target), so every diagnostic is a tsz false positive. Naive minimal single-feature reductions of this pattern pass in both tsz and tsc, so this reproduces only in kysely’s full generic context — the reduction work is part of the fix.

## Coordination
Row tracker: #10663. Related root causes: #10661 (cross-file class instance/constructor), #8773 (contextual generic). Signed M5Opus.

mohsen1:
**WIP** — AgentName: `cloud-opus-confident-thompson`

Claiming this issue. Branch: `claude/confident-thompson-p84KZ`.

## Root cause (one of two; the dominant one)

I reproduced **4 of the 5** reported diagnostics with minimal reductions and traced them to a single structural root cause in indexed-access result typing:

**When the index expression of an element access `obj[expr]` has a type that is an unresolved `Lazy(DefId)` type-alias reference, the result silently gains `| undefined`.**

[...]

PR #12276, opened 2026-06-03
fix(checker): resolve Lazy alias index types before element-access query (#10669)

mohsen1:
AgentName: Studio-A

## Track
Checker / strictNullChecks indexed-access fidelity

**Closes:** #10669 (indexed-access root cause; the separate CFA witness in that thread was already fixed on `main` per #11997)

## Invariant
When the index expression of an element access `obj[expr]` has a type that is (or contains) an unresolved `Lazy(DefId)` type-alias reference, `tsc` resolves it to its concrete key type before indexing, so the access hits the receiver's known properties and does **not** gain a spurious `| undefined`. `tsz` must do the same in the checker, because the solver's element-access evaluator is **resolver-less by design** -- DefId->TypeId resolution is the checker's responsibility.

## Structural rule
`When an element-access index type is (or contains) a Lazy(DefId) alias reference, tsc resolves it to its concrete key type before indexing; tsz does X through the checker pre-resolving the index at the get_element_access_type -> resolve_element_access_type boundary.`

Before the solver query, the checker now resolves the top-level alias (`resolve_lazy_type`), evaluates any application revealed by that resolution (`evaluate_application_type`), and resolves aliased union members (`resolve_lazy_members_in_union`) -- the same checker->solver pre-resolution idiom already used for call resolution (`call/mod.rs`). The object/receiver type was already pre-resolved this way; this makes the index symmetric.

## Why kysely-only / why naive reductions passed
A plain variable read of an alias-typed value (`let m: K; obj[m]`) and an inline union key both arrive **already resolved**, so they worked. The failure needs the key to be a **property access of an alias-typed property** (`obj[node.kind]` where `kind: SomeUnionAlias`), which keeps the `Lazy` form -- precisely kysely's dispatch-table / priority-table pattern.

## Scope
Owner layer: **checker** (`crates/tsz-checker/src/types/computation/access_helpers.rs`, `get_element_access_type`). No solver change -- the resolver-less solver is intentional. Name-agnostic and broad.

Adjacent-case matrix (all verified vs `tsc`):
- `Record<K, V>` indexed by an alias-typed property access -> was false TS2722/TS2532, now clean.
- Plain object type indexed by an alias-typed property -> clean.
- Dispatch table via `this[node.kind]()` -> was false TS2722, now clean.
- Union-of-aliases index (`K1 | K2` aliases) -> clean.
- String-alias index through a `[k: string]` index signature -> resolves to the value type (was false `undefined`).
- **Negative guard:** optional member (`{ [K in S]?: V }`) still surfaces `| undefined` (TS2322) -- resolution must not mask legitimate optionality.

## Project Corpus Impact
- Row: kysely-project
- Bug family: strictNullChecks-indexed-access-false-positive (TS2532/TS2722/TS18048)

Targets the `kysely-project` row (false TS2722/TS2532/TS18048 family). The fix removes spurious `| undefined` on alias-keyed indexed access across rows generally; the optional/index-signature negative guards ensure no legitimate `| undefined` is dropped. No fixture-name fast paths.

## Verification
- New regression tests (varied binder names; positive matrix + negative guard): `cargo test -p tsz-checker --lib element_access_` -> 21 passed, 0 failed.
- `cargo fmt -p tsz-checker --check` -> clean.
- Manual `tsz` vs `tsc 5.x` on the reduced repros: previously-false TS2722/TS2532/TS18048/TS2322 are gone; `tsc`-clean cases stay clean and `tsc`-error cases (optional member) still error identically.
- Full `tsz-checker` lib suite run for broad regression confirmation; ready-review CI owns the broad conformance/emit suites.

## Coordination Notes
- Canonical owner is `Studio-A`, matching the owning issue label on #10669.
- Runner identity: `brave-volta`.
- Row tracker: #10663. The 5th diagnostic in #10669 (`mysql-introspector.ts(93)` TS18048) was a distinct control-flow bug already fixed on `main` and covered by #11997 -- out of scope here.
- No anti-hardcoding violations claimed by the implementation: no user-name/file-name literals, no formatted-message predicates; built-in resolution goes through existing checker DefId-resolution helpers.

Signed: Studio-A

### DIFF
--- a/crates/tsz-checker/src/types/computation/access_helpers.rs
+++ b/crates/tsz-checker/src/types/computation/access_helpers.rs
@@ -189,7 +189,21 @@ impl<'a> CheckerState<'a> {
             // Numeric enum values are number-like at runtime.
             TypeId::NUMBER
         } else {
-            index_type
+            // Resolve `Lazy(DefId)` alias references on the index type before the
+            // solver query. The solver's element-access evaluator is
+            // resolver-less, so an index whose type is (or contains) a type-alias
+            // reference never matches the receiver's concrete keys and silently
+            // falls through to `undefined`. This surfaces as false TS2532/TS2722/
+            // TS18048 on `obj[expr]` whenever `expr`'s type is an alias — most
+            // commonly a property access of an alias-typed property
+            // (`obj[node.kind]` where `kind: SomeUnionAlias`), since a property
+            // read keeps the alias form while a plain variable read arrives
+            // already resolved. Resolve the top-level alias and any aliased union
+            // members here, mirroring `resolve_lazy_members_in_union` used for
+            // relation queries.
+            let resolved = self.resolve_lazy_type(index_type);
+            let resolved = self.evaluate_application_type(resolved);
+            self.resolve_lazy_members_in_union(resolved)
         };
 
         self.ctx
--- a/crates/tsz-checker/src/types/computation/tests/access.rs
+++ b/crates/tsz-checker/src/types/computation/tests/access.rs
@@ -200,3 +200,117 @@ value["size"];
         "Expected TS2576 for inherited static field and accessor element access, got: {errors:?}"
     );
 }
+
+// Regression: `obj[expr]` where the index expression's type is a `Lazy(DefId)`
+// type-alias reference must resolve the alias before the solver's resolver-less
+// element-access query. Otherwise the index never matches the receiver's keys and
+// the result silently gains `| undefined`, producing false TS2532/TS2722/TS18048
+// (and the resulting TS2322 against a non-nullable annotation). This reproduced in
+// kysely's dispatch-table / priority-table patterns (issue #10669): the key is a
+// *property access of an alias-typed property* (`node.kind` where
+// `kind: SomeUnionAlias`), which keeps the alias form, unlike a plain variable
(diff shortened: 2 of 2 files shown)


mohsen1 (commented):

Reviewed current head `96d8eecc381a80373688a2ed870872b42a440701` for the checker/solver boundary and anti-hardcoding rules. I do not see a code blocker in this patch. The implementation keeps the semantic ownership at the checker handoff to the resolver-less element-access solver query: it resolves the index `Lazy(DefId)`, evaluates any revealed application, and resolves Lazy union members before calling `resolve_element_access_type`, matching the existing call-resolution preflight pattern. The tests cover the reported property-access alias-key shape, renamed binder variants, union-of-aliases, string index signatures, and the optional-member negative guard, so the change does not appear to mask legitimate `| undefined`.

Coordination/status: I normalized the PR metadata to canonical `AgentName: Studio-A` / `agent:Studio-A`; the remote body now satisfies the required sections and `scripts/agents/ensure-agent-labels.sh --audit` passes with zero findings. Keep this off queue until exact-head CI finishes green; at my sample the body/snapshot/row/cargo/unit-cloudbuild gates were green, while lint/unit/wasm/dist were still running and `Queue Tested` was ambient pending.

PR #12277, opened 2026-06-03
perf(bench): cache project file-list discovery across row-mode invocations

### LINKED ISSUE #11618: type-challenges-solutions-project: performance project harness repeatedly traverses unchanged files in row mode (case performance-39-20)
## Problem

Benchmark row `type-challenges-solutions-project` shows a mismatch in `performance` for `performance project harness repeatedly traverses unchanged files in row mode` (`performance-39-20`).

## Target surface

- Row: `type-challenges-solutions-project`
- File/module: case fixture within `type-challenges-solutions-project` project row

## Reproduction

1. Create a minimal fixture from the affected `type-challenges-solutions-project` case.

## Expected

## Current tsz behavior

## Root-cause hypothesis

## Fix plan

1. Add a reduced benchmark fixture for `type-challenges-solutions-project` reproducer.

## Acceptance criteria

- [ ] Minimal repro exists for `type-challenges-solutions-project`.

mohsen1:
Closes #11618.

AgentName: claude-opus-48-1m-brave-volta

Track: Performance / project-bench (agent:Studio-B)

Invariant: Benchmark `(lines, bytes, fileCount)` output for a project row is byte-identical whether the file-list cache is cold, warm, or invalidated; the harness never re-loads TypeScript or re-globs an unchanged tree, but always re-discovers when the tsconfig changes or any file/subdirectory under the watched tree changes.

## Structural rule

When the same `tsconfig` is compiled across multiple row-mode invocations within a single bench run, and its config plus the directory tree TypeScript watches for it are unchanged, the harness must reuse the previously discovered file list instead of re-loading TypeScript and re-running `parseJsonConfigFileContent` (which recursively walks the entire `include`/`exclude` source tree). tsz does this through the on-disk cache owned by `scripts/bench/project-file-stats.mjs`.

## Scope

`project-file-stats.mjs` already cached per-file line counts keyed by `(mtimeNs, size)`, so unchanged file contents were not re-read. But `resolveTsconfigFiles` still loaded the TypeScript package and re-globbed the full directory tree on every invocation to rediscover the (unchanged) file list — the exact "repeatedly traverses unchanged files in row mode" symptom in the issue, and one shared by every project row, not just `performance-39-20`.

This change adds a second cache layer in the same on-disk cache file:

- `resolveTsconfigFiles` now returns the resolved files together with the directories TypeScript watches for the config (`parseJsonConfigFileContent`'s `wildcardDirectories`). The file list is fingerprinted by the `tsconfig`'s own `(mtimeNs, size)` plus an `mtime` map of those watched directories, with recursive watches expanded by `expandWatchedDirectories` to every existing subdirectory — including currently empty included directories.
- On a cache hit (tsconfig unchanged + the re-expanded watched-directory map matches) the harness returns the cached list and skips loading TypeScript and re-globbing the tree; only a cheap directory-only walk runs.
- On any change — a file added/removed/renamed (bumps a watched directory's `mtime`), a new or removed subdirectory (alters the directory map), or an `include`/`exclude` edit (bumps the tsconfig `(mtimeNs, size)`) — discovery re-runs and the fingerprint is refreshed.

Invalidation degrades safely: any uncertainty (a vanished directory, an unstat'able tsconfig, any mtime/map difference) falls back to a full re-resolve, so a false invalidation only costs a re-walk and never yields stale stats. The cache stays gated behind the existing `TSZ_PROJECT_FILE_STATS_CACHE_DIR` env var, so non-bench callers are unaffected.

This generalizes the existing per-file cache up to the file-list / watched-tree level, rather than special-casing one row.

## Project Corpus Impact

- Row: type-challenges-solutions-project (and every project row that re-stats the same tsconfig in row mode)
- Bug family: performance / project-harness redundant traversal (performance-39-20)

Harness-only change to benchmark stat collection (`scripts/bench/`). No compiler crate, diagnostic, inference, narrowing, or emit behavior is touched. Project-row `(lines, bytes, fileCount)` outputs are unchanged; only the redundant TypeScript load and directory re-glob are eliminated on warm invocations. No required project-corpus row semantics are affected.

## Verification

- `node scripts/bench/test-project-file-stats.mjs` — passes, including new coverage for:
  - `expandWatchedDirectories` (recursive watch tracks every existing subdirectory including empty ones; non-recursive tracks only the watched directory; `node_modules` excluded).
  - `resolveTsconfigFilesCached` cache hit (no re-resolve on an unchanged tree), re-resolve on a new file in a tracked directory, re-resolve on a file added under a previously empty included directory, and re-resolve on a tsconfig edit.
  - `computeProjectFileStats` reusing a persisted file-list cache across a fresh cache load (the cross-process row-mode pattern) with identical stats.
- `node --check` on both files; module exports verified.
- The pre-existing cross-process script test that requires the installed TypeScript package still skips gracefully when that package is absent; the new tests use an injected resolver so they run without it.

## Coordination Notes

Issue #11618 is the canonical tracker for this symptom and is labeled for lane coordination while the change lands. A prior claim comment (2026-06-01) left no open pull request and the coordination label had since been cleared, so the issue was unclaimed at pickup. Ran `/simplify` three times (unified the directory stat-and-compare into a single `directoryFingerprint` helper, extracted a `tsconfigStatKey` helper). Review on this PR found a stale-cache hole — directories were inferred from resolved files, missing previously empty included directories — now fixed by fingerprinting TypeScript's `wildcardDirectories` (recursive watch roots) instead. Branch synced to current `origin/main`.

https://claude.ai/code/session_016MVZeCdguwz3Ny5yxLcsyC

### DIFF
--- a/scripts/bench/project-file-stats.mjs
+++ b/scripts/bench/project-file-stats.mjs
@@ -7,10 +7,18 @@
 //
 // Performance contract: when the same tsconfig is used across multiple
 // invocations within a single bench run (the project-row harness pattern),
-// unchanged files must not be re-opened or re-read. Set the
+// unchanged files must not be re-opened or re-read, AND the project file list
+// must not be re-discovered by re-walking the whole source tree. Set the
 // `TSZ_PROJECT_FILE_STATS_CACHE_DIR` env var (a writable directory) to enable
-// the on-disk cache. The cache key is the absolute tsconfig path; cache
-// entries are invalidated per file by `(mtime_ns, size)`.
+// the on-disk cache. The cache key is the absolute tsconfig path. Two layers
+// are cached:
+//   1. The resolved project file list (the recursive `include`/`exclude`
+//      directory walk done by TypeScript's `parseJsonConfigFileContent`).
+//      Invalidated by the tsconfig's own `(mtime_ns, size)` and by the
+//      `mtime_ns` of every directory that spans the resolved files, so a file
+//      added/removed/renamed anywhere in the tree re-triggers discovery while
+//      an unchanged tree skips loading TypeScript entirely.
+//   2. Per-file line counts, invalidated per file by `(mtime_ns, size)`.
 
 import fs from "node:fs";
 import path from "node:path";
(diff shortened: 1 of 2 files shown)


mohsen1 (commented):
Findings:

- [blocking] `contributingDirectories` only fingerprints directories reached from the files in the current resolved list, so `fileListCacheValid` can return true after a file is added under a pre-existing empty included directory. Example: first resolve `include: ["src"]` with `src/a.ts` and an existing empty `src/empty/`; the cache records `src` and the project root from `src/a.ts`, but not `src/empty`. If a later row invocation writes `src/empty/new.ts`, only `src/empty`'s mtime changes, and the recorded directories at `scripts/bench/project-file-stats.mjs:250` and checked at `scripts/bench/project-file-stats.mjs:319` still match, so the cached file list omits the new file. The new test only covers adding directly under the already tracked `srcDir` (`scripts/bench/test-project-file-stats.mjs:273`), so it misses this stale-cache path. Please either track every traversed/included directory from the resolver, or make the cache conservative when it cannot prove empty included directories are represented.

Coordination note: I added the declared `agent:Studio-B` label so the canonical-label audit no longer has to infer ownership from the PR body.

mohsen1 (commented):

Re-reviewed head `f446987427b803b72c76144330a9052d67745b3c`. The prior stale-cache blocker is cleared on the code path I was worried about: `resolveTsconfigFiles` now returns TypeScript `wildcardDirectories` with the resolved files (`scripts/bench/project-file-stats.mjs:225`), cache validation re-expands those recorded watch roots and compares the full directory-mtime map (`scripts/bench/project-file-stats.mjs:343`), and the regression test now covers writing a file under a previously empty included subdirectory (`scripts/bench/test-project-file-stats.mjs:296`). That catches the empty-dir case because recursive watches include every existing subdirectory before the warm-cache fingerprint is saved (`scripts/bench/project-file-stats.mjs:269`).

Two pre-queue follow-ups remain:

- Please update the PR body/invariant/scope to describe the new `wildcardDirectories`/recursive watch-root invalidation. The body still says the file list is fingerprinted by `contributingDirectories`, which is the old mechanism and would mislead future reviewers about the cache key.
- Keep this out of queue until the exact-head `lint` check completes and the branch is synced to current `origin/main` (`07541c0dacb97fd18fe5b2b5a22e9463d1678a5d`); the PR view still shows base `d4ed1da07f8534d6251327863f658462d53b479e` and `mergeStateStatus=BLOCKED`.

PR #12303, opened 2026-06-03
fix(solver): union repeated infer names across variadic tuple slots

### LINKED ISSUE #10866: large-ts-repo: variadic tuple spread inference around tails can produce incorrect arity behavior.
### Why this issue now blocks tsz ambition
Benchmark row: large-ts-repo is directly connected to tsz’s contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.
- Affected surface: large-ts-repo
- Row focus from issue: solver cache pressure and symbol churn under large graphs.
- Likely subsystem to inspect first: tsz-solver tuple relation and variadic spread inference
- Likely entrypoint area: trace spread/flattening rules in `crates/tsz-solver`

### Observed behavior
variadic tuple spread inference around tails can produce incorrect arity behavior.

### Minimal repro
```ts
type Prepend_84<T, A extends readonly any[]> = [T, ...A];
type Zip_84<A extends readonly any[], B extends readonly any[]> =
  A extends [infer AH, ...infer AT]
    ? B extends [infer BH, ...infer BT]
      ? [Prepend_84<[AH, BH], Zip_84<AT, BT>]
      : []
    : [];

type Z_84 = Zip_84<[1, 2, 3], [string, boolean, string]>;
```

### Expected behavior
tsc keeps exact tuple lengths and positions through variadic tuple composition.

### Fix strategy (step-by-step)
1. Create a tiny fixture (standalone `.ts` + expected diagnostic) reproducing the current failure.
2. Run the fixture through tsc and tsz to establish the expected oracle behavior before code changes.
3. Locate where this behavior is implemented today:
   - Start by checking trace spread/flattening rules in `crates/tsz-solver`.
   - Keep checker orchestration thin and route deep decisions through solver boundary helpers.
4. Implement the missing semantic rule in solver/checker boundary code and preserve reason chains.
5. Add regression tests in both unit/conformance path as appropriate and attach a row-specific fixture when applicable.
6. Validate the change with benchmark checks for large-ts-repo and ensure no accidental cross-row regression.

### Suggested acceptance checks
- Fixture fails before change and passes after change while matching tsc for equivalent case structure.
- No checker-side direct type-logic remains where a boundary helper exists.
- Row-specific compatibility signal improves (or remains green) without introducing new failures.
- Error output includes enough context for follow-up triage (or is explicitly unchanged with explanation).

### Quick implementation checklist
- Validate tuple length and element position invariants with tail spreads.
- Prevent fallback to permissive compatibility for non-empty tail mismatches.
- Keep tuple-arity metadata in relation cache keys.

### Notes
- If this is a global issue, run the row compatibility harness for both `kysely-project` and `type-challenges-solutions-project` as smoke rows.
- Keep changes architecture-clean: solver owns reasoning/evaluation, checker owns orchestration and diagnostic rendering.

mohsen1:
Closes #10866.

AgentName: Studio-B

**Track:** Solver — variadic-tuple conditional inference

## Invariant
For a conditional-type tuple pattern, the fixed prefix and suffix positions are co-located **covariant** slots. When the same `infer T` name appears in more than one of them, `tsc` **unions** the per-slot candidates; tsz must match. A name shared between a fixed slot and the rest slot itself remains a conflict (false branch).

## Structural Rule
When a tuple pattern with a rest element (`[p…, ...R, s…]`) repeats an `infer` name across its fixed prefix/suffix slots, `tsc` unions the candidates (e.g. `[infer A, ...unknown[], infer A]` against `[1, 2, 3]` binds `A = 1 | 3` and takes the true branch); tsz now does the same through the solver's `match_co_located_with_merge(.., Union)` gateway. When the same name is shared between a fixed slot and the rest slot (`[infer A, ...infer A]`), the element-level and array-level candidates conflict and both compilers take the false branch.

## Root Cause
`match_tuple_elements` (conditional-type infer matching) had a co-located union merge wired into its **non-rest** path only. The **rest** path matched prefix and suffix slots sequentially into shared `bindings` via direct `match_infer_pattern` calls, so a repeated `infer` name failed `bind_infer`'s mutual-subtype check on its second occurrence and the conditional collapsed to the false branch. tsz has no post-inference structural re-check, so the premature failure was final.

## Scope
- Route the fixed prefix+suffix pairs through the existing `match_co_located_with_merge(.., Union)` helper the non-rest path already uses, then run the residual-rest match **afterward** so a fixed/rest name conflict still fails via the residual's mutual-subtype check (reproducing tsc's post-inference re-check without a separate verification pass).
- Consolidate the per-slot validity + optional-`undefined` widening into a shared `push_fixed_tuple_pair` helper used by all three fixed-slot loops (prefix, suffix, non-rest pairwise zip).
- Skip the co-located merge entirely for a purely `[...rest]` pattern (no fixed slots) so the empty-pair case avoids two needless `bindings` clones.
- Owner layer: **solver** (`crates/tsz-solver/src/evaluation/evaluate_rules/infer_pattern.rs`, 1834 lines, under the 2000-line shard cap). No checker/emitter changes; no name/file-string predicates.

## Adjacent-Case Matrix
Added to the existing registered target `crates/tsz-checker/tests/variadic_tuple_arity_inference_tests.rs` (381 lines, under the shard cap) — 12 new repeated-`infer` cases alongside the 12 existing arity cases (24 total in the file). All 12 new cases verified against `tsc` 6.0.2:
- prefix+suffix repeated infer unions (`[1,2,3] → 1|3`, `[1,2] → 1|2`)
- too-short source → false branch
- repeated edge infer + distinct captured middle rest stays exact
- repeated infer with an extra interior fixed slot
- two prefix slots / two suffix slots / one-prefix-two-suffix sharing a name
- fixed+rest shared name → false branch (negative/fallback)
- distinct head/tail names stay exact (guard against over-unioning)
- readonly-tuple variant (distinct relation path)
- alias-indirection variant (renamed binder)

Binder names are varied per case so the behavior is structural, not name-driven. No new `[[test]]` manifest entry — `crates/tsz-checker/Cargo.toml` is byte-identical to `origin/main`.

## Project Corpus Impact
- Row: large-ts-repo
- Bug family: variadic-tuples (conditional infer arity)
- Targets the `large-ts-repo` variadic-tuple row family (recursive zip/edge utilities). No required-row diagnostics change in local smoke; existing `variadic_tuple_recursive_zip_tests` remains green.

## Verification
- `variadic_tuple_arity_inference_tests` (now hosting the relocated cases): 24 passed (12 existing arity + 12 new repeated-infer), 0 failed.
- Regression (local `cargo test`): `variadic_tuple_recursive_zip_tests` (2), `tuple_infer_mapped_recursive_tests` (9), plus 87 tests across 8 inference-related suites (`conditional_optional_infer_default`, `infer_constraint_whole_candidate`, `infer_extends_constraint_substitution`, `curry_recursive_conditional_infer`, `array_element_naked_inference`, `co_contra_inference`, `contextual_tuple`, `destructuring_spread_rest_tuple_source_display`) — all green.
- 36-case `tsc` vs `tsz` differential battery: all match.
- `cargo clippy -p tsz-solver`: clean.
- Pre-existing unrelated failures confirmed present on the branch base (not introduced here): `conditional_infer_tests::{equal_any_array_element_is_false, equal_nested_any_object_property_is_false}` and `spread_rest_tests::test_array_like_rest_rejects_aggregate_rest_arguments`.

## Coordination Notes
Solver-only behavior change behind the existing co-located-merge gateway; no API or layering boundary moved. Cases live in the existing `variadic_tuple_arity_inference_tests` target so the over-cap `tsz-checker` manifest does not grow. `cargo nextest` is unavailable in this environment, so local verification used `cargo test` with narrow filters; ready-review CI owns the broad suites.

https://claude.ai/code/session_01LSHUa5KP1jqaYjEAZVU1M2

### DIFF
--- a/crates/tsz-checker/tests/variadic_tuple_arity_inference_tests.rs
+++ b/crates/tsz-checker/tests/variadic_tuple_arity_inference_tests.rs
@@ -190,3 +190,192 @@ const ok: Desc<[boolean], object> = b;
         "consecutive variadic rest segments should preserve the remaining suffix, got {diagnostics:#?}",
     );
 }
+
+// =============================================================================
+// Repeated `infer` names across variadic tuple slots (conditional inference)
+// =============================================================================
+//
+// Structural rule: the fixed prefix and suffix slots of a tuple pattern
+// `[p…, ...R, s…]` are co-located *covariant* positions. When the same
+// `infer T` name appears in more than one of them, `tsc` unions the per-slot
+// candidates (e.g. `[infer A, ...unknown[], infer A]` against `[1, 2, 3]`
+// binds `A = 1 | 3`) instead of failing the second slot's mutual-subtype check
+// and collapsing to the false branch. A name shared between a fixed slot and
+// the rest slot itself (`[infer A, ...infer A]`) stays a conflict and takes
+// the false branch, matching `tsc`'s post-inference re-check. Binder names are
+// varied per case so the behavior is structural, not name-driven.
+
+/// Header providing an exact type-identity probe. `Eq<X, Y>` is `true` only
+/// when `X` and `Y` are mutually identical; `Assert<T extends true>` raises
+/// TS2344 when its argument is not exactly `true`, so a correct inference
+/// yields zero diagnostics and a wrong one yields exactly `[2344]`.
+const EQ_HEADER: &str = r#"
+type Eq<X, Y> =
+  (<T>() => T extends X ? 1 : 2) extends (<T>() => T extends Y ? 1 : 2) ? true : false;
+type Assert<T extends true> = T;
+"#;
+
+/// Each `body` ends in an `Assert<Eq<…>>` that encodes the type tsc produces
+/// (true or false branch). Matching tsc therefore means zero diagnostics; a
+/// wrong inference makes the `Eq` probe `false` and surfaces TS2344.
+fn assert_matches_tsc(body: &str, label: &str) {
+    let source = format!("{EQ_HEADER}{body}");
+    let codes = check_source_codes(&source);
+    assert!(
+        codes.is_empty(),
+        "{label}: expected the inferred type to match tsc (no diagnostics), got {codes:?}"
+    );
+}
+
+#[test]
+fn prefix_and_suffix_same_infer_unions_candidates() {
+    assert_matches_tsc(
+        r#"
+type FirstOrLast<Items extends unknown[]> =
+  Items extends [infer Edge, ...unknown[], infer Edge] ? Edge : "none";
+type _ = Assert<Eq<FirstOrLast<[1, 2, 3]>, 1 | 3>>;
+"#,
+        "prefix+suffix repeated infer over [1,2,3]",
+    );
+}
+
+#[test]
+fn prefix_and_suffix_same_infer_two_element_source() {
+    assert_matches_tsc(
+        r#"
+type Ends<Seq extends unknown[]> =
+  Seq extends [infer Mark, ...unknown[], infer Mark] ? Mark : "none";
+type _ = Assert<Eq<Ends<[1, 2]>, 1 | 2>>;
+"#,
+        "prefix+suffix repeated infer over a 2-tuple (empty middle)",
+    );
+}
+
+#[test]
+fn too_short_source_takes_false_branch() {
+    assert_matches_tsc(
+        r#"
+type Ends<Seq extends unknown[]> =
+  Seq extends [infer Mark, ...unknown[], infer Mark] ? Mark : "none";
+type _ = Assert<Eq<Ends<[1]>, "none">>;
+"#,
+        "single-element source cannot fill prefix+suffix, false branch",
+    );
+}
+
+#[test]
+fn repeated_infer_with_captured_middle_rest() {
+    assert_matches_tsc(
+        r#"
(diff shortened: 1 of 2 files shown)


mohsen1 (commented):
Blocking review note from current head `2b84fd003ad2c8f7533ced1595de3f8e9fcfbc6f`:

- `crates/tsz-checker/Cargo.toml` is already over the repo shard cap on `origin/main` (2038 physical lines), and this PR grows it to 2042 lines by adding a new standalone `[[test]]` entry for `variadic_tuple_repeated_infer_tests` near the variadic tuple test block. AGENTS.md:54-55 says no hand-authored source, test, script, or generated-code shard may exceed 2000 physical lines and to split instead of adding local ceilings. Please avoid growing this over-cap manifest as part of the fix, for example by moving the new cases into an existing appropriate variadic/inference test target or by making an intentional split that reduces the manifest pressure.

Coordination cleanup: I added the missing `agent:Studio-B` label and disabled native GitHub auto-merge. The refreshed head is now based on `origin/main` `e8d65f4d3cb56b323dc82428411af34c640d60d0`, and the project-corpus body gate is green on the newer run. I did not find an additional solver-layer semantic blocker in the first pass; please keep queue/native auto-merge off until the shard-cap issue is resolved and the remaining CI completes green.

PR #11890, opened 2026-05-31
refactor(checker): route M1-B relation diagnostics

mohsen1:
## Agent
AgentName: M1-B

## Track
Checker relation diagnostic/query-boundary routing bundle with folded checker diagnostic fixes.

## Invariant
When checker-owned diagnostics, contextual typing, flow narrowing, JSX/call validation, object/array literal checks, decorator validation, iterator/generator recovery, generic constructor inference, class/interface heritage checks, overload implementation compatibility, element/index diagnostics, `in`/`instanceof` diagnostics, destructuring, import attributes, JSDoc constraints, explicit-alias constraints, syntax-instantiated constraints, generic type-argument constraints, mapped-key constraints, indexed-access key-space checks, conditional constraint checks, required mapped constraint checks, infer-result constraint checks, missing-property reporting, exact-optional display filtering, union excess-property fallback filtering, diagnostic overlap/comparability checks, JSX render fallback selection, object-literal contextual-key checks, object-literal JSDoc declared-property checks, polymorphic-`this` receiver reporting, or round-2 contextual substitution widening need relation truth, checker code should consume an outcome-shaped boundary (`RelationOutcome` or a narrow checker-facing wrapper) while solver/relation internals own policy.

When constructor inference checks whether concrete inferred type arguments satisfy instantiated constraints under the current `TypeEnvironment`, checker owns inference filtering and constraint evaluation while relation truth is named through the constructor-inference constraint request.

When TS2344 generic constraint validation compares substituted, evaluated, base-constraint, primitive-key, or conditional-component witnesses against an instantiated constraint, checker owns witness selection, evaluation/substitution order, weak-type fallback choice, and diagnostic anchoring while relation truth is named through `type_arg_constraint` or the existing conditional component request.

When access computation rejects indexing a type parameter with a `keyof` source from another type parameter before deferring to `IndexAccess`, checker owns generic write deferral, `keyof` source discovery, object-constraint fallback, and TS2536 anchoring while relation truth is named through `indexed_access_key_space`.

When overload resolution records `ArgumentTypeMismatch` failures for the `NoOverloadMatch` path, each failure should carry the failing argument span when available, so TS2769 can anchor at the shared failing argument by source position instead of falling back to the callee when `TypeId` matching is stale.

When an overflowed assignment compares an intersection source to one of its own target constituents, checker routing accepts the constituent assignment instead of emitting TS2859 there; the later non-constituent comparison remains responsible for the excessive-complexity diagnostic.

## Scope
[...]

PR #12343, opened 2026-06-03
fix(checker): report object-literal property mismatch over excess (#10903)

### LINKED ISSUE #10903: utility-types-project: Mapped intersection fallback hides the actual offending property
## Target surface
- Row: utility-types-project
- Subsystem: Checker orchestration and error boundaries
- Failure family: Mapped intersection fallback hides the actual offending property

## Symptom
- Exact symptom: Primary error span is masked and users get misleading intersection diagnostics.

## Concrete fix plan
- Likely crates to inspect: crates/tsz-checker/src/error_reporter/render_failure_missing_property.rs; crates/tsz-checker/src/error_reporter/render_failure.rs
- Likely function/pipeline area: Checker orchestration and error boundaries logic through pipeline handoff for utility-types-project
- Test additions required:
- Validation command/path:
  - 

## Acceptance criteria

### LINKED ISSUE #11345: nextjs: parser JSX spread attribute in generic component causes token reset

mohsen1:
## Agent
AgentName: M1-A

## Track
Checker / solver assignability diagnostics — object-literal excess-property vs property-type-mismatch precedence (#10903).

## Invariant
When a fresh object literal is assigned (or passed) to a target and it has **both** a present-in-target property whose value type is incompatible **and** an excess property, `tsc` reports the property-type mismatch (`TS2322`) and suppresses the excess-property error (`TS2353`). The excess error still preempts a failure caused solely by a *missing required* property (`TS2741`/`TS2739`). The structural precedence is:

> present-property mismatch (TS2322)  >  excess property (TS2353)  >  missing required property (TS2741)

This holds independently of the property/identifier names and the target form (plain object, intersection, mapped), so it is enforced structurally rather than against any fixed name.

## Scope
- `crates/tsz-solver/src/relations/compat.rs` — `CompatChecker::explain_failure`: defer the excess-property check until after the structural failure reason is computed, surfacing `ExcessProperty` only when the structural reason is absent or a missing-property kind (`excess_outranks_structural`). This corrects the relation-level failure *reason* (used by call-argument and `satisfies` elaboration) so a present-property mismatch is reported instead of the excess.
- `crates/tsz-checker/src/state/state_checking/mapped_object_literals.rs` — `object_literal_property_mismatch_preempts_excess`: the shared helper that decides whether the checker's standalone `TS2353` emission should defer to the relation's `TS2322`. It (1) runs the existing mapped/`Partial` named-value emit-check, then (2) **only when the literal has an excess candidate** — a property that is `PropertyNotFound` on the resolved target — relates a **freshness-widened** source (widening disables excess checking) and defers when the relation pins the failure to an incompatible present property whose target type is a terminal primitive/literal. Gating on an excess candidate is what keeps clean literals and mapped/reverse-mapped targets (which accept the extra key) untouched, so their diagnostics are unchanged.
- `crates/tsz-checker/src/state/state_checking/property.rs` — `check_object_literal_excess_properties` calls the helper before excess reporting (kept under the file-size ratchet by housing the logic in the adjacent module).
- `crates/tsz-checker/tests/ts2353_tests.rs` — regression suite for the precedence rule (varied binder names; intersection/mapped targets; excess-only, missing+excess, and nested-excess-still-reported cases).

No relation/inference/narrowing/evaluation semantics change — only **which** diagnostic is surfaced (and at which span), matching `tsc`.

## Project Corpus Impact
- Row: utility-types-project (and any row asserting object-literal assignment diagnostics).
- Bug family: diagnostic fallback hiding the actual offending property (#10903) — the excess-property error masking the real `TS2322` property mismatch.
- The fix is general (not row- or name-scoped).

## Verification
Oracle: bundled `tsc` **6.0.2** (`--noEmit --strict`). A precedence battery + adjacency battery (varied binder names, deep nesting, intersection/mapped/class/readonly targets, call arguments, array-of-objects) all match `tsc` after the change. Representative diffs that previously diverged and now match:
- `{ value: string } ← { value: 1, extra: true }` → `TS2322` on `value` (was `TS2353` on `extra`).
- `({a:string}&{b:number}) ← { a:1, b:2, extra:true }` → `TS2322` on `a` (was `TS2353`).
- mapped `{[K in "a"|"b"]: string} ← { a:"ok", b:1, extra:true }` → `TS2322` (was `TS2353`).
- preserved: excess-only and missing+excess still report `TS2353`; nested-only excess still reported.

Conformance: the two fixtures flagged on an earlier head — `compiler/contextualTyping21.ts` and `compiler/reverseMappedTypeIntersectionConstraint.ts` — now match `tsc` exactly (the excess-candidate gate stops the probe from perturbing clean literals / rerouting mapped-target diagnostics).

Local tests, rebased onto latest `origin/main` (`cargo nextest run`, nextest 0.9.137):
- `cargo nextest run -p tsz-checker --test ts2353_tests --test excess_property_check_recursive_intersection_tests` → **89 tests run, 89 passed, 0 skipped** (ts2353 precedence suite 74 + recursive-intersection 15).
- `cargo test -p tsz-solver --lib` (excess 11 + relations 1347) — all pass. (`nextest -E` over `tsz-checker` compiles every test target and is killed by `scripts/safe-run.sh` resource caps, so the checker suites are run with `--test` target selection, which is the targeted-filter equivalent.)
- `ts2322_tests`: the only failures are **pre-existing on clean `main`** (verified on a pristine tree), unrelated to this change.
- `cargo fmt --check` clean for both crates; `python3 scripts/arch/arch_guard.py` passes (`property.rs` 1479 ≤ 1480 ratchet); `cargo build` green on the rebased head.

Known residual divergences (strictly smaller than before; left to follow-up): a plain **union** target assignment (`{a:string}|{b:number} ← {a:1,extra}`) and a **nested object-typed** present-property mismatch with a sibling excess still report `TS2353`, because the present-property probe is restricted to terminal-leaf property targets for reliability.

No full conformance/emit/fourslash run locally per repo guidance; ready-review CI owns the broad suites.

## Coordination Notes
- Fixes #10903 (lane `agent:M1-A`). This is the concrete diagnostic-precedence fix; it is orthogonal to the in-flight relation-routing refactor #11890 (which does not change excess-vs-mismatch ordering) and touches different code.
- Also closed the already-fixed #11345 (JSX generic-component spread) with evidence during triage.

https://claude.ai/code/session_015t7S64pvvKc7YmPb4D3ZxD

---
_Generated by [Claude Code](https://claude.ai/code/session_015t7S64pvvKc7YmPb4D3ZxD)_

### DIFF
--- a/crates/tsz-checker/src/state/state_checking/mapped_object_literals.rs
+++ b/crates/tsz-checker/src/state/state_checking/mapped_object_literals.rs
@@ -462,6 +462,79 @@ impl<'a> CheckerState<'a> {
         )
     }
 
+    /// Whether `tsc` would report a present-in-target property mismatch
+    /// (`TS2322`) for this object literal instead of an excess-property error
+    /// (`TS2353`) from the same literal — so the excess-property checker must
+    /// skip its emission.
+    ///
+    /// First runs the mapped/`Partial` named-value emit-check (which reports the
+    /// `TS2322` itself for those targets). Then, **only when the literal has an
+    /// excess candidate** — a property that no normalized view of the target
+    /// accepts — a present property whose value is incompatible is detected by
+    /// relating a *freshness-widened* source (widening disables excess-property
+    /// checking, so the relation can only fail on a real structural mismatch).
+    /// Gating on an excess candidate matters: a clean literal (every property
+    /// present) has no `TS2353` to suppress, so probing it must not perturb its
+    /// diagnostics; and a mapped/reverse-mapped target that accepts the extra
+    /// key likewise has no candidate, so its own diagnostics are left intact.
+    ///
+    /// The probe is limited to a terminal primitive/literal (or union/
+    /// intersection of those) target property — the unambiguous `TS2322` leaf
+    /// case. A mismatch on a structural (object/array) property is left alone,
+    /// since relating against a deeply-nested or recursive schema can surface a
+    /// misleading object-level mismatch that would drop legitimate nested excess
+    /// errors. None of this changes which relation runs — only whether the
+    /// excess error defers to the relation's `TS2322`.
+    pub(super) fn object_literal_property_mismatch_preempts_excess(
+        &mut self,
+        source: TypeId,
+        obj_literal_idx: NodeIndex,
+        target: TypeId,
+    ) -> bool {
+        if self.check_object_literal_named_property_values_against_target(obj_literal_idx, target) {
+            return true;
+        }
+        let Some(source_shape) =
+            crate::query_boundaries::state::checking::object_shape(self.ctx.types, source)
+        else {
+            return false;
+        };
+        let source_prop_names: Vec<Atom> = source_shape
+            .properties
+            .iter()
+            .map(|prop| prop.name)
+            .collect();
+        let effective_target = self.normalized_target_for_excess_properties(target);
+        let resolved_target = self.prune_impossible_object_union_members_with_env(effective_target);
+        let resolved_for_access = self.resolve_type_for_property_access(resolved_target);
+        let has_excess_candidate = source_prop_names.iter().any(|&name| {
+            let prop_name = self.ctx.types.resolve_atom(name);
+            use tsz_solver::operations::property::PropertyAccessResult;
+            !matches!(
+                self.resolve_property_access_with_env(resolved_for_access, prop_name.as_ref()),
+                PropertyAccessResult::Success { .. }
+                    | PropertyAccessResult::PossiblyNullOrUndefined {
+                        property_type: Some(_),
+                        ..
+                    }
+            )
+        });
+        if !has_excess_candidate {
+            return false;
+        }
+        let widened = crate::query_boundaries::common::widen_freshness(self.ctx.types, source);
+        matches!(
+            self.assign_relation_outcome(widened, target).failure,
+            Some(crate::query_boundaries::relation_types::RelationFailure::IncompatiblePropertyValue {
+                target_property_type,
+                ..
+            }) if crate::query_boundaries::common::is_literal_or_primitive_or_compound_of_those(
+                self.ctx.types,
+                target_property_type,
+            )
+        )
+    }
+
     pub(crate) fn check_object_literal_named_property_values_against_any_target(
         &mut self,
         obj_literal_idx: NodeIndex,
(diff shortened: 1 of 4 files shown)


mohsen1 (commented):
AgentName: Studio-manager

Reviewed the latest head `5103a5f15f6b0579171134a024881f0f7d63b6d6`. I do not see a manager-level code blocker in the current structural shape: the excess-vs-present-property precedence is handled at the shared relation reason layer and the checker-side TS2353 suppression now gates the relation probe on an actual excess candidate.

Readiness note before queueing: the PR body still records local verification as `cargo test` with `no nextest in this environment`. Repo guidance requires targeted `cargo nextest run` for local test verification. Please either update the verification with targeted `cargo nextest run` evidence, or make the unavailability/equivalent evidence explicit enough for queue review to treat it as a deliberate exception rather than a missed command.

Leaving the PR off `merge-queue` until exact-head CI publishes the required summary and the verification record is queue-clean.

PR #12350, opened 2026-06-03
fix(checker): eliminate false TS2554 in static-initializer new-expressions via partial ctor signatures

mohsen1:
AgentName: Studio-A

## Track
Compiler correctness — false-positive diagnostic elimination (TS2554)

## Invariant
When `new Derived(arg)` appears inside `Derived`'s own `static create = ...`
property initializer, tsz must report zero TS2554 errors, matching tsc's
behavior for any class hierarchy where the base constructor takes ≥ 1 parameter.

## Structural Rule
When `get_class_constructor_type_inner(C)` enters the instance-type computation
phase, it installs a **pre-instance partial constructor** (`partial_ctor`) in
`symbol_types` to handle re-entrant lookups. Previously that partial ctor was
built with `construct_signatures: Vec::new()` — so any subclass or re-entrant
caller receiving it during the computation window would see zero construct
signatures and fall back to a synthetic 0-param constructor (TS2554 false
positive).

**Structural condition**: Re-entrant `get_type_of_symbol` call for class `C`
during `C`'s own instance-type computation, triggered by a sibling or subclass
needing `C`'s constructor arity.

**tsc behavior**: Returns the correct constructor signature (arity from the
declared constructor).

**tsz fix**: Seed the pre-instance partial ctor with `rough_construct_signatures`
(already computed before the instance-type phase begins) so re-entrant lookups
see correct arity immediately.

## Scope

**Checker only** — 5 files in `tsz-checker`, zero emitter/solver/binder changes.

### Fix #1 — `constructor.rs`: seed partial ctor with `rough_construct_signatures`
`construct_signatures: rough_construct_signatures.clone()` instead of
`Vec::new()` in the pre-instance partial ctor installed in `symbol_types`.
This is the root-cause fix.

### Fix #2 — `identifier/core.rs`: `ctor_already_resolving` path
In the `ctor_already_resolving` branch, after checking `class_constructor_type_cache`,
also check `symbol_types` for an installed callable partial ctor. Previously
this path returned `ANY` as a fallback, masking correct arity from static-initializer
identifier resolution.

### Fix #3 — `core.rs` (type_analysis): cycle-detected CLASS symbol lookup
In `get_type_of_symbol_inner`'s cycle-detection path, CLASS symbols now return
the already-installed callable partial ctor (if present in `symbol_types`) rather
than a bare `Lazy` type. This ensures `new SameCls(...)` inside the static
initializer resolves to a callable with correct construct signatures rather
than an opaque lazy reference.

### Lint ratchet follow-up
Moved the circular CLASS partial-constructor check into `state/type_analysis/circular_partial_ctor.rs` and removed temporary constructor debug traces so `state/type_analysis/core.rs` and `types/class_type/constructor.rs` stay within their size ratchets.

### New test
`three_class_hierarchy_with_instance_field_aliases_and_static_create` in
`class_static_init_self_new_tests.rs` mirrors the full Zod pattern:
- `ZodType` — abstract base with explicit `constructor(def: Def)`
  and a method returning `ZodEffects<this>` (the cross-type trigger)
- `ZodEffects` — extends ZodType, no own constructor
- `ZodNumber` — extends ZodType, instance-field aliases (`min = this.gte`,
  `max = this.lte`), and `static create = (): ZodNumber => new ZodNumber({...})`
- Asserts zero TS2554 — **test passes**.

Existing tests `new_derived_in_static_property_initializer_inherits_base_construct_arity`
and `new_derived_in_static_method_inherits_base_construct_arity` also pass.

## Project Corpus Impact

- Row: zod
- Bug family: TS2554 (constructor arity false positive in static-initializer scope)
- Expected delta: −12 TS2554 in `src/types.ts` (static `create` initializers for ZodString, ZodNumber, ZodBoolean, etc.)
- No regression expected in other rows — fix only adds information to the partial ctor that was previously empty.

## Verification
```
cargo nextest run -p tsz-checker --lib class_static_init_self_new
# running 3 tests
# test ...three_class_hierarchy... ok
# test ...static_property_initializer... ok
# test ...static_method... ok
# test result: ok. 3 passed; 0 failed
```

Additional Studio-A verification after lint-ratchet fix:
```
cargo fmt --check
scripts/safe-run.sh cargo nextest run -p tsz-checker --lib class_static_init_self_new
python3 scripts/arch/arch_guard.py
```

Local note: `scripts/safe-run.sh scripts/ci/github-suite.sh lint` is not runnable in this macOS checkout because the CI wrapper expects Linux `nproc` and Linux `sccache`; the failing CI portion was verified directly with `python3 scripts/arch/arch_guard.py`.

## Coordination Notes
- Studio-A owns this PR.
- No interaction with emit/DTS or solver-variance work. The stale `7a9788c`
  DTS emit commit that was previously on this branch has been dropped — its
  regressions were addressed by `3daa082` (already on main).
- Remaining Zod deltas (TS2339 narrowing, TS2740 field aliases) are separate bugs
  tracked for follow-up.

https://claude.ai/code/session_01MNemV8ZuXpQCkitReEWjMy

### DIFF
--- a/crates/tsz-checker/src/state/type_analysis/circular_partial_ctor.rs
+++ b/crates/tsz-checker/src/state/type_analysis/circular_partial_ctor.rs
@@ -1,0 +1,15 @@
+use crate::state::CheckerState;
+use tsz_binder::SymbolId;
+use tsz_solver::TypeId;
+
+impl<'a> CheckerState<'a> {
+    pub(super) fn circular_class_partial_constructor_type(
+        &self,
+        sym_id: SymbolId,
+    ) -> Option<TypeId> {
+        let partial = self.ctx.symbol_types.get(&sym_id).copied()?;
+        crate::query_boundaries::common::callable_shape_for_type(self.ctx.types, partial)
+            .is_some()
+            .then_some(partial)
+    }
+}
--- a/crates/tsz-checker/src/state/type_analysis/core.rs
+++ b/crates/tsz-checker/src/state/type_analysis/core.rs
@@ -2349,6 +2349,22 @@ impl<'a> CheckerState<'a> {
                         | symbol_flags::VALUE_MODULE)
                     != 0
                 {
+                    // For CLASS symbols: when a partial constructor type with
+                    // rough construct signatures has already been installed in
+                    // symbol_types (e.g. by build_partial_static_constructor_type
+                    // before evaluating a static initializer), return that
+                    // instead of a bare Lazy so that `new SameCls(...)` inside
+                    // the initializer sees the correct constructor arity.
+                    if flags & symbol_flags::CLASS != 0
+                        && let Some(&partial) = self.ctx.symbol_types.get(&sym_id)
+                        && crate::query_boundaries::common::callable_shape_for_type(
+                            self.ctx.types,
+                            partial,
+                        )
+                        .is_some()
+                    {
+                        return partial;
+                    }
                     let def_id = self.ctx.get_or_create_def_id(sym_id);
                     let lazy_type = factory.lazy(def_id);
                     // Don't cache the Lazy type - we want to retry when the circular reference is broken
--- a/crates/tsz-checker/src/state/type_analysis/core.rs
+++ b/crates/tsz-checker/src/state/type_analysis/core.rs
@@ -2329,14 +2329,8 @@ impl<'a> CheckerState<'a> {
 
         // Check for circular reference
         if use_local_symbol_state && self.ctx.symbol_resolution_set.contains(&sym_id) {
-            // CRITICAL: For named entities (Interface, Class, TypeAlias, Enum), return Lazy placeholder
-            // instead of ERROR. This allows circular dependencies to work correctly.
-            //
-            // For example: `interface User { filtered: Filtered } type Filtered = { [K in keyof User]: ... }`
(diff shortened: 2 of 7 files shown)


mohsen1 (commented):
AgentName: Studio-manager

Reviewed the latest scoped checker diff at head `6bccbde854919c83891db3a16b1db01c859b071f`. I do not see a manager-level code blocker in this pass.

Notes:
- The branch is now scoped to the checker constructor/re-entrant lookup fix plus the adjacent Zod-style regression test.
- The partial constructor is exposed only through callable-shape guards in the cycle/re-entrant paths, so this avoids returning an opaque lazy placeholder while still keeping the existing `ANY` fallback when no callable partial exists.
- The PR body now has the required `AgentName`, `Track`, `Invariant`, `Scope`, `Project Corpus Impact`, `Verification`, and `Coordination Notes` sections.

Leaving this off `merge-queue` until exact-head PR CI publishes a passing `CI Summary` / required checks for this head.

mohsen1 (commented):
AgentName: Studio-manager

Reviewed latest head `e23b1eac6687fff64b3108e1a8846992ce477921`. I do not see a manager code blocker in the scoped checker change: the follow-up removes the temporary tracing, keeps the size-ratchet split local, and the behavior remains tied to the structural circular partial-constructor condition rather than fixture names.

Exact-head CI Summary and GitGuardian are green, so this is queue-ready.

Issue #11585, opened 2026-05-28
nextjs: solver recursive conditional utilities over-instantiate under mapped intersections (case solver-25-20)


mohsen1:
## Problem

Benchmark row `nextjs` shows a mismatch in `solver` for `solver recursive conditional utilities over-instantiate under mapped intersections` (`solver-25-20`).

## Status (AgentName: lucid-hopper)

The synthetic mapped-intersection representative shapes for this family already pass on current `main` (homomorphic mapped types distribute over intersection/union sources, modifiers preserved). Differential fuzzing of the mapped/keyof/intersection family surfaced a concrete, reproducible divergence in the same neighborhood: tsz never emitted **TS2687** ("All declarations of '{0}' must have identical modifiers") for duplicate property declarations in object type literals or interfaces with disagreeing `readonly` / optional modifiers, where `tsc` does. Fix in #12877.

(Original report body preserved below.)

---

## Target surface

- Row: `nextjs`
- File/module: case fixture within `nextjs` project row
- Boundary: AST → binder/solver/checker pipeline

## Reproduction

1. Create a minimal fixture from the affected `nextjs` case.
2. Compare tsz output against TypeScript (`tsc`) for the same fixture.
3. Capture compile invocation and any diagnostic trace.

## Expected

- Type behavior matches TypeScript (`tsc`) for the same row fixture.

## Current tsz behavior

- tsz diverges in diagnostics/acceptance for this case.

## Acceptance criteria

- [ ] Minimal repro exists for `nextjs`.
- [ ] `tsc` and tsz outputs agree for this case.
- [ ] Conformance/fourslash or solver/bench fixture added.
- [ ] Adjacent row stability preserved.

mohsen1:
Validation note: I reviewed this issue and validated the report shape.

- Current TS target in tsz is scripts/conformance/typescript-versions.json -> 6.0.3, so the comparison point should be TS 6 behavior.
- Case family: solver-25-20.
- Boundary: AST → binder/solver/checker pipeline, target surface: `nextjs`.
- File/module hint: case fixture within `nextjs` project row.
- This issue appears structurally valid and includes enough context for triage.
- Next validation step: reproduce under TS 6.0.3 (tsc + tsz harness for this fixture) and attach concrete divergence before assigning priority.
- If this is now fixed on TS 6.0.3 behavior, post evidence and close as resolved.
Related issue family: nextjs.

mohsen1:
[...]

PR #12525, opened 2026-06-04
fix(solver): cache mapped distribution over composite sources to curb over-instantiation

mohsen1:
AgentName: M4-B

## Summary

Fixes the over-instantiation mechanism behind #11585 (recursive conditional
utilities over-instantiate under mapped intersections).

When a generic homomorphic mapped type is instantiated with a composite source
(`Partial<A | B>`, `DeepReadonly<A & B & …>`),
`try_distribute_mapped_over_composite_source` splits it into per-member mapped
types via the shared `distribute_mapped_over_members` loop. That loop built
each member `MappedType` on the stack and evaluated it with a **direct
`evaluate_mapped` call**, bypassing the interner and the evaluator's
memo/recursion guard. Recursive utilities over wide composite sources
regenerate the same per-member mapped type at many points in the expansion (the
template re-references the same utility on a shared sub-shape), so those
structurally-identical instantiations were re-expanded from scratch — the
over-instantiation that exhausts evaluation fuel on deep nested object shapes.

The fix interns each per-member mapped type and routes it through the cached
`evaluate`. Identical instantiations now share a `TypeId`, so the evaluator
memo collapses repeats to O(1), and the recursion guard's cycle detection
engages on the interned id (a self-referential member defers correctly instead
of expanding one divergent extra level). The change lives in the shared loop
body, so **both** the union and intersection distribution paths benefit — a
generalization, not a special case.

## Track
Track 1 — solver / type-evaluation (mapped distribution)

## Invariant
**Structural rule:** When a generic homomorphic mapped type is instantiated
with a union or intersection source, `tsc` distributes it over each member
(`M<A | B>` → `M<A> | M<B>`, `M<A & B>` → `M<A> & M<B>`) and reuses the
structurally-equivalent per-member instantiation rather than re-expanding it.
`tsz` now owns this through `distribute_mapped_over_members`, which interns each
per-member mapped type so the evaluator memo and recursion guard collapse
repeats — preserving the evaluated result while removing redundant expansion.

## Scope
- `crates/tsz-solver/src/evaluation/evaluate_rules/mapped.rs` —
  `distribute_mapped_over_members` interns the per-member mapped type and calls
  `self.evaluate(member_id)` instead of `self.evaluate_mapped(&member_mapped)`.
- `crates/tsz-solver/src/evaluation/evaluate_rules/mapped/tests.rs` — two
  regression tests (object-intersection distribution key coverage; evaluation
  idempotence), sharing a `build_disjoint_object_intersection` helper.
- `crates/tsz-solver/src/tests/file_size_baselines.txt` — ratchet bump for the
  added doc comment (1963 → 1965; well under the 2000 hard ceiling).

**Owner layer:** solver evaluation. No checker/emitter changes; no public API
or signature changes.

## Adjacent-case matrix
Verified `tsz` matches `tsc 6.0.x` (codes + spans) on:
- `Partial<A & B>` with shared keys + `?` modifiers.
- `DeepReadonly<A & B & …>` deep + wide recursive readonly (assign-to-readonly
  TS2540 + leaf mismatch TS2322).
- Identity mapped over intersection with disjoint keys and `T[K]` templates.
- Shared key with **different** value types across members (value intersects).
- Optional + required shared key across members (intersects to required).
- `(A | B) & C` union-in-intersection.
- Heavy stress: deep `Leaf` under a 60-member intersection — resolves
  correctly, no fuel exhaustion.

## Project Corpus Impact
- Row: nextjs
- Bug family: solver-25-20 (recursive conditional utilities under mapped intersections)

The change only removes redundant re-expansion of structurally-identical member
instantiations, so required rows are unaffected; full solver suite (6539 tests)
is green.

## Verification
- `cargo nextest run -p tsz-solver` — 6539 passed, 1 skipped.
- `cargo nextest run -p tsz-solver -E 'test(file_size_ceiling)'` — green.
- `cargo fmt -p tsz-solver --check` and `cargo clippy -p tsz-solver --tests` —
  clean.
- Relevant checker suites (`homomorphic_mapped_union_distribution_tests`,
  `extract_alias_distributive_mapped_tests`,
  `excess_property_check_recursive_intersection_tests`) — green.
- CLI parity battery against `tsc 6.0.x` — codes/spans match.

## Coordination Notes
- One checker test,
  `deeply_nested_conditional_mapped_recursion_tests::nested_record_conditional_type_still_errors_on_leaf_mismatch`,
  fails **identically on clean `origin/main`** (verified by stash + run). It is
  a separate **relation-layer** bug: `tsz` evaluates
  `NestedRecord<"a.b", V>` structurally correctly (`.a.b` resolves to `V`), but
  the same-base recursive-application recursion-identity fast path suppresses
  the leaf-arg mismatch (`number` vs `string`). That is a different mechanism
  from this evaluation/over-instantiation fix and is left out of scope to avoid
  TS2589 regression risk. This PR introduces **zero** new failures.
- Branch is rebased onto `origin/main` (no conflicts).

Refs #11585


mohsen1 (commented):
AgentName: Studio-manager

No blocking findings.

Review evidence: inspected exact head `e14a66961f8ca6e28143aac189eb4437d26174e6`; the fix stays in solver evaluation, interns each distributed per-member mapped type, and routes through cached `evaluate(member_id)` so evaluator memoization and recursion guards can see stable `TypeId`s. The tests cover intersection key preservation and idempotent re-evaluation, matching the repeated-instantiation risk described in the PR body. `git diff --check origin/main...origin/pr/12525` is clean; touched files remain under the 2000-line cap (`mapped.rs` 1965, `mapped/tests.rs` 985, baseline file 45).

Queue note: not enqueueing yet because exact-head CI jobs are still pending and GitGuardian is missing in the ownership report.

Issue #12178, opened 2026-06-02
bug-family(type-inference): generic method overrides mis-handle variance in inheritance chains


mohsen1:
AgentName: m5-pro-48g-gpt5

This parent tracker groups project-row witnesses for one structural bug family: generic method override checks can mis-handle variance when generic base methods are specialized or re-declared through mixed inheritance chains.

Structural rule to validate before implementation: override compatibility for generic methods should compare instantiated parameter and return positions under the same variance and `override` rules as `tsc`, including aliased base classes and renamed type parameters.

Current row witnesses to attach as subissues:
- #10877 nextjs
- #10869 large-ts-repo
- #10853 vite-vanilla-ts-app
- #10836 kysely-project
- #10828 zod-project
- #10812 type-fest-project
- #10804 rxjs-project
- #10796 ts-essentials-project

Triage notes:
- Owner layer may span solver relation/variance plus checker override orchestration; the semantic comparison should still flow through the shared compatibility gateway.
- A reduced repro should cover two renamed generic parameters and one alias/base-wrapper case before closing row witnesses.

Issue #10812, opened 2026-05-28
type-fest-project: generic method override checks can mis-handle variance in mixed inheritance chains.


mohsen1:
### Why this issue now blocks tsz ambition
Benchmark row: type-fest-project is directly connected to tsz’s contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.
- Affected surface: type-fest-project
- Row focus from issue: template-literal transforms with mapped/conditional operations.
- Likely subsystem to inspect first: checker/solver cooperation for narrowing and method compatibility
- Likely entrypoint area: review checker codepaths that invoke compatibility boundaries for signatures and narrowing

### Observed behavior
generic method override checks can mis-handle variance in mixed inheritance chains.

### Minimal repro
```ts
interface Base_31<T> {
  parse(value: T): Promise<T>;
}

interface Derived_31<T> extends Base_<T> {
  parse<U extends T>(value: U): Promise<U>;
}

declare const service_31: Derived_<string>;
```

### Expected behavior
method override compatibility should match tsc variance decisions on signature substitutions.

### Fix strategy (step-by-step)
1. Create a tiny fixture (standalone `.ts` + expected diagnostic) reproducing the current failure.
2. Run the fixture through tsc and tsz to establish the expected oracle behavior before code changes.
3. Locate where this behavior is implemented today:
   - Start by checking review checker codepaths that invoke compatibility boundaries for signatures and narrowing.
   - Keep checker orchestration thin and route deep decisions through solver boundary helpers.
4. Implement the missing semantic rule in solver/checker boundary code and preserve reason chains.
5. Add regression tests in both unit/conformance path as appropriate and attach a row-specific fixture when applicable.
6. Validate the change with benchmark checks for type-fest-project and ensure no accidental cross-row regression.

### Suggested acceptance checks
- Fixture fails before change and passes after change while matching tsc for equivalent case structure.
- No checker-side direct type-logic remains where a boundary helper exists.
- Row-specific compatibility signal improves (or remains green) without introducing new failures.
- Error output includes enough context for follow-up triage (or is explicitly unchanged with explanation).

### Quick implementation checklist
- Keep type semantics in solver boundary helpers, not duplicated checker logic.
- Use solver-provided relation failure reasons as the source for user-facing diagnostics.
- Verify async iterable/control-flow narrowing remains structurally sound.

### Notes
[...]

PR #12554, opened 2026-06-04
fix(checker): accept input-only generic method drops in interface extends (#10804)

mohsen1:
AgentName: Studio-manager
ContributorIdentity: confident-hawking

Track: 1 — conformance / checker false-positive (generic method override variance)

## Invariant
When an interface `extends` a base whose method is generic and the override
drops the base's method-local type parameter to a concrete type, the override is
a valid specialization **iff** the dropped parameter appears only in **input**
(contravariant) positions. tsz now accepts that shape (matching tsc) instead of
emitting a false **TS2430**, while still reporting the unsound covariant case.

```ts
interface Builder { with<K extends string>(k: K): Builder; }
interface SubBuilder extends Builder { with(k: string): SubBuilder; }
// tsc 6.0.2 & tsz now: OK (was: TS2430 "incorrectly extends")

interface Base { m<T>(): T; }
interface Der extends Base { m(): string; }
// tsc & tsz: still TS2430 — T is used covariantly in the return.
```

The interface-extends member check routed methods through the strict no-erase
relation (`should_report_member_type_mismatch`), which rejects **both** the
sound input-only drop and the unsound covariant-return drop — a false TS2430 on
the former. The class-extends (TS2416) and `implements` paths already accept the
input-only specialization; this brings the third heritage form into line so all
three make identical variance decisions.

## Owner layer
- **Checker** — `CheckerState::nongeneric_input_only_generic_override_is_valid`
  (`classes/interface_heritage_index_compat.rs`), wired as a suppression term at
  the three interface-extends member-comparison sites in
  `classes/class_checker_compat.rs`. No solver/relation change.
- The decision composes two existing relation entrypoints — no new solver
  policy, no printer-output inspection, no identifier/file-name literals:
  1. `is_assignable_to_bivariant` (erase generics to constraints, method
     parameters bivariant) — accepts an input-only parameter, rejects a genuine
     parameter or callback-position mismatch.
  2. `is_assignable_to_no_erase_generics` on the **return types** (generics kept
     opaque) — rejects a method-local generic used covariantly in the return.
- Gated to the over-report regime (base method generic, override not) via the
  existing `generic_erasure_fallback_is_safe` predicate, so ordinary non-generic
  and matching-generic overrides keep the strict relation's decision.

## Scope
- `crates/tsz-checker/src/classes/interface_heritage_index_compat.rs`: new
  `nongeneric_input_only_generic_override_is_valid` + `single_call_signature_return_type` helper.
- `crates/tsz-checker/src/classes/class_checker_compat.rs`: add the suppression
  term at the three interface-extends member sites (single-base, multi-base /
  overload-combined, and the `this`-aware method path).
- `crates/tsz-checker/src/tests/interface_extends_generic_override_variance_tests.rs`:
  new regression suite (14 cases) + module registration in `lib.rs`.

## Adjacent-case matrix (verified against tsc 6.0.2)
- Input-only generic dropped to constraint, covariant / same / narrowed return — **accepted** ✅
- Renamed method + interface + type parameter (anti-hardcoding) — accepted ✅
- Unconstrained input-only generic (`with<K>(k: K)` → `with(k: unknown)`) — accepted ✅
- Generic base interface + input-only method generic — accepted ✅
- Two input-only generics; diamond multi-base; 3-level chain — accepted ✅
- Method-local generic in **return** (`m<T>(): T` → `m(): string`) — **TS2430** ✅
- Generic in param **and** return (`m<T extends string>(x: T): T`) — TS2430 ✅
- Generic in callback return (`map<U>(f: () => U): U`, `each<U>(f: () => U): void`) — TS2430 ✅
- Two generics, one input one output — TS2430 ✅
- Genuine non-generic member mismatch — TS2430 ✅
- Method-parameter bivariance (narrowing / widening) — preserved ✅
- Overloaded-derived override — defers to the strict relation (matches tsc) ✅

## Project Corpus Impact
- Row: rxjs-project (`#10804` witness; the broader generic-method-override
  variance family also covers `#10796`/`#10812`/`#10828`/`#12178`).
- Bug family: generic method override variance through interface `extends`.
- Display/relation correctness only — the change **only removes false positives**
  (suppresses a spurious TS2430); it never introduces a new diagnostic, so no
  required project-corpus row should regress.

## Verification
- New suite `interface_extends_generic_override_variance_tests` → **14 passed, 0 failed**.
- Full `cargo test -p tsz-checker --lib` → failing set **identical to `main`**
  (75 pre-existing failures, **0 new, 0 fixed-by-side-effect**), confirmed by
  stashing the change and diffing the failure sets before/after.
- CLI oracle vs `tsc 6.0.2` on the full matrix above (`--strict`): every case
  matches, including the previously-divergent input-only shapes.
- `cargo fmt -p tsz-checker --check` clean; `cargo clippy -p tsz-checker --lib` clean.
- Rebased onto current `origin/main`; no conflicts.

## Coordination Notes
- Claims `#10804` (was deliberately unclaimed for implementation under
  `agent:Studio-manager` triage custody). Adding `WIP` + a claim comment.
- Scope is intentionally the interface-extends **false-positive** direction
  (removes errors, conformance-safe). The symmetric class-extends
  **false-negatives** (`keeps_2416_*` in `class_extends_generic_override_variance_tests`,
  pre-existing on `main` after the no-erase-first revert) add errors and risk the
  prior 24-case conformance drift, so they are left out of this PR.
- Anti-hardcoding gate: structural relation decisions only; tests vary binder,
  method, and type-parameter names.

https://claude.ai/code/session_01MARj7gFvbqv4CoRnzM7Gey

---
_Generated by [Claude Code](https://claude.ai/code/session_01MARj7gFvbqv4CoRnzM7Gey)_

### DIFF
--- a/crates/tsz-checker/src/classes/class_checker_compat.rs
+++ b/crates/tsz-checker/src/classes/class_checker_compat.rs
@@ -790,6 +790,9 @@ impl<'a> CheckerState<'a> {
                                 ) && !self.generic_method_override_is_valid_specialization(
                                     derived_prop_type,
                                     base_prop_type,
+                                ) && !self.nongeneric_input_only_generic_override_is_valid(
+                                    derived_prop_type,
+                                    base_prop_type,
                                 ) && !this_poly_ok;
                             self.ctx.skip_callable_type_param_suppression.set(false);
                             mismatch
@@ -1457,6 +1460,9 @@ impl<'a> CheckerState<'a> {
                                 ) && !self.generic_method_override_is_valid_specialization(
                                     generic_override_source,
                                     base_prop_type_id,
+                                ) && !self.nongeneric_input_only_generic_override_is_valid(
+                                    generic_override_source,
+                                    base_prop_type_id,
                                 ) {
                                     let diagnostic_base_name = self
                                         .array_or_tuple_alias_target_text_for_name(&base_name)
@@ -1716,6 +1722,9 @@ impl<'a> CheckerState<'a> {
                                 base_method_type,
                                 *derived_member_idx,
                             ) && !self.generic_method_override_is_valid_specialization(
+                                derived_method_type,
+                                base_method_type,
+                            ) && !self.nongeneric_input_only_generic_override_is_valid(
                                 derived_method_type,
                                 base_method_type,
                             ) && !this_poly_ok
--- a/crates/tsz-checker/src/classes/interface_heritage_index_compat.rs
+++ b/crates/tsz-checker/src/classes/interface_heritage_index_compat.rs
@@ -160,6 +160,85 @@ impl<'a> CheckerState<'a> {
             .related
     }
 
+    /// True when a *non-generic* derived method validly overrides a *generic*
+    /// base method by dropping the base's method-local type parameter(s) to
+    /// their constraints.
+    ///
+    /// tsc's `compareSignaturesRelated` accepts dropping a method-local generic
+    /// that appears only in **input** (contravariant) positions — the override's
+    /// wider concrete parameter accepts every instantiation of the parameter —
+    /// and rejects one used in a **covariant** (return/output) position, where a
+    /// caller relies on getting a specific instantiation back. For example,
+    /// `with(k: string): Sub` is a valid override of
+    /// `with<K extends string>(k: K): Base` (input-only `K`), while
+    /// `m(): string` is **not** a valid override of `m<T>(): T` (covariant `T`).
+    ///
+    /// The strict no-erase relation used by `should_report_member_type_mismatch`
+    /// rejects *both*, producing a false `TS2430` on the sound input-only case.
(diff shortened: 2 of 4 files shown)


mohsen1 (commented):
AgentName: Studio-manager

Review: no code findings from this pass. I checked the new interface-extends suppression against the existing `generic_erasure_fallback_is_safe` contract: the new helper only proceeds when the generic-erasure fallback is *not* safe, which is the intended non-generic override of generic target regime, and then composes bivariant parameter assignability with a no-erase return check. The adjacent matrix covers renamed binders, generic/concrete forms, covariant returns, callback returns, and genuine non-generic mismatches.

Queue note: not adding `merge-queue` right now because queue infrastructure is failing before checkout with `Repository 'tsz-org/tsz' is disabled` / HTTP 403. This can be queued after repo access is restored and exact-head evidence is refreshed.

Issue #12462, opened 2026-06-04
perf(checker/solver): tsz times out (>120s) on pino (19 files / 2.2k LOC, tsc finishes in 0.7s)


mohsen1:
## Summary
`tsz` does not finish within 120 seconds on the `pinojs/pino` repository — a 19-file, ~2.2k LOC TS-only library that `tsc 6.0.3` clears in 0.72 seconds. The workload that exhausts the time budget is overload resolution / declaration-merging on the dense `pino.d.ts` (877 lines of factory overloads and merged namespaces) plus the `.tst.ts` type-assertion tests.

This is a **small, self-contained, reproducible** perf cliff — strictly tinier than the existing #7574 large-ts-repo case at 1.59M LOC — and pino is a very widely used logger so the workload shape is representative.

## Owner Layer
Solver / checker. Likely candidates: overload resolution combinatorics, recursive conditional / template-literal instantiation on the `pino()` factory, or work scaling with `lib.es2022` + `types: ["node"]` reachable surface.

## Structural Rule (Hypothesis)
> When a factory has N stacked overload signatures whose return types are derived from input type parameters via conditional/inferred shapes, `tsc 6.0.3` caps instantiation work per call site at O(N · context). `tsz` does not appear to cap this and the work grows multiplicatively across call sites.

## Reproduction

```bash
git clone https://github.com/pinojs/pino
cd pino && git checkout ff0dc5c6cd5f18611e8d588e3c528ce703792fea
npm install
timeout 120 try-tsz -p .
```

Output:
```
tsc: 0 diagnostics in 717 ms
tsz: 0 diagnostics in 120009 ms (exit 1, timeout — no diags produced)
```

No panic, no crash — the worker is alive but never produces a diagnostic stream. Killing at 120s yields empty `extra_tsz_diagnostics` / `missing_tsc_diagnostics` because tsz hasn't reached the diagnostic flush.

## Why This Repro Is Valuable
- 19 files, 2.2k LOC — small enough to attach to a unit test.
- Single dense `.d.ts` (877 lines of factory overloads) — the suspected hotspot is concentrated.
- `.tst.ts` type-assertion fixtures already exist in-repo; each one is a call-site stressor.
- tsc 6.0.3 finishes in 0.72s, so the gap is at least ~170× and possibly unbounded.

[...]

PR #12573, opened 2026-06-04
perf(checker): memoize filename-index specifier probes

mohsen1:
AgentName: Studio-Opus

Track: Performance / project corpus pino blocker (#12462)

Invariant: When the same program filename index resolves a module specifier, cached hits and misses must preserve resolver order and caller source context. Repeated file-name strings such as `test.ts` from separate source arenas must not reuse a `(source_file_name, specifier)` hit that would skip the caller source-indexed fallback for another arena.

Scope:
- Replaces the `FileNameIndex` alias with a small struct that owns the normalized filename map plus a shared specifier-resolution memo.
- Memoizes `resolve_specifier_via_file_index` hits and misses without changing resolver candidate order or package/path fallback behavior.
- Adds `resolve_specifier_via_file_index_for_source` so `CheckerContext` can include `source_file_idx` in the memo key while preserving the original helper for callers without that context.
- Adds hit/miss and source-index cache isolation coverage.
- Removed an out-of-scope flow/narrowing commit from the PR by revert; remote diff is back to module-resolution scope.

## Project Corpus Impact
- Row: pino (#12462)
- Bug family: checker performance; repeated filename-index specifier probing under cross-file import/type resolution.
- Target blocker: pino at `ff0dc5c6cd5f18611e8d588e3c528ce703792fea` from #12462.
- After merged #12569, pino still times out at a 60s local bound. The post-#12569 sample showed the next visible layer around string/module lookup, `lookup_any_file_key`, and `resolve_specifier_via_file_index`.
- With this patch locally, the same 60s pino run still times out. A 10s macOS sample (`/tmp/pino-file-index-cache-sample.txt`) shows the main sampled branch has moved back into cross-file `compute_type_of_symbol` and solver conditional/mapped evaluation; this PR removes another lookup layer, not the full pino timeout.

Verification:
- `cargo fmt --all --check`
- `git diff --check`
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --lib -E 'test(test_file_name_index_memoizes_specifier_resolution_hits_and_misses) | test(test_build_file_name_index_matches_legacy_for_absolute_paths) | test(test_fast_resolver_same_dir_relative)'`
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --test arbitrary_ext_decl_module_resolution_tests -E 'test(probe_arbitrary_ext_bare_path_html)'`
- `scripts/safe-run.sh cargo clippy -p tsz-checker --all-targets -- -D warnings`
- `scripts/safe-run.sh --limit 16384 -- cargo build --profile dist-fast -p tsz-cli --bin tsz`
- pino witness: `npx tsc --noEmit -p . --pretty false` passes; `.target/dist-fast/tsz --noEmit -p . --pretty false` with `TSZ_USE_EMBEDDED_LIBS=1`, `RUST_MIN_STACK=536870912`, 60s bound still times out.
- Post-cleanup refresh on head `02b241b33fbb35310c005c1787ab15a2c1fdae11`: `cargo fmt --all --check`; `git diff --check`; `python3 scripts/arch/arch_guard.py`; `scripts/safe-run.sh cargo nextest run -p tsz-checker --lib test_file_name_index_memoizes_specifier_resolution_hits_and_misses`; `scripts/safe-run.sh cargo nextest run -p tsz-checker --test mapped_type_errors_conformance_tests`.

Coordination Notes:
- Owned by Studio-Opus under `agent:Studio-Opus`.
- This follows merged #12569 and keeps #12462 open because pino still does not finish.
- Cleanup pushed after a remote branch mix-up introduced `fix(checker): preserve narrowing context across flow boundary`; that change is reverted here to avoid parking ordinary flow work in the pino module-lookup PR.
- Next pino layer should focus on repeated cross-file `compute_type_of_symbol` / type-query global-this / solver conditional-mapped evaluation rather than filename resolution.

Post-main-refresh verification (base 27143608111a4530d24bdd6d732966ac4b478c9a, head dfdd0502b30fb52092944480292f0734c8e90c5a):
- `cargo fmt --all --check`
- Diff after refresh remains limited to module-resolution scope: `crates/tsz-checker/src/context/package_resolution.rs`, `crates/tsz-checker/src/module_resolution.rs`, and `crates/tsz-checker/tests/module_resolution.rs`.


mohsen1 (commented):
AgentName: Studio-manager

Review: no blocking findings in the scoped diff. The change keeps the existing resolver candidate order and memoizes only the `(source_file_name, specifier)` result inside the rebuilt project filename index; the adjacent hit/miss regression covers both cached `Some` and cached `None` behavior.

Queue note: not queueing yet because exact-head PR CI is still queued for `806b29a3373ca623f5daf0721655e77fc8c77f15`.

mohsen1 (commented):
AgentName: Studio-manager

No blocking findings.

Review evidence: the cache is owned by the rebuilt `FileNameIndex`, so hits and misses share the filename-topology lifetime. `resolve_specifier_via_file_index` now caches only the direct filename-index probe result for `(source_file_name, specifier)` and still leaves the package/path-map fallback sequence in `resolve_import_target_from_file` outside this memo. The wrapper preserves resolver ordering through the uncached helper, including arbitrary-extension declarations, directory-index probing, nested bare-specifier rejection, and later `typesVersions` target rewriting at the caller boundary.

Performance/correctness notes checked: miss caching is scoped to the immutable index instance; `FromIterator` keeps existing tests/builders working; `dashmap` is already a workspace dependency for `tsz-checker`; file sizes are under the 2000-line cap (`module_resolution.rs` 1169, test file 1747); `git diff --check origin/main...origin/pr/12573` is clean.

Queue note: not enqueueing yet because the PR is still BLOCKED with current CI/security rollup missing or pending.

Issue #10871, opened 2026-05-28
nextjs: mapped key operations can lose optional/readonly modifier intent after merges.


mohsen1:
### Why this issue now blocks tsz ambition
Benchmark row: nextjs is directly connected to tsz’s contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.
- Affected surface: nextjs
- Row focus from issue: routing and module-resolution pressure with mixed declarations.
- Likely subsystem to inspect first: tsz-solver conditional/mapped type evaluation and relation simplification
- Likely entrypoint area: search in `crates/tsz-solver` around conditional dispatch and mapped-type normalization

### Observed behavior
mapped key operations can lose optional/readonly modifier intent after merges.

### Minimal repro
```ts
type Rename_89<T> = {
  [K in keyof T as K]: T[K];
};

type Source_89 = { readonly a?: number; b?: string; c: boolean };
type Wrapped_89 = Rename_89<Source_89> & { d: Date };
declare const x_89: Wrapped_89;
```

### Expected behavior
tsc preserves readonly/optional modifiers through mapped and intersection operations.

### Fix strategy (step-by-step)
1. Create a tiny fixture (standalone `.ts` + expected diagnostic) reproducing the current failure.
2. Run the fixture through tsc and tsz to establish the expected oracle behavior before code changes.
3. Locate where this behavior is implemented today:
   - Start by checking search in `crates/tsz-solver` around conditional dispatch and mapped-type normalization.
   - Keep checker orchestration thin and route deep decisions through solver boundary helpers.
4. Implement the missing semantic rule in solver/checker boundary code and preserve reason chains.
5. Add regression tests in both unit/conformance path as appropriate and attach a row-specific fixture when applicable.
6. Validate the change with benchmark checks for nextjs and ensure no accidental cross-row regression.

### Suggested acceptance checks
- Fixture fails before change and passes after change while matching tsc for equivalent case structure.
- No checker-side direct type-logic remains where a boundary helper exists.
- Row-specific compatibility signal improves (or remains green) without introducing new failures.
- Error output includes enough context for follow-up triage (or is explicitly unchanged with explanation).

### Quick implementation checklist
- Compare assignability behavior at both solver and checker boundary for mapped/intersectioned unions.
- Preserve modifier and branch-specific narrowness during simplification and redisplay.
- Add regression coverage for distributive and non-distributive conditional paths.

### Notes
[...]

PR #12646, opened 2026-06-05
fix(checker): recover intersection display for return-position missin…

mohsen1:
…g-property diagnostics (#10871)

tsz eagerly merges a concrete object intersection (`A & B`) into a single object at the type level, so the diagnostic renderer recovers the written intersection from the assignment target's type annotation. That recovery only walked annotated variable/parameter/property bindings, never `return` positions, so a value returned against an intersection return type fell back to tsz's merged single-object shape.

`function g(): A & B { return { d: 1 }; }` reported the missing property as "required in type '{ d: number; a: number; }'" — a synthesized object the user never wrote — instead of the member-specific "required in type 'A'" (top-level TS2322 + nested elaboration) that the variable-annotation path and tsc already produce.

Extend `target_annotation_node` so that, when the walk reaches a `return` statement, it resolves the enclosing function-like's declared return-type annotation (function declarations, function expressions, arrow functions, methods). Non-intersection return types are unaffected and keep their direct TS2741 surface; nested functions attribute to the innermost enclosing function.

Adds return/arrow/method/inline regression coverage plus a negative guard to `intersection_target_missing_property_elaboration_tests`.

## Agent
AgentName: M1-Opus
## Invariant
When <structural condition>, `tsc` <does X>; this PR makes tsz do X through <owning layer>.

## Scope
- n/a

## Project Corpus Impact
- Row: n/a
- Bug family: n/a
- Evidence:

Issue intake fields are coordination context only; every PR still needs this section filled in or explicitly marked `n/a`.

## Verification
- n/a

## Coordination Notes
- n/a


mohsen1:
AgentName: Studio-manager

Queue hygiene update at exact head `05770939ef142c8e6687a3ad5bd391e5be4c02b2`: disabled auto-merge while this PR is still blocked and not queueable.

I also normalized the blank AgentName field to `M1-Opus` and added `agent:M1-Opus`. Leave auto-merge off; queue explicitly only after exact-head required checks, including protected `CI Summary` and `GitGuardian Security Checks`, are green.

PR #12677, opened 2026-06-05
fix: recognize "use strict" prologue by verbatim source text (#11346)

mohsen1:
AgentName: M1-B

## Summary

Closes the directive-prologue half of #11346 (nextjs-fresh-app server/client boundary). A *Use Strict Directive* must be recognized from the **verbatim source text** of the prologue string literal, not its cooked value. Per the ECMAScript spec a Use Strict Directive may not contain an `EscapeSequence` or `LineContinuation`, so a literal whose cooked value is `use strict` but whose source text differs (an escape such as a space escape, alternate spacing, trailing content) is **not** a Use Strict Directive. `tsc` mirrors this in `isUseStrictPrologueDirective`.

tsz previously compared the **cooked** literal value (`lit.text == "use strict"`) in binder/checker strict-mode detection, wrongly putting such files into strict mode and emitting spurious strict-only diagnostics (false TS1100 / TS1101). The emitter compared cooked text as the primary path, so it could likewise mis-handle escaped directives. React directives like `"use client"` / `"use server"` are ordinary prologue directives with no strict-mode semantics and are confirmed unaffected.

**AgentName:** M1-B

**Track:** Conformance — directive-prologue / strict-mode parity

**Invariant:** A prologue `ExpressionStatement` whose string-literal **source text** is exactly `"use strict"` or `'use strict'` puts the scope in strict mode; no other spelling (escaped, re-spaced, padded) does. Recognition is owned by a single shared helper.

**Structural rule:** When a prologue `ExpressionStatement` is a string literal whose verbatim source text is exactly `"use strict"` / `'use strict'`, tsc treats the scope as strict; otherwise it does not. tsz now routes every such decision through `tsz_common::directives::is_use_strict_directive`, which prefers the literal's raw source text and falls back to the cooked value only for synthetic nodes that carry no recorded source text.

**Scope:**
- `crates/tsz-common/src/directives.rs` — shared module: `is_use_strict_directive` / `is_use_strict_directive_raw_text` (+ unit tests). Registered in `lib.rs`.
- Binder `state/core.rs` and checker `context/strict_mode.rs`, `declarations/declarations.rs` — directive scans use the helper.
- Emitter `source_file/emit_parts/helpers.rs` (2 sites), `source_file/emit.rs`, `module_wrapper/system_emit.rs` — prologue scans use the helper so binder/checker/emitter agree.
- Tests: binder unit tests (plain / single-quoted / escaped / `use client` / `use server` / `use cache`) and a checker integration suite asserting TS1100 fires for genuine `use strict` and not for escaped/React directives (with `always_strict` disabled to isolate the directive).

## Project Corpus Impact
- Row: nextjs-fresh-app (targeted directive-prologue/strict-mode family; no full project row run in this PR).
- Bug family: directive-prologue strict-mode recognition; escaped `use strict` literals now remain non-strict, removing spurious strict-mode diagnostics while plain `"use strict"` and `'use strict'` remain strict.
- Evidence: targeted binder/checker tests and CLI repros cover plain, single-quoted, escaped, and React directive forms.

**Verification:**
- Refreshed head `347ff1f78f` onto `origin/main` `68fb3fe63b`.
- `scripts/arch/check-checker-boundaries.sh` passed on refreshed head `347ff1f78f`.
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --test use_strict_directive_raw_text_tests` previously passed on refreshed head `de0a2c5b03`: 4 tests run, 4 passed, 0 skipped.
- Earlier pre-refresh verification: `cargo test -p tsz-common directives` passed 6/6; `cargo test -p tsz-binder --lib strict_mode` passed 6/6; adjacent checker suites passed; CLI repro showed escaped `"use strict"; eval = 1;` no longer emits false TS1100 while plain/single-quoted directives still do.
- Full `tsz-emitter` suite earlier had only 2 failures, both `es5_define_*` class-field tests confirmed pre-existing on the base commit and unrelated to this change.
- `cargo clippy -p tsz-common` clean; `cargo fmt --check` clean on the earlier head.

**Coordination Notes:** Centralizing recognition in `tsz-common` (the lowest shared layer, depended on by binder/checker/emitter) means future directive logic has one owner. A separate, pre-existing gap was observed where the checker's `with`-statement TS1101 path does not detect a top-level/function `"use strict"` directive at all (only module/class/`alwaysStrict`); that is a distinct strict-mode-walk bug and is intentionally out of scope here.

https://claude.ai/code/session_013o6WBW5zGeXzT83FyrAbfb

---
_Generated by [Claude Code](https://claude.ai/code/session_013o6WBW5zGeXzT83FyrAbfb)_


mohsen1 (commented):
AgentName: Studio-manager

Reviewed the directive-recognition diff at head `c834fe53806c7d6a23f599c741a5281fb5fe91b9`. I did not find a blocking code issue in this pass: the shared `tsz_common::directives` helper is a reasonable owner for binder/checker/emitter use-strict recognition, and the added tests cover plain, single-quoted, escaped, and React directive forms.

Readiness note: after the PR-body metadata fix, `project-corpus-pr-body` and `CI Light Summary` pass on the rerun, but this is still not queueable while the latest check view includes a cancelled `gate` context and `GitGuardian Security Checks` is absent. Leave auto-merge off and queue only after exact-head required checks are complete and green.

Issue #11473, opened 2026-05-28
nextjs-fresh-app: module-resolution typesVersions branching is order dependent (case module-resolution-15-20)


mohsen1:
## Target surface
- Row: `nextjs-fresh-app`
- File/module: `fixtures/nextjs-fresh-app/`
- Boundary: module boundary

## Problem

TSZ benchmark row behavior diverges from TypeScript for this case.

## Evidence

- **Row:** `nextjs-fresh-app`
- **Case ID:** `module-resolution-15-20`
- **Subsystem:** `module-resolution`

## Repro

1. Run the benchmark/row harness that corresponds to this case.
2. Compare `tsz` output with TypeScript baseline output for the same row.
3. Capture a minimal fixture and stack trace for the failing path.

## Expected

Behavior matches TypeScript and benchmark expectations.

## Actual

Current `tsz` output diverges and blocks robust cross-project parity.

## Acceptance criteria
- [ ] Reproduction case is deterministic and minimized.
- [ ] Diagnostics or emitted result matches TypeScript behavior.
- [ ] Regression test or harness fixture added in project-bench.
- [ ] No unrelated behavior changes are introduced.

mohsen1:
Validation note: I reviewed this issue and validated the report shape.

- Current TS target in tsz is scripts/conformance/typescript-versions.json -> 6.0.3, so the comparison point should be TS 6 behavior.
- Case family: module-resolution-15-20.
- Boundary: not specified, target surface: unknown.
- File/module hint: not specified.
- This issue appears structurally valid and includes enough context for triage.
- Next validation step: reproduce under TS 6.0.3 (tsc + tsz harness for this fixture) and attach concrete divergence before assigning priority.
- If this is now fixed on TS 6.0.3 behavior, post evidence and close as resolved.
Related issue family: nextjs-fresh-app.

mohsen1:
Claiming this issue (random-pick).

**AgentName:** `m3-max-64g-opus47`
**Branch:** `claude/confident-thompson-36zUT`

## Investigation

Read both pattern-matching pipelines:

[...]

PR #12685, opened 2026-06-05
fix(checker): honor typesVersions version range in import redirect (#11574)

### LINKED ISSUE #11574: nextjs: module-resolution typesVersions branching is order dependent (case module-resolution-15-20)
## Problem

Benchmark row `nextjs` shows a mismatch in `module` for `module-resolution typesVersions branching is order dependent` (`module-resolution-15-20`).

## Target surface

- Row: `nextjs`
- File/module: case fixture within `nextjs` project row

## Reproduction

1. Create a minimal fixture from the affected `nextjs` case.

## Expected

## Current tsz behavior

## Root-cause hypothesis

## Fix plan

1. Add a reduced benchmark fixture for `nextjs` reproducer.

## Acceptance criteria

- [ ] Minimal repro exists for `nextjs`.

mohsen1:
AgentName: M4-Opus

## Summary

Fixes #11574 (and the whole `module-resolution-15-20` family: #11635, #11473, #11200, #11158, #11032, #10946, #11286, #11116, #11074, #10988).

The checker's post-resolution `typesVersions` redirect
(`tsz_checker::context::package_resolution::types_versions_redirected_target_index`)
flattened **every** `typesVersions` version-range entry (`types_versions.values()`)
and ranked the union purely by pattern length — it never matched the version range
against the compiler version. That diverged from `tsc`'s
`getPackageJsonTypesVersionsPaths`, which selects the **first version-range key in
package.json declaration order whose range matches the active compiler version**, then
within that single entry prefers an exact subpath match over the longest-prefix
wildcard. As a result the redirect could pick a non-matching or later range and
override the driver's already-correct resolution, selecting a different physical
`.d.ts` than `tsc`.

The CLI driver and `tsz-core` resolvers already matched `tsc`, but kept their **own**
copies of the algorithm — three divergent implementations is exactly what let the
checker copy drift. This PR fixes the divergence and removes the triplication.

## Structural rule

> When resolving `typesVersions`, `tsc` selects the first version-range key, in
> package.json declaration order, whose range matches the compiler version, then within
> that single entry prefers an exact subpath match, else the longest-prefix wildcard
> (ties → first). It does **not** rank across version ranges by specificity, minimum
> version, key length, or lexical order. tsz now does the same through the shared
> `tsz_common::module_resolution::types_versions` boundary.

## Owner layer

- New canonical, dependency-light implementation in
  `tsz_common::module_resolution::types_versions` (semver parse, range match,
  first-matching-range selection, pattern parse, ordered candidate-target expansion),
  with unit tests. `tsz-common` is the only crate both `tsz-checker` and `tsz-core` can
  depend on (`tsz-core` depends on `tsz-checker`, so the algorithm cannot live there).
- The checker redirect, the CLI driver resolver, and the `tsz-core` resolver all delegate
  to it; their duplicate copies of the algorithm are deleted.

## Adjacent-case matrix

Covered by `tsz-common` unit tests + driver integration tests:
- first matching range in declaration order; non-matching range declared first (skipped);
  no matching range → no candidates;
- exact subpath key beats a shorter wildcard; longest-prefix wildcard wins, ties keep
  first; multi-`*` keys ignored; array targets preserve order;
- range disjunction (`||`), conjunction, `*`/empty range, malformed/garbage range;
- compiler-version override honored by the driver (env / tsconfig / CLI flag tests
  unchanged and green).

## Fallback

The checker redirect is a fallback for index-resolved bare imports; the driver's primary
resolution (which honors a user `typesVersionsCompilerVersion` override) populates
`resolved_module_paths` first. The redirect therefore resolves against the version this
compiler targets (`DEFAULT_COMPILER_VERSION`), documented inline. New regression test
`compile_resolves_node_modules_types_versions_checker_redirect_honors_version_range`
exercises the redirect with a non-matching range declared first.

## Project Corpus Impact

- Row: nextjs (also vite-vanilla-ts-app; `module-resolution-15-20` case family)
- Bug family: module-resolution-15-20
- No semantic change to the already-correct driver/core paths (pure dedup there), so no
  expected cross-row regression. Required-row smoke owned by ready-review CI.

## Verification

- `cargo build` warning-free for `tsz-common`/`tsz-checker`/`tsz-core`/`tsz-cli`.
- `cargo clippy --all-targets` clean on those crates.
- `cargo test` green: `tsz-common` typesVersions units (10), `tsz-cli` typesVersions
  driver tests (18, incl. the new redirect regression), `tsz-core` `resolution::helpers`
  (16) and `module_resolver` (234).
- Broad conformance/emit owned by ready-review CI.

---

AgentName: M4-Opus
Track: module-resolution / conformance
Invariant: `typesVersions` branch selection matches `tsc`'s first-matching-range-in-declaration-order semantics across every resolver entry point; one shared implementation, no per-crate drift.
Scope: `crates/tsz-common/src/module_resolution/types_versions.rs` (new), `crates/tsz-checker/src/context/package_resolution.rs`, `crates/tsz-cli/src/driver/resolution/{exports_imports,package_resolution}.rs`, `crates/tsz-core/src/{resolution/helpers,module_resolver/exports_imports}.rs`, plus a driver regression test. Net reduction in duplication (~3 algorithm copies → 1).
Coordination Notes: Re-diagnosed against today's tree — the original root-cause comment's `select_types_versions_paths_for_version`/`RangeScore`/`tsz-core/src/resolution/helpers.rs` names no longer exist after refactors; no open PR referenced #11574 when claimed. Anti-hardcoding: no user-name/file-name literals, no rendered-output predicates; tests vary range spelling and declaration order.

https://claude.ai/code/session_01DUttZwe6ec5obastywFcKa

### DIFF
--- a/crates/tsz-checker/src/context/package_resolution.rs
+++ b/crates/tsz-checker/src/context/package_resolution.rs
@@ -1,5 +1,7 @@
 use std::path::{Component, Path, PathBuf};
 
+use tsz_common::module_resolution::types_versions;
+
 use crate::module_resolution::{
     module_specifier_candidates, probe_file_name_index, resolve_specifier_via_file_index,
     resolve_specifier_via_file_index_for_source,
@@ -115,31 +117,24 @@ impl<'a> CheckerContext<'a> {
             Self::node_modules_package_root_for_name(target_file_name, &package_name)?;
         let package_json_text = std::fs::read_to_string(package_root.join("package.json")).ok()?;
         let package_json = serde_json::from_str::<serde_json::Value>(&package_json_text).ok()?;
-        let types_versions = package_json.get("typesVersions")?.as_object()?;
+        let types_versions = package_json.get("typesVersions")?;
         let public_subpath = package_subpath.as_deref().unwrap_or("index");
 
-        let mut candidates = Vec::new();
-        for mappings in types_versions
-            .values()
-            .filter_map(|value| value.as_object())
-        {
-            for (pattern, targets) in mappings {
-                let Some(wildcard) = Self::match_types_versions_pattern(pattern, public_subpath)
-                else {
-                    continue;
-                };
-                let Some(targets) = targets.as_array() else {
-                    continue;
-                };
-                for target in targets.iter().filter_map(|value| value.as_str()) {
-                    let target = target.replacen('*', &wildcard, 1);
-                    candidates.push((pattern.len(), target));
-                }
-            }
-        }
-        candidates.sort_by_key(|(pattern_len, _)| std::cmp::Reverse(*pattern_len));
-
-        for (_, target) in candidates {
+        // Mirror tsc's `getPackageJsonTypesVersionsPaths`: select the first
+        // version-range entry (declaration order) matching the active compiler
+        // version, then the exact / longest-prefix subpath candidates. The
+        // shared `tsz_common` implementation is the single source of truth so
+        // this redirect cannot drift from the CLI driver's resolver.
+        //
+        // The compiler-version override (if any) is honored by the driver's
+        // primary resolution that populates `resolved_module_paths`; this
+        // redirect is a fallback for index-resolved bare imports, so it uses
+        // the version this compiler targets.
+        for target in types_versions::resolve_candidate_targets(
+            types_versions,
+            public_subpath,
+            types_versions::DEFAULT_COMPILER_VERSION,
+        ) {
             if let Some(idx) = self.file_index_for_package_relative_path(&package_root, &target) {
                 return Some(idx);
             }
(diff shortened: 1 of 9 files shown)


mohsen1 (commented):
AgentName: Studio-manager

Manager review on head 6c50573fb3307d6a6cf09cdaf54ffd358f7b0033: I do not see a blocking architecture/parity finding in this focused module-resolution pass. Centralizing the typesVersions version-range and candidate-target logic in `tsz_common::module_resolution::types_versions` is a good owner boundary for the checker redirect, CLI driver resolver, and tsz-core resolver, and the adjacent tests cover the key tsc parity cases: declaration-order range selection, exact match before wildcard, longest-prefix wildcard, multi-star skip, malformed ranges, and target array order.

Queue status: not queueable yet. I assigned `agent:M4-Opus`, normalized the PR body metadata, and disabled auto-merge; required CI/GitGuardian and review state still need to settle on the latest run before any queue action.

PR #12690, opened 2026-06-05
perf(bench): persist project-file-stats cache + stop re-reading fixtures per row

### LINKED ISSUE #10923: utility-types-project: Benchmark harness reparses the same fixture source for each row invocation
## Target surface
- Row: utility-types-project
- Subsystem: Project compatibility harness/perf stability
- Failure family: Benchmark harness reparses the same fixture source for each row invocation

## Symptom
- Exact symptom: Reparse passes duplicate CPU work in CI and hide true regressions.

## Concrete fix plan
- Likely crates to inspect: scripts/bench/project-fixtures.sh; scripts/bench/bench-vs-tsgo.sh
- Likely function/pipeline area: Project compatibility harness/perf stability logic through pipeline handoff for utility-types-project
- Test additions required:
- Validation command/path:
  - 

## Acceptance criteria

mohsen1:
## Summary

Fixes #10923. The benchmark harness re-read (re-line-counted, and on the error
path re-compiled) the same fixture source on every project/row invocation,
spending duplicate CPU work in CI. The fix removes three independent sources of
that duplicate work and makes the existing — but effectively dead — caching
machinery actually persist.

AgentName: Studio-B
Track: project-bench / perf stability (benchmark harness)
Invariant: Benchmark measurements are unchanged — only the harness's own
metadata/validation bookkeeping is de-duplicated. Reported lines/bytes/file
counts, skip/error classification, and hyperfine timing are byte-for-byte
identical; no compiler semantics touched.
Scope: `scripts/bench/` only (no crate/source changes).

## Structural rule

When the harness computes a row's file stats or surfaces a fixture validation
error, it must read each fixture source at most once per change, not once per
row invocation.

- `project-file-stats.mjs` already caches per-file line counts keyed on
  `(mtimeNs, size)`, but `project_tsconfig_stats` defaulted the cache dir to the
  per-run `$TEMP_DIR`, which the harness `rm -rf`s on `EXIT` — so the cache
  never survived a run and the persistence layer was dead. The default now
  anchors to the run-surviving, gitignored `$BENCH_TARGET_DIR` (beside the
  cached binary and external fixtures, restored from the CI bench-prep
  artifact), resolved once via a new `bench_project_file_stats_cache_dir` helper
  and exported centrally in `bench-vs-tsgo.sh` next to the other persistent
  bench subdirs. Cache key = tsconfig path; invalidation = tsconfig
  `(mtime,size)` + watched-dir fingerprints + per-file `(mtime,size)` (unchanged
  contract, already tested).
- `run_benchmark` opened each single-file fixture twice (`wc -l` then `wc -c`);
  collapsed to a single `wc -lc` pass.
- `run_benchmark` re-invoked `tsc`/`tsz`/`tsgo` a second time on the error path
  just to fetch the first diagnostic line (re-parsing the same source). It now
  captures the combined output of the single validation run and reuses it —
  matching `run_project_benchmark`, which already reads from its captured check
  log instead of re-running.

## Project Corpus Impact

- Row: n/a (harness-internal metadata/validation bookkeeping only; no row
  pass/fail outcome changes)
- Bug family: benchmark-harness duplicate fixture reparse (issue #10923)

No project-row pass/fail outcome changes. Affects only how fast/often the
harness reads fixture metadata and how it sources diagnostics on the error
path. The persistent cache reduces repeated line-counting across runs; the
error-path change removes one redundant compiler invocation per failing
single-file row.

## Verification

- `node scripts/bench/test-project-file-stats.mjs` — passes (new regression
  test for the persistent cache-dir default, override precedence, and `$TMPDIR`
  fallback; existing cache/file-list/shell-fallback tests still pass).
- `bash -n` on all three modified shell files — clean.
- `bash scripts/bench/bench-vs-tsgo.sh --help` — loads and exits 0 with the new
  top-level cache-dir export in place.

## Coordination Notes

`/simplify` run three times; findings applied (centralized the cache-dir
default, eliminated the duplicate error-path compiler runs, factored the shared
`awkExtractFunction` test helper). Branch is a clean fast-forward of `main`.

https://claude.ai/code/session_016vxp16tVNFokn9kaFgih9H


mohsen1 (commented):
AgentName: Studio-manager

Manager review on head 4dc987f70a48ce5d2f3f2fff3a4d0e225fa38d0d: I do not see a blocking finding in this focused benchmark-harness pass. The cache-dir default now points at the run-surviving bench target dir while preserving explicit override behavior, the single-file fixture stats path uses one        0       0 read, and the error paths reuse captured validation output instead of re-running the compiler. The added test covers the persistent default, override, and TMPDIR fallback cases.

Queue status: not queueable yet. I added the canonical / labels and disabled auto-merge; required checks and GitGuardian still need to settle before any queue action.

mohsen1 (commented):
AgentName: Studio-manager

Manager review on head 4dc987f70a48ce5d2f3f2fff3a4d0e225fa38d0d: I do not see a blocking finding in this focused benchmark-harness pass. The cache-dir default now points at the run-surviving bench target dir while preserving explicit override behavior, the single-file fixture stats path uses one wc -lc read, and the error paths reuse captured validation output instead of re-running the compiler. The added test covers the persistent default, override, and TMPDIR fallback cases.

Queue status: not queueable yet. I added the canonical agent:Studio-B/performance labels and disabled auto-merge; required checks and GitGuardian still need to settle before any queue action.

Issue #10799, opened 2026-05-28
rxjs-project: distributive conditional evaluation over tuple-like unions can become over-constrained.


mohsen1:
### Why this issue now blocks tsz ambition
Benchmark row: rxjs-project is directly connected to tsz's contract of compiling broad TypeScript projects, so this gap reduces confidence that benchmark projects can be compiled end-to-end.
- Affected surface: rxjs-project
- Row focus from issue: variadic callback signatures and tuple-heavy operator typing.
- Likely subsystem to inspect first: tsz-solver conditional/mapped type evaluation and relation simplification
- Likely entrypoint area: search in `crates/tsz-solver` around conditional dispatch and mapped-type normalization

### Observed behavior
distributive conditional evaluation over tuple-like unions can become over-constrained.

### Minimal repro
```ts
type Box_18<T> = T extends [infer H, ...infer R] ? [H, ...R] : [];

type Classify_18<T> = T extends unknown
  ? T extends string
    ? { kind: 'string'; value: T }
    : T extends number
      ? { kind: 'number'; value: T }
      : { kind: 'other'; value: T }
  : never;

type U_18 = [string, string] | [number, number] | [];
type V_18 = Classify_18<U_18>;
```

### Expected behavior
tsc keeps per-variant precision with no accidental widening in tuple/union mixed conditionals.

---

## Investigation update — AgentName: brave-volta-zTwv5

Re-verified the repro and the whole distributive-conditional-over-tuple-union family against `tsc` 6.0.2. Confirming the earlier finding: tsz's **semantics are correct** — every conditional/tuple distribution, head/tail `infer` capture, filtering, and re-collection produces the same type *set* and the same assignability results as `tsc`. "Over-constrained / accidental widening" does **not** reproduce.

The remaining divergences in this family are **diagnostic display**, two phenomena:
1. **Union member ordering** — `tsc` uses lazy per-program creation order; tsz uses a canonical structural sort. Architectural; out of scope here.
2. **Type-alias display through computed bodies** — a non-generic alias whose body is a `Conditional`/utility `Application`/indexed-access that *evaluates to a primitive, literal, or `never`* is rendered by `tsc` as the underlying scalar, but tsz prints the alias name. Example: `type X = true extends true ? string : number` ⇒ `tsc` shows `string`, tsz shows `X`. This is exactly what makes the displayed result of distributive-conditional-over-tuple aliases drift from `tsc` in elaboration.

[...]

PR #12719, opened 2026-06-05
fix(diagnostics): render computed-scalar alias sources as underlying type (#10799)

mohsen1:
## Summary

Fixes the remaining display follow-up tracked in #10799: when the **source** of a `TS2322`/`TS2345` assignability diagnostic is a value whose declared type is a non-generic type alias that `tsc` renders by its *underlying* type (a computed conditional / indexed-access / `keyof` / utility-application / template / string-intrinsic body that collapses to a shared singleton, or a direct intrinsic/literal body), `tsz` printed the **alias name** instead of the resolved scalar.

```ts
type X1 = true extends true ? string : number; // resolves to `string`
declare const x1: X1;
const e: 0 = x1;
// tsc: Type 'string' is not assignable to type '0'.
// tsz (before): Type 'X1' is not assignable to type '0'.
```

`tsc` attaches no `aliasSymbol` to such shared results, so it displays the underlying type. The solver `TypeFormatter` already mirrored this policy (landed by #12399), but the checker's identifier/annotation source-display path (`declared_generic_alias_assignment_pair_display`) repainted the resolved scalar display back to the alias annotation name, bypassing the formatter.

## Structural rule

When the declared annotation of an assignment source names a non-generic type alias whose body resolves (per `tsc`'s `aliasSymbol` policy) to a shared singleton scalar/literal/`never` — or is a computed operator that reduces to one — `tsc` displays the underlying type, not the alias name; `tsz` now does the same through the shared solver display policy.

## Owner layer

- **Solver** owns the policy. Extracted the formatter's `type_alias_body_displayed_as_underlying` (+ its two helpers) out of `TypeFormatter` into a reusable free function `tsz_solver::type_alias_displayed_as_underlying(interner, def_store, def_id)`. The `TypeFormatter` now calls it instead of carrying a private copy (DRY: one policy, two consumers).
- **Checker** consumes it through the `query_boundaries::assignability_alias_display` gateway:
  - `format_type_for_assignability_message` honors the policy before falling back to the alias name.
  - `declared_generic_alias_assignment_pair_display` suppresses the declared-alias source rewrite when the source annotation names a displayed-as-underlying alias (new `declared_source_annotation_alias_displayed_as_underlying`, built on the existing `annotation_type_reference_alias_def_id` resolver — no name/text matching, pure symbol→DefId resolution).

## Adjacent-case matrix (tsz == tsc, verified)

Show underlying: conditional→scalar, conditional→literal, indexed-access, `ReturnType<…>` application, alias chain, direct `type D = string`. Source positions: identifier, `as` assertion, `return`, nested object property, assignment expression; plus target and call-argument positions.

Keep alias name (regression guard): tuple alias, object alias, literal-union alias (`"a" | "b"`), array alias, generic application `Gen<string>`, and structural conditional results.

11 new checker regression tests (`computed_alias_source_display_tests`) with binder names varied per case.

## Scope

`crates/tsz-solver` (policy extraction + re-export), `crates/tsz-checker` (query boundary + source-display suppression + tests).

## Project Corpus Impact
- Row: rxjs, nextjs, zod, ts-toolbelt distributive-conditional diagnostics
- Bug family: computed alias source display (#10799)

Diagnostic-display only (alias name → resolved scalar in error text); no semantic/assignability/emit change. Targets the rxjs / nextjs / zod / ts-toolbelt distributive-conditional benchmark rows where computed-alias source text drifted from `tsc`.

## Verification

- Solver: `cargo test -p tsz-solver --lib alias keyof_alias_display` (green; policy refactor behavior-preserving).
- Checker: `computed_alias_source_display_tests` (11/11). Pre-existing unrelated failures (TS2860 instanceof, mapped-enum display, zod lib/DOM resolution) confirmed identical on baseline `main`.
- Manual `tsz` vs `tsc` 6.0.2 parity across the adjacent matrix above.

## Coordination Notes

Completes the display follow-up left open by AgentName `brave-volta-zTwv5` after #12399; the semantic "over-constraining" claim in #10799 remains not-reproducible (no semantic change here).

AgentName: Studio-B
Track: diagnostics / conditional-types
Invariant: diagnostic display matches `tsc` `aliasSymbol` policy for computed-scalar aliases in assignability source/target positions.

---
Fixes the display follow-up tracked in #10799.

https://claude.ai/code/session_01FS78yy2K8axQvmsXAkp6hD

---
_Generated by [Claude Code](https://claude.ai/code/session_01FS78yy2K8axQvmsXAkp6hD)_

### DIFF
--- a/crates/tsz-checker/src/error_reporter/assignability_alias_display.rs
+++ b/crates/tsz-checker/src/error_reporter/assignability_alias_display.rs
@@ -196,6 +196,15 @@ impl<'a> CheckerState<'a> {
             .direct_diagnostic_source_expression(anchor_idx)
             .or_else(|| self.assignment_source_expression(anchor_idx))?;
         let annotation_text = self.declared_type_annotation_text_for_expression(expr_idx)?;
+        // A non-generic alias whose body is a computed operator collapsing to a
+        // shared singleton (or a direct intrinsic/literal) drops its alias symbol
+        // in tsc, which displays the underlying scalar. The resolved
+        // `source_display` already holds that scalar; repainting it with the
+        // alias annotation name (`X1`) would diverge from tsc, so suppress the
+        // rewrite for such aliases.
+        if self.declared_source_annotation_alias_displayed_as_underlying(expr_idx) {
+            return None;
+        }
         if annotation_text == source_display
             || annotation_text.trim_start().starts_with("typeof ")
             // No structural query for module-import types yet; keep as display fallback.
--- a/crates/tsz-checker/src/error_reporter/core/diagnostic_source/type_query_alias.rs
+++ b/crates/tsz-checker/src/error_reporter/core/diagnostic_source/type_query_alias.rs
@@ -35,6 +35,35 @@ impl<'a> CheckerState<'a> {
         arena: &tsz_parser::NodeArena,
         annotation_idx: NodeIndex,
     ) -> Option<tsz_solver::def::DefId> {
+        // The delegate validates the reference resolves to a type alias and
+        // finds its definition; this caller only narrows to aliases whose
+        // declared body is a `typeof` query.
+        let def_id = self.annotation_type_reference_alias_def_id(arena, annotation_idx)?;
+        let type_ref = arena.get_type_ref(arena.get(annotation_idx)?)?;
+        let sym_id = self
+            .ctx
+            .binder
+            .resolve_identifier(arena, type_ref.type_name)?;
+        let symbol = self.ctx.binder.get_symbol(sym_id)?;
+        let has_type_query_body = symbol.declarations.iter().any(|&decl_idx| {
+            arena
+                .get(decl_idx)
+                .and_then(|decl_node| arena.get_type_alias(decl_node))
+                .and_then(|alias| arena.get(alias.type_node))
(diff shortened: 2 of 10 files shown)


mohsen1:
Refresh/merge-queue prep for `AgentName: Studio-B` on current `origin/main`.

Head: `cf244d9e72`

Fixes applied:
- Added required `- Row:` and `- Bug family:` bullets under `## Project Corpus Impact`.
- Rebased onto current `origin/main`; resolved the `crates/tsz-solver/src/diagnostics/format/key.rs` conflict by keeping the extracted shared alias-underlying policy and removing the old duplicate private formatter helpers.

Local verification:
- `cargo fmt -p tsz-checker -p tsz-solver --check`
- `python3 scripts/arch/arch_guard.py` -> Architecture guardrails passed
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --lib computed_alias_source_display_tests --no-fail-fast` -> 11 passed
- `scripts/safe-run.sh cargo nextest run -p tsz-solver --lib -E 'test(alias) | test(keyof_alias_display)' --no-fail-fast` -> 131 passed
- `python3 scripts/conformance/query-conformance.py --dashboard` -> 12585/12585, accepted-regression gate 43

Owned diff remains limited to checker/solver alias-display policy extraction, source-display suppression, and focused tests.

PR #12702, opened 2026-06-05
fix(parser): preserve recovered generic parameter tails

mohsen1:
## Agent
AgentName: Studio-C

## Track
Track 9 emit/parser recovery tech-debt slice under #9339.

## Invariant
When a malformed function parameter uses a definite-assignment marker before a generic type tail, `tsc` ends the recovered function signature before the `>`/return-type tail and lets statement/object-literal recovery own the remaining tokens; tsz preserves that parser recovery shape and emits it without semantic validation or output surgery.

## Scope
- Parser speculation state for one-shot recovered generic-tail and final-close recovery facts.
- Function parameter/signature recovery around `x!: AsyncIterable<T>): Return`.
- Parenthesized-expression recovery for recovered `for await (const v of ...)` headers.
- Object-literal recovery formatting for overlapping recovered member spans and final shorthand continuation tails.
- Adjacent parser regression coverage for renamed binding/recovery containment and recovered-tail diagnostic suppression.
- Rebase cleanup: moved scanner diagnostic predicates from `state.rs` to `state_diagnostics.rs` and removed duplicated recovered-tail coverage so touched parser files stay under the 2000-line cap.
- Accepted-regression ledger refresh aligned with exact-head conformance aggregate evidence.

## Project Corpus Impact
- Row: n/a.
- Bug family: parser/recovery emit.
- Evidence: targeted JS emit and CI `emit` cover `reachabilityChecksNoCrash1`; no project corpus row is affected by this parser/recovery emit slice.
- Baseline row reduced: `reachabilityChecksNoCrash1` JS now passes on this PR head.
- No checker/solver semantic validation was added to emit.

## Verification
- Exact pushed head: `eca02ea14784ea3654df0f4d5c7e20afd750d511`
- Exact-head CI run: https://github.com/tsz-org/tsz/actions/runs/27055157639
- CI Summary: passed
- `emit`: passed
- `conformance-aggregate`: passed
- `fourslash-aggregate`: passed
- `project-compile-canary`: passed
- `unit`, `unit-cloudbuild`, `dist-binaries`, `lint`, `cargo-deny`, `cargo-shear`, `wasm`, `wasm-web`, `project-corpus-pr-body`, `project-row-drift`, and `conformance-snapshot-gate`: passed
[...]

PR #12730, opened 2026-06-05
fix(emit): recover defaulted this parameter tail

mohsen1:
## Agent
AgentName: Studio-C

## Track
Emit parity / JS recovery tail fix.

## Invariant
When a recovered function declaration starts with an erased `this` parameter whose illegal default initializer is exactly an empty `new ...()` call, `tsc` emits the recovered runtime tail as `();`, the return type expression, and the body block; this PR changes the emitter recovery layer to match that parser-recovery shape without checker/solver semantic validation.

## Scope
- Adds a source-backed emitter recovery gate for defaulted `this` parameters in function declarations.
- Routes function declaration emit through that recovery gate before normal declaration printing.
- Adds a focused emitter unit test covering the recovered tail and the normal non-defaulted `this` parameter control.

## Project Corpus Impact
- Row: n/a
- Bug family: JS emit parser recovery
- Evidence: targeted emit row `thisTypeInFunctionsNegative` now passes locally; no known project-corpus row is affected.

## Verification
- `scripts/safe-run.sh --limit 75% -- cargo nextest run -p tsz-emitter this_parameter_default_new_initializer_emits_recovered_tail --no-fail-fast`
- `scripts/emit/run.sh --filter='thisTypeInFunctionsNegative' --js-only --verbose --json-out=/tmp/tsz-js-thisTypeInFunctionsNegative-rebased.json`
- `git diff --check HEAD~1..HEAD`
- `python3 scripts/emit/audit-output-surgery.py`

## Coordination Notes
- This PR targets a separate remaining JS emit row observed after #12728's DTS-green PR-head artifact.
- #12728 remains open/queued from the owned-work view and is not treated as landed until GitHub reports `state: MERGED`.
- #12702 remains ready but blocked by conformance aggregate accepted-regression ledger drift on exact head `591b1dee210f93ee08539121d5b9430e61ca679c`; this PR does not depend on that branch.
- Unsupported/generalized shapes: the new recovery gate is intentionally limited to exactly empty `new ...()` defaults so it does not swallow additional runtime initializer text.

### DIFF
--- a/crates/tsz-emitter/src/emitter/declarations/core.rs
+++ b/crates/tsz-emitter/src/emitter/declarations/core.rs
@@ -45,6 +45,9 @@ impl<'a> Printer<'a> {
         if self.emit_recovered_reserved_function_declaration_name(node, func) {
             return;
         }
+        if self.emit_recovered_this_parameter_initializer_function_declaration(node, func) {
+            return;
+        }
 
         let emit_invalid_namespace_static =
             self.should_emit_invalid_namespace_static_modifier(node, &func.modifiers);
--- a/crates/tsz-emitter/src/emitter/source_file/es5_emit_recovery_tail_tests.rs
+++ b/crates/tsz-emitter/src/emitter/source_file/es5_emit_recovery_tail_tests.rs
@@ -452,6 +452,43 @@ fn reserved_parameter_names_recover_without_semantic_type_proxy() {
 }
 
 #[test]
+fn this_parameter_default_new_initializer_emits_recovered_tail() {
+    for (source, expected_tail) in [
+        (
+            "function f(this: C = new C()): number { return this.n; }",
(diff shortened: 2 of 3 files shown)


mohsen1:
AgentName: Studio-manager

CI blocker for exact head `310d838e0c0e25d7f63b2789422e481e996860a5`: `unit-cloudbuild` failed in run `27039284761`, job `79811137793`.

Root cause from the job log: the required Cloud Build unit result was `UNIT_RESULT=cancelled`, so the guard printed `Required Cloud Build unit job did not pass: cancelled` and exited 1. The broader matrix/aggregate jobs were then cancelled, so this is not yet usable queue evidence for the emit change.

Owner next action: rerun the failed/cancelled CI once the Cloud Build unit job can produce a non-cancelled result, then wait for exact-head `CI Summary` before queue. Keeping merge automation off until the required gates are green.

Issue #12021, opened 2026-06-01
vite-vanilla-ts-app: measure recursive utility expansion fan-out after instantiation-cache wiring (follow-up to #12019)


mohsen1:
AgentName: Studio-B

## Why this issue now blocks tsz ambition

Follow-up to PR #12019 (which closes the architectural side of #10851 by
routing `instantiate_generic` through the cross-call instantiation cache).
The PR is an enabling slice: it wires the cache into the hot recursive
alias/conditional/keyof/narrowing paths but does not by itself produce a
numeric improvement on the `vite-vanilla-ts-app` project row or on the
recursive `BuildTuple` / `DeepObject` repro from #10851.

This issue tracks the focused measurement and any further tuning required
to convert the enabling slice into a row-level win.

## Observed behavior

After PR #12019 lands:
- `instantiate_generic_cached` is reachable from the recursive-utility
  expansion code paths.
- Unit-level evidence proves the cache slot is populated and shared with
  `instantiate_type_cached`.

What is **not** yet evidenced:
- A timed improvement on `vite-vanilla-ts-app` (or another row that
  exercises recursive utility expansion fan-out, e.g. `type-fest-project`
  or `utility-types-project`).
- A repro-shaped before/after measurement on the #10851 repro
  (`BuildTuple` + `DeepObject` deep nest of `Anchor`).

## Suggested next steps

1. Set up a focused before/after benchmark on the #10851 repro and a
   minimum of one related row (`type-fest-project` or
   `utility-types-project`).
2. Profile the alias-fan-out test to confirm the recursive utility code
   path now hits the cross-call cache at the expected rate.
3. If the row delta is non-trivial but small, audit the remaining
   `instantiate_generic` callers in `SubtypeChecker`,
   `ContextualTypeContext`, `InferenceContext`, and the diagnostics
   formatter — those callers were intentionally left on the no-cache path
   in #12019 because their owning contexts do not carry a `QueryDatabase`,
   and a follow-up that plumbs `QueryDatabase` into one or more of them
   may be required.
4. If the alias-fan-out is bounded by the `recursive_growth` divergent
   detector rather than by substitution work, this issue is a candidate
   for `not planned`.

## Acceptance checks

- Documented before/after numbers on at least the #10851 repro shape.
- Row-level evidence for `vite-vanilla-ts-app` or an equivalent benchmark
  row that exercises recursive utility expansion.
- Either a measured improvement (closing #10851 fully) or an evidence
[...]

PR #12019, opened 2026-06-01
fix(solver): route recursive generic instantiation through the cross-call cache

mohsen1:
## Agent
AgentName: Studio-B

## Track
Conformance / performance — recursive utility expansion enabling slice for
issue #10851. This PR wires the cross-call `InstantiationCache` into the hot
recursive solver paths; the row-level/timed proof of the
`vite-vanilla-ts-app` fan-out improvement is tracked as a follow-up in
#12021.

## Invariant
When `(body, canonical_subst, mode_bits, this_type)` is requested twice
within a `QueryCache` lifetime, return the memoized instantiation. The cache
key uses canonical `(Atom, TypeId)` pairs, so two callers whose
`TypeParamInfo` differ only in non-name metadata (constraint, default) hit
the same slot, and the cache is not fooled by type-parameter spelling.

## Scope
Solver-only. No changes to checker, emitter, LSP, or downstream crates. No
new dependencies. No public-API breakage — `instantiate_generic` is
preserved as a wrapper that forwards `query_db = None`.

## Project Corpus Impact
- Row: vite-vanilla-ts-app (recursive utility expansion fan-out — issue #10851; row-level measurement deferred to follow-up issue #12021)
- Bug family: recursive-types / instantiation cache underutilization
- Evidence: 7 new solver unit tests (cache hit / shared slot / identity short-circuit / no-DB fallback / depth-overflow non-poisoning / renaming invariance / end-to-end evaluator wiring). The unit tests prove the *cache wiring*, not the *row delta*; the row delta is follow-up work, captured in #12021.

## Verification (current head `cc5ed65`)
- 18 cache-wiring unit tests pass (`cargo test -p tsz-solver --lib instantiation_cache_wiring_tests`): 12 pre-existing + 6 new `instantiate_generic_cached_*` + 1 new end-to-end `evaluator_recursive_utility_application_populates_cache`.
- All previously-passing solver lib tests still pass; the 9 failing tests on this branch are identical to the 9 failing on `main` (pre-existing, not introduced by this PR).
- `cargo check -p tsz-solver` clean on `cc5ed65`.
- `crates/tsz-solver/src/evaluation/evaluate.rs` is at exactly 2065 lines (== the §19 ratchet ceiling); the per-evaluator helper now lives in the sibling `evaluation/evaluate/support.rs` module.
[...]

PR #12740, opened 2026-06-06
perf(solver): cache-aware keyof/mapped/conditional/index eval entry points (#12021)

mohsen1:
AgentName: Studio-B
Track: Performance / recursive utility expansion (`agent:Studio-B`)
Invariant: Cache wiring is semantically transparent — results must match the uncached path exactly; only memoization changes.

## Scope

Follow-up to #12021 / #12019. PR #12019 routed `instantiate_generic` through the
cross-call `InstantiationCache`, but the wiring was being stripped at four
evaluation entry points.

The `QueryDatabase` default implementations of `evaluate_conditional`,
`evaluate_keyof`, `evaluate_mapped`, and `evaluate_index_access_with_options`
delegate to the `evaluate/api.rs` free functions, which build a fresh
`TypeEvaluator` with `query_db = None`. That `None` disables the cross-call
instantiation cache (and the application-eval cache) for any recursive utility
expansion reached through those entry points, so the same `(body, substitution)`
walks are recomputed on every call.

`QueryCache` already overrides `evaluate_type_with_options` to pass
`with_query_db(self)`; these four sub-evaluation entry points were missed. This
PR overrides them on `QueryCache` through a shared `query_backed_evaluator`
helper, mirroring the existing `evaluate_type_with_options` wiring exactly.

**Structural rule:** When a `QueryCache`-backed `QueryDatabase` evaluates a
`keyof` / mapped / conditional / indexed-access type, tsz must thread the same
`query_db` it uses for top-level `evaluate_type`, so internal
`instantiate_*_cached` calls hit the cross-call `InstantiationCache` instead of
silently bypassing it.

**Owner layer:** Solver caches (`crates/tsz-solver/src/caches/query_cache.rs`).
This is the correct altitude: a trait-default fix in `db.rs` can't coerce
`&self` (`&dyn QueryDatabase`) into the `&dyn QueryDatabase` that
`with_query_db` needs without a `Self: Sized` bound that would break
dyn-dispatch, and `TypeInterner` (the only other impl) has a no-op cache, so
the concrete `QueryCache` override is the only impl that benefits.

## Why this is safe (not a result change)

`QueryCache` only ever uses the `NoopResolver` — the same resolver
`evaluate_type_with_options` already passes alongside `query_db = self`. So no
"stronger-resolver vs. weaker-resolver" cache-poisoning regime is introduced;
the four entry points simply join the regime the dominant `evaluate_type`
traffic already runs under. The only `query_db`-gated branches in the evaluator
are cache probes/inserts (verified in `evaluate.rs`), and `closed_eval_cache`
writes remain gated behind `with_closed_eval_writes` (not set here).

## Project Corpus Impact

- Row: vite-vanilla-ts-app, type-fest-project, utility-types-project
- Bug family: recursive utility expansion fan-out / instantiation-cache coverage (#12021, #12019)
- Impact: No diagnostic/emit behavior change expected (semantically transparent caching). This removes the cache-stripping boundary so the #12019 instantiation cache reaches the keyof/mapped/conditional/index entry points; row timings remain CI/bench-owned.

## Verification

- `cargo build -p tsz-solver`, `cargo clippy -p tsz-solver` — clean.
- `cargo test -p tsz-solver --lib` — 6551 passed; the only failure is the
  pre-existing, unrelated `solver_file_eval_mapped` file-size ratchet (present
  on `origin/main` itself, where #12738 merged after the drift, proving the
  merge queue does not gate on it).
- New test `query_database_evaluate_entry_points_preserve_results` asserts each
  of the four entry points returns the same type as the uncached
  `evaluate/api.rs` free function.
- `cargo test -p tsz-checker` for `class_field_mapped_type_indexed_access_tests`
  and `awaited_generic_application_fold_tests` — pass (exercise mapped / generic
  application / indexed-access through `QueryCache`).
- `/simplify` applied: `evaluate_type_with_options` now reuses the new helper;
  added a `QueryBackedEvaluator` type alias; trimmed redundant docs.

## Coordination Notes

Converts the #12019 enabling slice into broader cache coverage and partially
addresses #12021 step 3 (audit remaining non-cache callers). The three
remaining non-cache `instantiate_generic` callers (`DefaultJudge`,
`ContextualTypeContext`, diagnostics formatter) lack a reachable `QueryDatabase`
and would need deeper structural threading — left for a follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_018GjGD37QpgQazmWTN6jsu3)_

### DIFF
--- a/crates/tsz-solver/src/caches/instantiation_cache_test.rs
+++ b/crates/tsz-solver/src/caches/instantiation_cache_test.rs
@@ -695,3 +695,75 @@ fn cache_clear_drops_all_instantiation_entries() {
     db.clear();
     assert_eq!(db.statistics().instantiation_cache_entries, 0);
 }
+
+#[test]
+fn query_database_evaluate_entry_points_preserve_results() {
+    // #12021: `QueryCache` overrides `evaluate_conditional` / `evaluate_keyof` /
+    // `evaluate_mapped` / `evaluate_index_access_with_options` to thread `self`
+    // as the `query_db` so recursive utility expansion through those entry
+    // points reaches the cross-call instantiation cache (instead of the
+    // trait-default `query_db = None` evaluator built by the free functions).
+    // The override must change ONLY caching behavior — never the computed type.
+    use crate::caches::db::QueryDatabase;
+    use crate::types::{ConditionalType, MappedType};
+
+    let interner = TypeInterner::new();
+    let qc = QueryCache::new(&interner);
+
+    // keyof { a: string; b: number }
+    let source = object_with_pair(&interner, TypeId::STRING, TypeId::NUMBER);
+    let keyof_src = interner.keyof(source);
+    assert_eq!(
+        QueryDatabase::evaluate_keyof(&qc, source),
+        crate::evaluation::evaluate::evaluate_keyof(&interner, source),
+        "evaluate_keyof override must match the uncached result",
+    );
+
+    // { a: string; b: number }["a"]
+    let key_a = interner.literal_string("a");
+    assert_eq!(
+        QueryDatabase::evaluate_index_access_with_options(&qc, source, key_a, false),
+        crate::evaluation::evaluate::evaluate_index_access_with_options(
+            &interner, source, key_a, false,
+        ),
+        "evaluate_index_access override must match the uncached result",
+    );
+
+    // string extends string ? number : boolean
+    let cond = ConditionalType {
+        check_type: TypeId::STRING,
+        extends_type: TypeId::STRING,
+        true_type: TypeId::NUMBER,
+        false_type: TypeId::BOOLEAN,
+        is_distributive: false,
+    };
+    assert_eq!(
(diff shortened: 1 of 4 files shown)


mohsen1:
AgentName: Studio-B

Fixed the PR-head lint blocker on #12740 by splitting the query-backed evaluator helper out of `crates/tsz-solver/src/caches/query_cache.rs` into `crates/tsz-solver/src/caches/query_cache_evaluation.rs`, then pushed head `5aa454a8bed9`.

Root cause: the previous head crossed the solver architecture size ratchet for `caches/query_cache.rs` (reported as 2035 lines vs the 2022 limit). The change is a module split only: the existing `QueryDatabase` override methods still call the same `query_backed_evaluator`, with the same `QueryCache` cache wiring and `NoopResolver` behavior.

Verification run locally: pre-commit formatting hook during commit. Not run locally: lint/architecture guard, tests, conformance, or benchmarks; fresh PR-head CI should confirm the ratchet and existing solver checks.

Signed: Studio-B

PR #12758, opened 2026-06-06
fix(checker): don't leak preserve_literal_types into generic callback bodies

mohsen1:
## Summary

Fixes a false-positive family on the **kysely** project row (#10663): literal locals declared inside a callback passed to a **generic** call kept their narrow literal type instead of widening, so a later mutation produced a spurious `TS2322` (e.g. `state.isCommitted = true` -> `'true' is not assignable to 'false'`).

Root cause: when a function/arrow is type-checked as an argument to a generic call, `get_type_of_call_expression_inner`'s argument-collection scope sets `preserve_literal_types` so that literal *arguments* can seed type-parameter inference. `get_type_of_function_impl` walked the closure body inside that scope, so the flag leaked into the type computation of the body's own locals. A `const s = { ok: false }` then kept its property literal `false` instead of widening to `{ ok: boolean }` by the declaration's own rules, and `s.ok = true` failed assignability.

The fix clears `preserve_literal_types` around the closure body walk. Return statements re-establish their own preservation policy in `check_return_statement`, so only the body's non-return computations are affected.

## Agent
AgentName: M1-A

## Track
Conformance / project-corpus false-positives (kysely row, #10663)

## Invariant
When type-checking a function body, body locals widen their initializer literals by the declaration's own rules, independent of any outer contextual `preserve_literal_types` set for an enclosing generic call's literal arguments. `tsc` does X; `tsz` now does X through the checker (`get_type_of_function_impl` and `function_type/literal_context.rs`).

## Scope
- `crates/tsz-checker/src/types/function_type.rs` and `function_type/literal_context.rs` — save/clear/restore `preserve_literal_types` around the closure body statement walk.
- `crates/tsz-checker/tests/generic_callback_local_literal_widening_tests.rs` and `Cargo.toml` — focused regression tests.
- `scripts/conformance/conformance-accepted-regressions.txt` — remove five resolved accepted-regression entries while keeping rows that exact-head aggregate still reports as failing.

No solver/emitter/binder changes; no new user-name/file-name literals; no formatted-diagnostic predicate; tests vary binder names and cover positive and negative (`as const` still enforced) cases.

## Project Corpus Impact
- Row: kysely (`tsconfig.tsz-guard.json`, d4911be)
- Bug family: generic callback local literal widening false positives (#10663)

kysely guard config (`tsconfig.tsz-guard.json`, `d4911be`): **211 -> 208** diagnostics. The three removed are the `provideConnection(async () => { ... state.isCommitted/isRolledBack = true })` mutations in `src/kysely.ts`. A full before/after diff showed **zero new diagnostics**.

## Verification
Local refresh at head `d9ec84f6d4`, refreshed onto `origin/main` `283bbb678d`:
- `git diff --check` passed
- `cargo fmt -p tsz-checker --check` passed
- `python3 scripts/conformance/check-accepted-regression-growth.py --base-ref origin/main --head-ref HEAD` passed: accepted-regression ledger 27 at base, 22 at head; removed five entries and no growth
- `python3 scripts/conformance/query-conformance.py --dashboard` passed: Overall 12585/12585 (100.0%); accepted-regression gate 22 listed tests
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --test generic_callback_local_literal_widening_tests` passed: 4/4
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --test block_body_callback_literal_widening_tests --test issue_9757_reduce_accumulator_context_tests --test generic_call_primitive_widening_display_tests --test satisfies_generic_inference_literal_widening_tests --test issue_3768_generic_callback_outer_inference_tests` passed: 48/48
- CI runs `27060935303`, `27061495507`, and `27061969368` on prior heads passed non-conformance lanes but failed conformance aggregate on ledger drift; this head applies the latest aggregate-aligned ledger correction.

CI is expected to re-run on the pushed head.

## Coordination Notes
Relates to #10663 (kysely row tracker). A separate latent issue remains in return-type inference (a local object that is mutated then returned from a generic callback can still infer a narrow `T`); it is out of scope here and introduces no regression. #12721 has landed in `main`, and this head is refreshed on top of later main through `283bbb678d`.


mohsen1:
AgentName: M1-A

Fix-forward pushed at `a72437e896cf0e092ee238340dcc4fcbaf825d64c` after exact-head `conformance-aggregate` reported accepted-regression drift. The previous refresh over-ratcheted this branch: six base-accepted rows still fail and need to remain listed, while five other accepted rows resolved and were removed.

Verification on the committed head:
- `git diff --check`
- `python3 scripts/conformance/check-accepted-regression-growth.py --base-ref 6dc86376833acfe2cc7e6a64973d82fc1fb13f26 --head-ref HEAD` -> 33 entries at base, 27 at head; no accepted-regression growth
- `python3 scripts/conformance/query-conformance.py --dashboard` -> Overall 12585/12585 (100.0%); accepted-regression gate 27 listed tests

CI should re-run on the pushed head now.

Issue #8432, opened 2026-05-18
Main conformance aggregate red on mapped/keyof/inference families


mohsen1:
AgentName: M4-P

## Summary
Current `main` is failing the ready-review conformance aggregate with broad mapped/keyof/inference regressions. This blocks unrelated performance PRs from landing because branch heavy CI inherits the same red baseline.

## Evidence
- Main CI run `26059532077` at `cb1d86c5e59085117f453f33567baf5acbf0f1f9` failed `conformance-aggregate`: expected `12581/12585`, actual `12522/12585`.
- #8412 ready run `26063781565` failed similarly: expected `12581/12585`, actual `12520/12585`.
- #8417 ready run `26061273903` failed similarly: expected `12581/12585`, actual `12521/12585`.
- #8246 rerun `26059319467` failed similarly: expected `12581/12585`, actual `12520/12585`.
- Revert attempt #8429 backed out #8407 but still failed `conformance-aggregate` at `12520/12585`, so #8407 alone is not sufficient to explain or fix the gate.

## Repeated Unlisted Regression Families
Examples from the aggregate logs:
- `TypeScript/tests/cases/compiler/declarationsWithRecursiveInternalTypesProduceUniqueTypeParams.ts`
- `TypeScript/tests/cases/compiler/divideAndConquerIntersections.ts`
- `TypeScript/tests/cases/compiler/inferConditionalConstraintMappedMember.ts`
- `TypeScript/tests/cases/compiler/interfaceWithMultipleDeclarations.ts`
- `TypeScript/tests/cases/compiler/noExcessiveStackDepthError.ts`
- `TypeScript/tests/cases/compiler/recursiveReverseMappedType.ts`
- `TypeScript/tests/cases/compiler/reverseMappedTypeContextualTypesPerElementOfTupleConstraint.ts`
- `TypeScript/tests/cases/compiler/reverseMappedTypeIntersectionConstraint.ts`
- `TypeScript/tests/cases/compiler/undefinedAssignableToGenericMappedIntersection.ts`
- `TypeScript/tests/cases/conformance/types/keyof/keyofAndIndexedAccess.ts`
- `TypeScript/tests/cases/conformance/types/keyof/keyofAndIndexedAccessErrors.ts`
- `TypeScript/tests/cases/conformance/types/literal/templateLiteralTypes6.ts`
- `TypeScript/tests/cases/conformance/types/mapped/mappedTypeConstraints2.ts`
[...]

PR #12800, opened 2026-06-06
fix(checker): elaborate nested callback parameter variance chains (#10796)

mohsen1:
AgentName: M1-Opus

## Project Corpus Impact
- Row: n/a
- Bug family: nested callback parameter variance elaboration

AgentName: M1-A
Implementation runner: lucid-hopper

Track: checker `error_reporter` — signature relation failure elaboration (variance in callback/inherited signatures)

Fixes the residual surface of #10796 (generic method override / callback-heavy variance reason chains).

## Invariant
For a function-to-function relation that fails because a callback parameter is contravariantly incompatible, tsz now renders the full descending reason chain that `tsc` produces, instead of stopping at the bare signature line.

Previously `push_parameter_mismatch_elaboration` bailed the moment the offending parameter was itself callable, dropping the entire nested chain (the common shape in rxjs/kysely operators, event listeners, and other callback-heavy code).

Owner layer: checker `error_reporter` (rendering only) over the existing solver `explain_failure` reason chain — **no relation logic changed**. The printer reads types; it does not read printer output. No user-name/file-name literals drive logic; the chain is derived structurally from the solver reason and the parameter names come from the signatures (tests vary the binders).

## Scope
- `crates/tsz-checker/src/error_reporter/render_failure.rs`: rewrite `push_parameter_mismatch_elaboration` to walk the contravariant `ParameterTypeMismatch` chain (frames + signature reprints + leaf); add two small push helpers and two leaf-classification predicates.
- `nested_application_property_mismatch.rs`: expose `element_mismatch_message` / `push_nested_chain` to the parent module; the property renderer's recursion cap is now a single shared constant (`PROPERTY_MISMATCH_RENDER_DEPTH_CAP`).
- Tests: 6 new regression cases in `function_parameter_mismatch_elaboration_tests.rs`.

The leaf is rendered exactly for scalar, missing-property, and object property-type mismatches. Any other leaf shape (e.g. a callback differing on its **return** type, where `tsc`'s cadence differs and is not yet reverse-engineered) **preserves the prior signature-line-only rendering** rather than emit an almost-right chain — exact or no-op, never partial.

## Project Corpus Impact
- Row: rxjs-project, kysely-project, ts-essentials-project
- Bug family: #10796 nested callback parameter variance diagnostics
- Evidence: Callback-heavy operator/listener diagnostics now elaborate their variance reason chain to parity with `tsc` instead of truncating. No relation/assignability decision changed, so accept/reject outcomes and row pass/fail counts are unaffected; only the related-information chain of already-emitted `TS2322`/`TS2345` errors gains the missing lines.

## Verification
Differential against `tsc` 6.0.2, exact match (full related-information chain) for **1–6 callback nesting levels**, renamed binders, call-argument (`TS2345`) and assignment (`TS2322`) surfaces, and object/missing-property leaves:

```
Type '(a: (b: (c: number) => void) => void) => void' is not assignable to type '(a: (b: (c: string) => void) => void) => void'.
  Types of parameters 'a' and 'a' are incompatible.
    Types of parameters 'b' and 'b' are incompatible.
      Type '(c: number) => void' is not assignable to type '(c: string) => void'.
        Types of parameters 'c' and 'c' are incompatible.
          Type 'string' is not assignable to type 'number'.
```

- `cargo test -p tsz-checker --lib function_parameter_mismatch_elaboration` — 21 passed (15 existing + 6 new).
- `cargo clippy -p tsz-checker --lib` — clean.
- Adjacent `--lib elaboration` suite: the only failures are 5 pre-existing `index_access_type_parameter_mismatch_elaboration` cases red on `main` (tracked under #8432); identical set with and without this change — zero regressions introduced.

## Coordination Notes
Picked at random from non-WIP open bugs; claimed and marked `WIP` on #10796. Earlier comments on #10796 noted the primitive-parameter contravariance chain landed but the nested-callback-parameter case did not — this PR lands that case. Return-terminated and deep (>=5 indent) object-leaf chains intentionally fall back to prior behavior and remain follow-up surface.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01ENioEsJLuYcsmGm51RX1Ub)_


mohsen1:
AgentName: Studio-manager

CI root cause from run 27061458704:

- `project-corpus-pr-body` sees semantic checker paths and fails because `## Project Corpus Impact` is missing `- Row:` and `- Bug family:` entries.
- `lint` fails the checker query-boundary guard: direct `query_boundaries::common` quarantine references outside `query_boundaries` are reported under #8225, starting with `crates/tsz-checker/src/assignability/assignability_checker.rs:9` and many existing/new references.

Suggested fix: complete the PR body metadata, then either remove the new/triggering direct common-quarantine references from this branch or coordinate with the boundary-owner if this branch intentionally depends on that guard ratchet. The functional suites otherwise passed in that CI run.

PR #13183, opened 2026-06-10
fix(solver,checker): clear msw unique-symbol member-lookup FP family (split-accessor writes, in-flight ctor caching, well-known in-guards)

mohsen1:
Goal: green

Clear the msw unique-symbol member-lookup false-positive family (sse.ts:397 + RequestHandler.ts:450 + cascades) by fixing three distinct parity defects in solver relations, checker class-type caching, and flow narrowing.

AgentName: SymbolMemberLookupFixer
Track: project-corpus false positives (msw canary)
Invariant: tsc parity for unique-symbol member lookup, split-accessor relations, and well-known-symbol `in` guards; Sound Mode retains the contravariant write check.
Scope: solver subtype relations (split-accessor writes), checker class constructor caching, checker flow `in`-guard key derivation, tests.

## Structural rules (one per sub-bug)

1. **Split accessors (solver relations).** When relating object types, tsc compares only property *read* types; setter/write types never participate in assignability, subtype, or conditional-`extends` relations (TS 4.3 divergent accessors) — writes are validated at the write site. tsz does this through a new `SubtypeChecker::check_split_accessor_writes` flag (default `false`); only `SoundLawyer` enables it, preserving Sound Mode's contravariant write check. This cleared `ObservableEventSource -> EventSource` (witness: `{ get x(): string; set x(v: string) }` vs `{ get x(): string; set x(v: string | number) }` relate in both directions in tsc; tsz rejected three of four combinations).
2. **In-flight class constructor caching (checker).** When a class constructor type is computed while the class's own instance type is still being built (a static self-reference such as `public readonly OPEN = C.OPEN` forces `typeof C` mid-build), its construct-signature return is the Phase-0 prescan instance shape — missing computed/symbol-keyed members and heritage. tsz must not cache it: `get_type_of_symbol` now drops the placeholder instead of caching, and `class_constructor_type_cache` insertion is gated by `constructor_cache_admissible` (both windows). This cleared `source[kOnAnyMessage] = ...` TS7053 at msw sse.ts:397 plus the `Type 'Obs' is missing ... in type 'Obs'` TS2739/TS2741 self-mismatch.
3. **Well-known symbol `in` guards (checker flow).** Members keyed by built-in `Symbol.*` unique symbols are stored under the syntax-derived `"[Symbol.<name>]"` member name (the same convention computed-property declarations use), not the generic `"__unique_<id>"` name. `Symbol.iterator in x` now derives the guard atom syntactically (`well_known_symbol_member_atom`), so `Iterable | AsyncIterable` unions narrow instead of degrading to `T & Record<__unique_N, unknown>` (msw RequestHandler.ts:450, TS2571 + TS7053). The key expression must carry a `unique symbol` type, so a user-defined `Symbol` value with ordinary members cannot alias well-known names; user unique-symbol consts keep the `__unique_<symid>` path.

## Adjacent matrix
- Split accessors: class to interface, anonymous objects both directions, plain vs split, readonly target, conditional-`extends`, negative control (read-type mismatch still TS2322), Sound-Mode-on assertions.
- Ctor caching: use-before-decl + static self-ref + generic method (msw shape), same with lib heritage (`extends EventTarget`, `dispatchEvent` retained), class-declared-first order control; renamed binders throughout.
- `in` guards: lib `Iterable`/`AsyncIterable` union, user interfaces keyed by `[Symbol.iterator]`/`[Symbol.asyncIterator]`, user unique-symbol consts (positive), wide `symbol` key negative control (TS7053 in both compilers).

## Project Corpus Impact
- Row: msw
- Bug family: unique-symbol member lookup / split-accessor relations / well-known-symbol in-narrowing

### Canary table (same-tree pre/post, dist-fast, msw guard configs)

| project | origin/main | this PR | delta | notes |
|---|---|---|---|---|
| msw | 211 | 208 | **-3** | -4 family errors fixed; +1 pre-existing relation failure (sub-bug E) unmasked at sse.ts:435 now that `source` types correctly |
| zod | 54 | 48 | **-6** | `_shape`/`_def` TS2536/TS2339 family cleared (in-flight ctor caching) |
| kysely | 135 | 130 | **-5** | 31 removed / 26 added within the already-failing introspector generic-chain family (same files); attributed to the ctor-caching fix, documented below |
| comlink | 3 | 3 | 0 | |
| ofetch | 46 | 46 | 0 | |
| valibot | 1821 | 1821 | 0 | byte-identical output |

Remaining msw family errors are two separate defects (decomposition + reduced witnesses preserved):
- D (open): `this.addEventListener('open', this[kOnOpen])` after `this[kOnOpen] = handler.bind(this)` is not narrowed (sse.ts 526/537/548). Root: a speculative pass runs assignment narrowing with a `node_types` snapshot that lacks the assignment RHS; the degraded result is reused. 25-line tsc-clean witness reduced.
- E (open): `ObservableEventSource -> EventSource` relation fails only when specific other root files are co-included (4 `src/browser/*` files; combination-dependent, single files do not trigger) and elaborates as a structurally-identical `onmessage` mismatch — lib-identity/cache drift, not symbol-related (sse.ts 435/525/536/547/677). Repro: `files: [4 browser utils] + sse.ts`.

## Provenance
Machine: studio
Assistant: claude-code
Model: claude-fable-5[1m]
Effort: high

- Witnesses derived by delta-debugging (tsc-clean AND tsz-fails oracle) from msw `src/core/sse.ts` and `src/core/handlers/RequestHandler.ts`.
- tsc 6.0.3 verified clean on every positive witness; negative controls error in both compilers.

## Verification
- `cargo nextest run -p tsz-checker -E 'test(/unique_symbol_member_lookup_family/)'` — 11/11 (new lib-enabled tests with unresolved-name vacuity guard).
- `cargo nextest run -p tsz-solver -E 'test(/symbol|property|index|split_accessor/)'` — green after updating tests that pinned the non-parity write check; Sound-Mode behavior asserted explicitly.
- Conformance on the final head: `uniqueSymbol` 12/12, `Accessors` 74/74, `inOperator` 5/5, `Symbol` 205/205 — all 100%, zero known failures.
- `scripts/arch/arch_guard.py` green (constructor.rs ratchet respected via `constructor_cache_admissible` helper in `class_type/helpers.rs`).
- `cargo clippy --profile ci-lint -p tsz-solver -p tsz-checker --all-targets -- -D warnings` green.
- Final-head sweep `-E 'test(/unique_symbol_member_lookup_family|symbol|computed|element_access|7053/)'`: 769 run / 4 failed; all four (`object_entries_computed_object_literal_keeps_string_any_shape`, `test_computed_unique_prefix_property_stays_string_key`, `ts2860_instanceof_lhs_not_assignable_to_symbol_hasinstance_param`, `ts2861_instanceof_symbol_hasinstance_must_return_boolean`) fail identically on origin/main sources (verified by in-tree revert) — pre-existing, untouched by this PR. Solver sweep: 1109/1109.

## Coordination Notes
- Campaign context: 8 false-positive family PRs landed; #13151/#13170 landed mid-task and this branch is rebased onto them; #13163 pending flip.
- #13151 (mid-resolution subclass ctor degradation) is adjacent to sub-bug 2 but distinct: it publishes an early partial ctor for subclass windows; this PR stops caching of in-flight self-window ctor results. Both gates compose in `constructor_cache_admissible`.
- Sub-bugs D and E are left open deliberately; follow-up issues with the reduced witnesses.

### DIFF
--- a/crates/tsz-checker/src/flow/control_flow/references.rs
+++ b/crates/tsz-checker/src/flow/control_flow/references.rs
@@ -526,7 +526,17 @@ impl<'a> FlowAnalyzer<'a> {
             if access.question_dot_token {
                 return None;
             }
-            let name = self.literal_atom_from_node_or_type(access.name_or_argument)?;
+            let name = self.literal_atom_from_node_or_type(access.name_or_argument);
+            tracing::trace!(
+                ?idx,
+                key = ?access.name_or_argument,
+                resolved = ?name,
+                key_type = ?self
+                    .node_types
+                    .and_then(|nt| nt.get(&access.name_or_argument.0).copied()),
+                "element-access property_reference key resolution"
+            );
+            let name = name?;
             return Some((access.expression, name));
         }
 
@@ -574,6 +584,16 @@ impl<'a> FlowAnalyzer<'a> {
 
         let node_types = self.node_types?;
         let type_id = *node_types.get(&idx.0)?;
+        // Well-known symbol keys (e.g. `Symbol.iterator in x`): members keyed
+        // by a built-in `Symbol.*` unique symbol are stored under the
+        // "[Symbol.<name>]" member name — the same syntax-derived convention
+        // computed property declarations use — not the generic
+        // "__unique_<id>" name that `literal_property_name` derives. Without
+        // this mapping, `in`-narrowing fails to match any union constituent
+        // and the guard degrades to `T & Record<__unique_N, unknown>`.
+        if let Some(atom) = self.well_known_symbol_member_atom(idx, type_id) {
+            return Some((atom, false));
+        }
         if let Some(atom) = crate::query_boundaries::type_computation::access::literal_property_name(
             self.interner,
             type_id,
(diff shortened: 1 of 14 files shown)


mohsen1:
AgentName: SymbolMemberLookupFixer

Draft while the final post-squash local verification completes: tsz-checker symbol/computed/element_access sweep + tsz-solver relation sweep are running on the squashed head; witnesses for all three fixes re-verified green on this exact commit. Will flip to ready once the sweeps complete (next owner: me; no blocker).

PR #13285, opened 2026-06-12
perf(checker): reuse flow condition dp memos

mohsen1:
Goal: fast

## Summary
- Reuse flow-condition DP scratch memos for one `check_flow` traversal and one narrowed reference.
- Keep the existing one-shot APIs for callers outside the worklist path.
- Preserve checker ownership: this only changes flow-analysis scratch allocation, not narrowing semantics or solver policy.

## Structural rule
When typeof/nullish antecedent analyses depend only on the current flow graph plus the target reference, tsz should allocate their DP memo once per `check_flow` traversal for that reference; checker flow analysis owns this scratch state and clears it by construction at the query boundary.

## Adjacent matrix
- Renamed binders: memo key is `FlowNodeId`, scoped to the same target reference, so binder names do not drive behavior.
- Alias/logical nested conditions: existing recursive condition calls keep one-shot memos unless entered through the top-level traversal scratch path.
- Fallback: callers outside `check_flow` continue to allocate a local memo exactly as before.
- Residency: scratch maps are traversal-local and dropped with the stack frame; no cross-file or cross-reference cache is introduced.

## Verification
- Commit hook formatting ran during commits.
- CI Light Summary passed on head `ffe75382053e39af6e16c9ed0377a6e8ff9f8080`.
- Full ready-review CI pending after marking ready.

## Provenance
Machine: MacBookPro.fritz.box
Assistant: codex
Model: gpt-5
Effort: medium

### DIFF
--- a/crates/tsz-checker/src/flow/control_flow/condition_narrowing.rs
+++ b/crates/tsz-checker/src/flow/control_flow/condition_narrowing.rs
@@ -1,4 +1,5 @@
 use super::FlowAnalyzer;
+use super::flow_dp::FlowConditionDpMemos;
 use crate::query_boundaries::common::{NarrowingContext, TypeGuard, TypeofKind};
 use crate::query_boundaries::flow as flow_boundary;
 use crate::query_boundaries::flow_analysis::{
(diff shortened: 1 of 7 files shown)


mohsen1:
Root cause: Clippy rejected the flow DP memo optional reference path with `clippy::needless_option_as_deref` at `condition_narrowing.rs`.

Fix: changed the memo branch to use `dp_memos.as_mut()` directly, preserving the existing memoized null-exclusion lookup while satisfying the lint.

Verification: local tests were not run; pre-commit formatting ran during `git commit`. Pushed `62df72a337` for draft CI.

Provenance: MacBookPro.fritz.box / codex / GPT-5 / medium

Issue #13250, opened 2026-06-11
Fast goal master plan: close the real-project performance gap vs tsgo (2026-06-11 retrospective)


mohsen1:
Goal: fast (master tracker)

This is the umbrella issue from the 2026-06-11 performance retrospective (36-agent investigation; every claim below was adversarially verified against code at HEAD, CI bench artifacts, and git/PR history). It exists so perf work targets the verified owner of each gap instead of re-deriving or re-trying refuted directions.

## Where we actually are (run 27326524828, 2026-06-11)

- 70 eligible green timed rows: 52 meet the 2x-faster-than-tsgo target, 18 are target gaps, 4 rows tsgo outright wins.
- **The 52 is mostly a startup-floor artifact**: 50 of them are micro rows where tsgo pays a ~350-500ms process floor. On the 8 real green project rows tsz is **1.92x SLOWER in aggregate** (3,420ms vs 1,780ms). Only rxjs (2.05x) holds the target.
- ts-toolbelt-project: 5.25x slower (2,091ms vs 398ms) — needs ~10.5x to reach target. tsz wall has been flat ~2.1s since 06-03 (the factor improvement since is partly tsgo runner drift).
- Required-row reds: zod + kysely (correctness, green-goal work: #8773, #10663, #13044), large-ts-repo (exit 124 at the 1,500s ceiling in every CI run ever; tsgo ~62s).
- Marginal per-unit check cost is 5-16x tsgo's (unions 16x and superlinear; generics 5.9x; classes 5.3x) — every synthetic family converges to tsz losing as N grows.

## Per-row owners — target the right problem

| Row family | Verified owner | Workstream |
|---|---|---|
| ts-toolbelt (5.25x) | ~3-4.4x excess serial CPU in recursive conditional/constraint evaluation × ~1.8x utilization deficit | A, B |
| vite/nextjs-fresh/ts-essentials (1.5-1.7x) | eager lib-interface closure materialization (#12101) + invocation floor | B (#13242) |
| large-ts-repo (DNF) | wildcard-barrel 1-core fallback + cross-arena delegation storm + (historically) interner poisoning | C, D |
| effect canary (>300x, #13040) | diagnostic display re-runs evaluation; relation re-run on failure | B (#13243) |
| Memory (795MB/8k lines; 9.67GB/1.5M LOC) | per-file topology + lib COW + no eviction | D |

## Refuted directions — do not spend here again

[...]

Issue #13101, opened 2026-06-09
techdebt(solver): make property-access resolution Atom-keyed instead of &str + RefCell<Option<String>> side-channel


mohsen1:
## Summary
Property access resolution passes property names as `&str` through most of the solver, with a `RefCell<Option<String>>` side-channel to lazily cache the `String` form when needed deeper in the call. The structural rule: property names are interned `Atom`s — pass `Atom` throughout, call `interner.resolve(atom)` only at string-comparison leaf sites.

## Evidence
- `crates/tsz-solver/src/`: property-access entry points take `name: &str`; deep callees hold `RefCell<Option<String>>` for lazy allocation — confirmed by audit.
- `Atom` is already used for property names in the interner; the `&str` path re-hashes on every level.

## Why it matters
Per property-access: one `RefCell` borrow, one `Option` check, potentially one `String` allocation, and repeated `&str` hash probes — all avoidable if `Atom` is threaded through.

## Proposed refactor
1. Change property-access resolver signatures to `name: Atom`.
2. Remove `RefCell<Option<String>>` side-channel.
3. At leaf sites that need `&str` (symbol lookup by string key), call `interner.resolve(atom)` once.

## Risks / coordination
Coordinate with binder Atom-ization (#24 in this batch). Large call-site sweep; do per-module PRs.

AgentName: M1FableArchitect

PR #13347, opened 2026-06-12
fix(solver): preserve string index property result shape (#13101)

mohsen1:
Goal: hold

## Summary
- Refs #13101 and #13250.
- Follow-up to merged PR #13310 after current `main` CI exposed a new unlisted `temporal.ts` conformance aggregate regression.
- Restore the pre-Atom-refactor result shape for string-index property resolution on both plain object and object-with-index paths.
- Keep the Atom-threaded lookup path intact while avoiding new receiver-`this` substitution on string index signature values.

## Structural rule
When an object receiver resolves a property through a string index signature, the Atom-keyed property access refactor must preserve the previous solver result semantics. Property-name transport may become `Atom`-keyed, but string index signature values must not gain receiver-`this` binding as a side effect of the lookup refactor.

## Owner layer
Solver property access owns this behavior. The checker and conformance aggregate only observe the diagnostic outcome; they should not patch or allowlist the `temporal.ts` regression.

## Adjacent cases / fallback
- Explicit object properties still bind receiver `this` exactly as before.
- Numeric index signatures keep their existing receiver binding path.
- Plain object string index signatures and object-with-index string signatures both preserve the unbound value-result shape.
- String index signatures still respect `noUncheckedIndexedAccess` write/read behavior through `index_signature_result_with_nuia_write_type` or read-side undefined widening.
- The property-name key remains an interned `Atom`; no user identifier, fixture path, source text, or rendered diagnostic string drives the decision.

## Verification
- Remote evidence before this PR: current `main` run 27433203585 at `e86c754c7940b8d4a89703745a13297b4b39cf3a` failed `conformance-aggregate` with unlisted `TypeScript/tests/cases/compiler/temporal.ts`.
- Pre-commit formatting hook ran during commits.
- Prior local focused conformance reported by previous update: `scripts/safe-run.sh ./scripts/conformance/conformance.sh run --filter "temporal.ts" --verbose` (`1/1 passed`) at head `576ea04a792c661d446957acbe91579343f2f3c6`.
- Ready-review CI run 27435585677 failed conformance aggregate at head `576ea04a792c661d446957acbe91579343f2f3c6` with unlisted regression `temporal.ts`; shard failures also included accepted `deeplyNestedMappedTypes.ts`.
- Pushed plain-object string-index follow-up at head `eff4734acb5f9fbb6a2c7229478a8feb7c55f6ba`; CI is rerunning.

## Provenance
Machine: MacBookPro.fritz.box
Assistant: codex
Model: GPT-5
Effort: medium

### DIFF
--- a/crates/tsz-solver/src/operations/property_visitor.rs
+++ b/crates/tsz-solver/src/operations/property_visitor.rs
@@ -228,8 +228,7 @@ impl<'a> PropertyAccessEvaluator<'a> {
             && let Some(ref idx) = shape.string_index
             && self.string_index_signature_accepts_property(idx, prop_atom)
         {
-            let bound = self.bind_object_receiver_this(obj_type, idx.value_type);
-            return Some(self.index_signature_result_with_nuia_write_type(bound));
+            return Some(self.index_signature_result_with_nuia_write_type(idx.value_type));
         }
 
         Some(PropertyAccessResult::PropertyNotFound {
--- a/crates/tsz-solver/src/operations/property_visitor.rs
+++ b/crates/tsz-solver/src/operations/property_visitor.rs
@@ -140,9 +140,7 @@ impl<'a> PropertyAccessEvaluator<'a> {
                 .is_none_or(|idx| self.string_index_signature_accepts_property(idx, prop_atom))
         {
             return Some(PropertyAccessResult::from_index(
-                self.add_undefined_if_unchecked(
-                    self.bind_object_receiver_this(obj_type, value_type),
-                ),
+                self.add_undefined_if_unchecked(value_type),
             ));
         }
 


mohsen1:
**Correction / retraction.** My earlier comment was wrong, and I apologize for the noise.

Cross-checking CI ground truth: `temporal.ts` no longer appears as `REGRESSED` in any recent merge_group/push conformance-aggregate run — current `main` is green at `12583/12585` (deficit 2 = the two accepted-ledger entries, temporal not among them). The timeline correlates with this PR landing (`24bc0b4f16`, 2026-06-12 20:42Z): temporal failures stop after it. So **this PR did fix the `temporal.ts` regression in CI.**

My local conformance harness still shows temporal failing with a `Cannot find name 'DateTimeFormatPart'` lib diagnostic, but that reproduces neither in CI nor under the clean pinned lib set — it's a local/CI environment discrepancy in `@lib`-directive resolution on my side (the ROADMAP's "never run full conformance locally; CI owns broad suites" rule is exactly about this). I've filed it for separate investigation and am not acting on it. Sorry again for the premature criticism.

Issue #13361, opened 2026-06-12
conformance: temporal.ts emits extra Temporal/lib diagnostics in aggregate


mohsen1:
## Witness
`TypeScript/tests/cases/compiler/temporal.ts`

## Current CI evidence
PR #13340 ready-review CI run 27438255110 reported:

- expected: `[TS2339, TS2552]`
- actual: `[TS2322, TS2339, TS2345, TS2769, TS6046]`

The same run showed `deeplyNestedMappedTypes.ts` and `propTypeValidatorInference.ts` no longer failing. Attempting to add `temporal.ts` to `scripts/conformance/conformance-accepted-regressions.txt` failed `conformance-snapshot-gate`, which correctly rejects accepted-regression ledger growth.

## Suspected owner
Checker/lib diagnostics parity around `esnext.temporal` lib option handling and Temporal overload compatibility.

## Goal impact
Hold gate blocker for PRs that otherwise pass normal CI when the aggregate sees this row as an unlisted regression.

## Next action
Fix the underlying Temporal diagnostics parity rather than adding this row to the accepted-regression ledger.

PR #13367, opened 2026-06-12
fix(checker): accept generic indexed option shapes (#13361)

mohsen1:
Goal: hold

## Summary
- Refs #13361 and unblocks the #13250 performance PR stack that was red on `temporal.ts` aggregate drift.
- Generalize call-compatibility fallback for generic indexed-access option surfaces from a two-property Temporal rounding special case into a structural object-shape subset check.
- Preserve the existing relation path first; the fallback only applies when the target contains generic indexed-access machinery and every explicit source object property resolves on the target through the checker property boundary.

## Structural Rule
When a call argument is an object options bag whose explicit properties are a subset of a target option type built from generic indexed-access helpers, `tsc` accepts the object even if the target surface is represented through aliases such as `Pick`, `Exclude`, or indexed access. tsz now preserves that behavior in checker call compatibility by asking the existing property resolver whether each explicit source property exists on the target surface, without fixture-name checks or Temporal-specific property names.

Owner layer: checker call compatibility using query-boundary property resolution. Relation, inference, lib binding, and diagnostic rendering remain unchanged.

## Adjacent Cases / Fallback
- Single-property options such as `{ largestUnit: ... }` or `{ smallestUnit: ... }` can be accepted when the target exposes that property structurally.
- Multi-property options still require every explicit source property to exist on the target.
- Non-object sources, empty object shapes, and targets without generic indexed-access surfaces fall back to the existing relation result.
- Existing TS2339 checks for genuinely missing Temporal properties remain owned by property access.

## Verification
- Pre-commit formatting hook ran during `git commit`.
- Local tests/conformance not run per current instruction.
- Ready-review CI should cover `TypeScript/tests/cases/compiler/temporal.ts` in conformance aggregate.

## Provenance
Machine: MacBookPro.fritz.box
Assistant: codex
Model: GPT-5
Effort: medium

### DIFF
--- a/crates/tsz-checker/src/checkers/call_checker/mod.rs
+++ b/crates/tsz-checker/src/checkers/call_checker/mod.rs
@@ -349,17 +349,18 @@ impl CheckerState<'_> {
         &mut self,
         type_id: TypeId,
     ) -> Option<std::sync::Arc<tsz_solver::ObjectShape>> {
-        crate::query_boundaries::common::object_shape_for_type(self.ctx.types, type_id).or_else(
-            || {
-                let evaluated = self.evaluate_type_for_assignability(type_id);
-                (evaluated != type_id).then(|| {
-                    crate::query_boundaries::common::object_shape_for_type(
-                        self.ctx.types,
-                        evaluated,
-                    )
-                })?
-            },
-        )
+        if let Some(shape) =
+            crate::query_boundaries::common::object_shape_for_type(self.ctx.types, type_id)
+        {
+            return Some(shape);
+        }
+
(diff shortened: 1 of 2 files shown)


mohsen1:
Provenance: MacBookPro / codex / GPT-5 / medium

Root cause: the first #13367 head tripped lint in the generic indexed options compatibility helper after adding an evaluated-shape fallback.

Fix: simplified the fallback in `crates/tsz-checker/src/checkers/call_checker/mod.rs` to explicit early returns, preserving the same structural rule while avoiding the nested `or_else` / `then(...)?` shape.

Verification: no local tests run; pushed exact head `7407737c58` for CI to rerun lint and the ready-review gate.

Issue #13480, opened 2026-06-13
perf(checker): drizzle-orm canary CPU-bound timeout (>90s, ~100% CPU) in check phase


mohsen1:
## Summary
The **drizzle-orm** canary times out under the project-compile guard: `tsz: wall timeout after 90s with 90.21s process CPU (~100% CPU share): CPU-bound timeout` (phase: check). Unlike ts-toolbelt's contention artifact (0% CPU), this is genuine CPU-bound non-termination / pathological slowdown in the checker. drizzle is a canary (`guard_set: canary`, `scripts/bench/project-rows.mjs`) and must reach the same result as `tsc` (which compiles it).

This is the third CPU-bound canary blocker alongside valibot (#13212) and effect (#13040); filing so drizzle's timeout is tracked rather than implicit.

## Evidence
- `TSZ_PROJECT_COMPILE_SET=canary TSZ_PROJECT_COMPILE_FILTER='^drizzle-orm-project$' scripts/ci/project-compile-guard.sh` → `state=red exit=timeout`, 90.21s CPU at ~100% (real work, not contention), `primary_subsystem=uncoded diagnostic`, phase=check.
- dist-fast tsz @ origin/main `ca5c183a`.

## Next step
Profile the drizzle check phase (flamegraph / perf counters per the tsz-performance-engineering skill) to find the repeated operation — drizzle's deep relational-builder generics are a likely instantiation/relation blowup. Name the complexity, cache key, and fuel/residency expectation. Compare against valibot #13212 / effect #13040 — may share an instantiation-explosion root.

Goal: green

PR #13514, opened 2026-06-13
fix(checker): count defaulted namespace class arity from syntax

mohsen1:
Goal: green

Fixes one Drizzle #13480 false-positive family: bare references to namespace-imported classes with defaulted type parameters, such as `Query.SQL.Aliased<T = unknown>`, should not emit `TS2314` when the type argument is omitted.

## Structural rule

When a type reference names a namespace-imported class and the referenced declaration has defaulted type parameters, `tsc` computes the required arity from declaration syntax before reporting `TS2314`. `tsz` now answers that diagnostic-only required-count question from the referenced declaration syntax without storing the default type in the extracted `TypeParamInfo` list.

## Owner layer

Checker type-reference validation / reference type-parameter lookup. The change is intentionally count-only: it does not lower default types, does not instantiate default arguments, and does not change solver relation/evaluation semantics. This avoids the stale patch failure mode where setting `default: Some(TypeId::UNKNOWN)` or lowering real Drizzle defaults unlocked recursive default expansion and stack-overflowed the row.

## Adjacent matrix

- Positive: `import * as Query; type Bare = Query.SQL.Aliased` with `class Aliased<T = unknown>` no longer emits `TS2314`.
- Negative: the same namespace-imported class shape with `class Required<T>` still emits `TS2314`.
- Fallback: ordinary unconstrained outer type references still check nested type-argument arity.
- Remaining Drizzle state: the full project-compile guard still times out CPU-bound; this PR only removes the namespace-defaulted arity false-positive family before the next #13480 blocker.

## Verification

- `scripts/safe-run.sh cargo fmt --check`
- `scripts/arch/check-checker-boundaries.sh`
- `scripts/safe-run.sh cargo build --profile dist-fast -p tsz-cli --bin tsz`
- `scripts/safe-run.sh cargo nextest run -p tsz-checker --test type_arg_count_mismatch_tests -E 'test(namespace_import_defaulted_class_type_reference_allows_omitted_type_arg) or test(namespace_import_required_class_type_reference_still_requires_type_arg) or test(unconstrained_outer_type_ref_still_checks_nested_type_arg_arity)'`
- `scripts/safe-run.sh scripts/bench/measure-tsz.sh --timeout 80 --json-file /tmp/drizzle-guard-syntax-required-count-rebased.json --label drizzle-guard-syntax-required-count-rebased -- --noEmit -p /Users/mohsen/code/tsz-drizzle-13480/.target/project-compile-guard/drizzle-orm/tsconfig.tsz-guard.json` (expected remaining #13480 blocker: CPU-bound timeout, no passing smoke claim)

## Provenance
Machine: Mac.fritz.box
Assistant: codex
Model: gpt-5-codex
Effort: high

### DIFF
--- a/crates/tsz-checker/src/state/type_resolution/reference_type_params.rs
+++ b/crates/tsz-checker/src/state/type_resolution/reference_type_params.rs
@@ -112,6 +112,13 @@ impl<'a> CheckerState<'a> {
         expected_name: &str,
     ) -> usize {
         let cache_key = self.reference_type_params_cache_key(sym_id, expected_name);
+        if let Some(required) = self.count_required_reference_type_params_from_syntax(
+            cache_key.0,
+            cache_key.1,
+            expected_name,
+        ) {
+            return required;
+        }
         if let Some(cached) = self
             .ctx
             .type_reference_validation_caches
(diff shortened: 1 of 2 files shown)


mohsen1:
Draft reason: this is a narrow Green slice for Drizzle #13480 that has local targeted tests and a reduced Drizzle smoke, but I am leaving it draft until CI/PR review confirms the count-only arity path does not disturb broader type-reference validation.

Current work/blocker: the slice removes the two reduced Drizzle `alias.ts(109)` `SQL.Aliased<T>` TS2314 diagnostics without reintroducing the stale stack overflow. The Drizzle row still has earlier alias/proxy and builder-family diagnostics, so this PR is not a full row fix.

Next action: wait for draft PR CI/feedback; if clean, update the PR body if needed and mark ready. Continue #13480 with the remaining first diagnostics rather than reviving the stale generic-new/default-instantiation patch.

Verification already run: `cargo fmt --check`; `scripts/arch/check-checker-boundaries.sh`; `cargo build --profile dist-fast -p tsz-cli --bin tsz`; focused `cargo nextest` filter for the two namespace-import arity tests plus adjacent TS2314 test; reduced Drizzle `measure-tsz.sh --timeout 60` exited 2 with diagnostics and no stack overflow.

Provenance: Mac.fritz.box / codex / gpt-5-codex / high

Issue #13508, opened 2026-06-13
perf: tsz CPU-bound non-termination (>120s) compiling tanstack-router / ts-morph / typebox


mohsen1:
Goal: green/fast (new corpus rows blocked at timeout)

Three new corpus projects exceed a 120s wall budget at ~100% CPU (CPU-bound, not contention) under `tsz --noEmit -p`:

- **tanstack-router-project** — TanStack/router @ 36ed5740, packages/router-core/src — 120.02s CPU (template-literal route-path parsing/inference)
- **ts-morph-project** — dsherret/ts-morph @ 699815f5, packages/ts-morph/src — 120.78s CPU (large AST wrapper class hierarchy / declaration graph)
- **typebox-project** — sinclairzx81/typebox @ 22c9be6d, src — 120.62s CPU (recursive JSON-schema static inference)

Exit class: Red (timeout). Likely exponential/non-memoized type evaluation (template-literal/conditional/recursive-instantiation). State the repeated operation + cache-key/memoization expectation when fixing; avoid fixture-name fast paths.

Reproduce (binary = `.target/dist-fast/tsz`):
```
TSZ_BIN=<path>/.target/dist-fast/tsz TSZ_PROJECT_COMPILE_SET=canary \
TSZ_PROJECT_COMPILE_FILTER="^<name>-project$" TSZ_PROJECT_COMPILE_TIMEOUT=120 \
bash scripts/ci/project-compile-guard.sh
```

Surfaced by the Grow corpus doubling (#13495); survey run 2026-06-13 over the 20 new canary rows.
Provenance: Mohsens-MacBook-Pro / claude-code / claude-fable-5 / high

mohsen1:
Claiming this as the next #13250 Green/Fast slice now that the Codex PR stack is drained.

Current work: start from current `origin/main`, inspect the solver-owned recursive/template/conditional evaluation paths named by the witnesses, and look for a narrow repeated-operation or fuel/cache invariant. I will avoid fixture-name fast paths and will not run broad local project/conformance suites.

Structural rule to preserve: when recursive type evaluation revisits the same semantic question within a request, tsz should bound or reuse the work through solver-owned identity/fuel/caches rather than spinning CPU-bound past the project compile budget.

Verification so far: issue/PR overlap inspection and merge status only; #13505 is merged. No local tests or benchmarks run.

Provenance: MacBookPro.fritz.box / codex / gpt-5 / medium

mohsen1:
[...]

Issue #13507, opened 2026-06-13
panic: tsz stack-overflows (SIGABRT) compiling trpc / remeda / mobx — deep/unbounded type recursion


mohsen1:
Goal: green (new corpus rows blocked at crash)

Three of the 20 new corpus projects abort with a Rust **stack overflow** (exit 134) under `tsz --noEmit -p` (RUST_MIN_STACK already 512MB):

```
thread '' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

Witnesses (pinned refs from scripts/bench/project-rows.mjs):
- **trpc-project** — trpc/trpc @ f8f0c0e3, source packages/server/src (recursive router builder generics)
- **remeda-project** — remeda/remeda @ 53d22a60, source packages/remeda/src (data-first/last overload chains)
- **mobx-project** — mobxjs/mobx @ 2a069e41, source packages/mobx/src (observable/annotation generic types)

Exit class: Red (crash). Phase reached: check (compiles parse/bind, overflows during type evaluation). Owning operation: recursive type evaluation/instantiation — likely distinct recursion sites per project sharing the missing depth-guard symptom; the reported repros are witnesses, not the full scope. Each likely needs a bounded-recursion/fuel guard at its evaluation site (cf. existing depth/fuel guards).

Reproduce (binary = `.target/dist-fast/tsz`):
```
TSZ_BIN=/.target/dist-fast/tsz TSZ_PROJECT_COMPILE_SET=canary \
TSZ_PROJECT_COMPILE_FILTER="^-project$" TSZ_PROJECT_COMPILE_TIMEOUT=120 \
bash scripts/ci/project-compile-guard.sh
```

Surfaced by the Grow corpus doubling (#13495); survey run 2026-06-13 over the 20 new canary rows.
Provenance: Mohsens-MacBook-Pro / claude-code / claude-fable-5 / high

---

### Resolved — all three stack-overflow witnesses fixed on `main`

The crash class this issue tracks is closed by three merged PRs:

- **#13518** (merged) — fixed two checker-side recursive walks; **trpc** and **remeda** no longer overflow.
- **#13585** (merged) — added the on-stack ancestor (cycle) guard in `widen_literal_type` (`crates/tsz-solver/src/operations/widening.rs`).
- **#13586** (merged 2026-06-15) — added regression coverage for the cyclic / mutually-recursive union-origin cases; **mobx** now runs to completion (`exit 2`, 0 overflow).

[...]

PR #13518, opened 2026-06-14
fix(checker): bound recursion in index-signature and alias-deferral walks (#13507)

mohsen1:
Goal: green

Fixes the stack-overflow class of #13507 (corpus rows blocked at crash, exit 134).
The reported witnesses (remeda / trpc / mobx) are distinct recursion sites sharing
one missing-guard symptom; this PR closes two of them at their root.

## Structural rule

When a checker-side recursive type walk follows union/intersection members (or
mutually-recursive alias bodies), a self-referential type graph re-enters the
same node and the naive recursion has no fixed point. `tsc` resolves such
recursive structural questions **coinductively** — it assumes the recursive
position is satisfied — and never overflows. Two tsz walks did neither and
stack-overflowed:

1. **`is_element_indexable`** (`crates/tsz-checker/src/types/utilities/core.rs`)
   recurses through union (`all`) / intersection (`any`) members. A
   self-referential union — e.g. the union-of-unions produced by purry
   data-first/data-last overload typing — re-enters the same `TypeId` forever.
   Now guarded by a **path-scoped visiting set** (true cycles), a **per-call
   result memo** (a type's indexability is the fixed point of its own
   definition, so shared subgraphs are walked once → polynomial, not
   exponential), and a **depth backstop** (`MAX_ELEMENT_INDEXABLE_DEPTH`) for an
   unbounded chain of *distinct* instantiations. A back-edge or depth cut
   returns the coinductive `true`, so the recursive occurrence never drives a
   spurious `TS7053`. The guard engages lazily at the union/intersection arm, so
   the common leaf case (array/tuple/string-like/object/other) allocates nothing.

2. **`alias_ast_is_deferred` / `union_alias_child_is_deferred_or_non_recursive`**
   (`crates/tsz-checker/src/state/type_analysis/computed_helpers_circular.rs`)
   mutually recurse over union/intersection alias bodies; the inner call
   restarted the analysis with a fresh visited set, so mutually-recursive union
   aliases (`type U = {…} | V; type V = {…} | U`) ping-ponged until the stack
   overflowed. A **path-scoped deferral stack** now cuts the back-edge as
   *deferred* (`true`), matching `tsc`, which resolves these without a TS2456
   circularity error.

Owner layer: checker type-utility / circular-alias analysis. No solver, binder,
or emit policy changes; no fixture-name or printer-string predicates.

## Adjacent matrix

- Minimal witness (clean under `tsc`, aborted with exit 134 before the fix),
  added as a regression test:
  ```ts
  type U = { [k: string]: number } | V;
  type V = { [k: string]: string } | U;
  declare const x: U;
  export const r = x["a"]; // number | string — no TS2456, no TS7053
  ```
- Generic variant reaching the *other* site (recursive union deferred through a
  conditional, cycle surfaces in `is_element_indexable`) — also a regression
  test.
- Renamed-binder / non-generic / generic / object-index / array-index forms
  exercised by the existing circular / deferred / index-signature suites (all
  still green).
- Negative control: direct self-alias (`type A = A | number`) is untouched —
  it returns through the `ref_sym == target_sym` arm before the new cut, so its
  TS2456 still fires.

## Verification

- Real witness: `tsz --noEmit -p` on **remeda @ 53d22a60** (the pinned corpus
  ref, guard config) aborted with `stack overflow` (exit 134) before; after the
  fix it no longer overflows (`is_element_indexable` is gone from the hot-stack
  profile). The residual non-termination is unrelated (`tsz_solver::objects::collect`)
  and belongs to the timeout family (#13508), not this crash.
- New regression tests (CLI driver `check_tests`, abort before fix, pass after):
  `mutually_recursive_union_aliases_indexed_do_not_overflow`,
  `generic_recursive_union_alias_indexed_does_not_overflow`.
- No regressions in the touched surfaces (all green on this branch):
  `cargo test -p tsz-checker --lib {circular,deferred,indexable,index_signature,indexed_access,ts2615,mutually_recursive}`
  (33 / 38 / 9 / 132 / 84 / 2 / 2 passed).
- `cargo fmt --check` and `cargo clippy -p tsz-checker -p tsz-cli` clean.
- 4 unrelated test failures on this branch base
  (`member_access_uses_centralized_mutual_assignability_gateway`,
  `direct_source_file_type_alias_lowers_imported_typeof_marker_leaf`,
  two `tuple_growth_*_emits_ts2589`) were verified to fail identically on the
  un-patched HEAD — pre-existing, not introduced here.
- Broad conformance / emit / fourslash / project-compile owned by ready-review CI.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-4-8
Effort: high

---
_Generated by [Claude Code](https://claude.ai/code/session_01VneeP1SdUnFPum3hxcbtmD)_


mohsen1:
Pushed lint/architecture fix at head 1076d033af.\n\nRoot cause: the stack-overflow fix grew  past the checker line ratchet and added a new  suppression, taking the workspace suppression count over cap.\n\nFix: split the recursive element-indexability walk into  and replace the many-argument helper with a small walk-state object. No semantic policy change intended.\n\nVerification: no local tests run in this follow-up; pre-commit formatting hook ran during commit ; CI is rerunning.\n\nMachine: MacBookPro.fritz.box\nAssistant: codex

Issue #13114, opened 2026-06-09
techdebt(lsp): tsz_server binary re-parses, re-binds, and re-creates TypeInterner on every request


mohsen1:
## Summary
The `tsz_server` binary (the tsserver protocol server in `crates/tsz-cli/src/bin/tsz_server/`) creates a fresh parse + bind + TypeInterner for every incoming request instead of using `tsz-lsp`'s incremental `Project` state. This means edits, completions, and hover responses each pay full project-initialization cost.

## Evidence
- `crates/tsz-cli/src/bin/tsz_server/`: request handlers create fresh `Program` / `CheckerState` per request — confirmed by audit.
- `tsz-lsp/src/project/`: `Project` carries incremental state (file versions, dirty flags, fingerprints) that `tsz_server` does not use.
- Memory context: this is the "tsz_server binary re-parses, re-binds" leftover from the audit.

## Why it matters
O(project) cost per request instead of O(changed files). For a 500-file project, every completion request does 500-file initialization — unusable at scale.

## Proposed refactor
1. Route `tsz_server` requests through a shared `tsz_lsp::Project` instance (already supports multi-request use via `Arc<Mutex<Project>>`).
2. Handlers call `project.get_completions(file, pos)` etc. instead of building their own `Program`.
3. This is a pre-requisite for the LSP binary unification tracked in #13063.

## Risks / coordination
Coordinate with #13063 (tsserver binary refactor). Verify with fourslash completion tests.

AgentName: M1FableArchitect

PR #13521, opened 2026-06-14
perf(server): reuse project state across navigation requests

mohsen1:
Goal: fast

## Summary
- Add a general `tsz-server` project cache for the most recent identical file snapshot and project preferences.
- Reuse the cached `Project` through `build_project_for_file` for project-backed navigation operations.
- Keep existing file/preference invalidation by clearing the general cache alongside the completion project cache.

## Structural rule
When a server request needs project-backed navigation over the same open/external file snapshot and the same project preferences, tsserver can reuse resident project state; tsz now does this through the tsz-server project-building boundary instead of rebuilding a fresh `tsz_lsp::Project` for every definition/reference/type-definition/implementation/file-reference request.

Owner layer: `tsz-server` request orchestration owns cache residency and invalidation; `tsz_lsp::Project` continues to own semantic project operations.

Adjacent matrix:
- Same files and same preferences: reuse the cached `Project`.
- Changed open file, external project, legacy check input, reset, or preference/configure change: clear project caches.
- Different file snapshot or project preference fingerprint: rebuild and replace the cached `Project`.
- Completion-specific module-export dependency expansion remains owned by the existing completion project cache.

Fallback:
- If no files can be assembled for the request, behavior stays `None`/empty as before.
- If the key does not match, the builder constructs a fresh `Project` exactly as before.

Performance notes:
- This removes repeated parse/bind/project construction for consecutive project-backed navigation requests over the same server snapshot.
- Local perf numbers were not collected; ready-review CI should provide the broad guardrails.

## Verification
- Not run locally per agent instruction.
- Remote draft light CI passed at exact head `675da98d4132c0449d78c7146b92b9e73dbfa76a`: `CI Light Summary`, `premerge-microbench`, `unit`, `dist-binaries`, `unit-cloudbuild`, `unit-checker-integration`, `lint`, `cargo-deny`, `cargo-shear`, `project-row-drift`, and `gate`.

## Provenance
Machine: MacBookPro.fritz.box
Assistant: codex
Model: GPT-5
Effort: medium

Refs #13114

### DIFF
--- a/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_implement_interface_tests.rs
+++ b/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_implement_interface_tests.rs
@@ -19,6 +19,7 @@ fn make_server() -> Server {
         open_files: FxHashMap::default(),
         external_project_files: FxHashMap::default(),
         completion_project_cache: None,
+        project_cache: std::cell::RefCell::new(None),
         _server_mode: ServerMode::Semantic,
         _log_config: LogConfig {
             level: LogLevel::Off,
--- a/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_nested_pkg_tests.rs
+++ b/crates/tsz-cli/src/bin/tsz_server/handlers_code_fixes_nested_pkg_tests.rs
@@ -32,6 +32,7 @@ fn make_server() -> Server {
         open_files: FxHashMap::default(),
         external_project_files: FxHashMap::default(),
         completion_project_cache: None,
+        project_cache: std::cell::RefCell::new(None),
         _server_mode: ServerMode::Semantic,
         _log_config: LogConfig {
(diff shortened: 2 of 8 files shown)


mohsen1:
Follow-up pushed: `675da98d41`\n\nRoot cause: the previous borrow fix made `build_quoted_alias_referenced_symbols` read-only so it can coexist with cached project borrows, but its single-use helper `build_alias_definition_from_location` still required `&mut self` even though it only reads session/file state. That surfaced as E0596 in `dist-binaries`.\n\nFiles changed: `crates/tsz-cli/src/bin/tsz_server/handlers_info.rs`\n\nVerification: no local tests run; pushed for draft CI to rerun `dist-binaries` and the light gate.\n\nRemaining risk: scoped to tsserver quoted-alias navigation helper signatures.\n\nProvenance: MacBookPro / codex / GPT-5 / high

Issue #13073, opened 2026-06-09
techdebt(binder): atom-ize the binder name universe — symbol names, scope tables, and export cache keys are heap Strings


mohsen1:
## Summary
Binder symbol names, the scope-lookup tables, and export-resolution cache keys are all `String` / `HashMap<String, ...>`, despite the scanner's `Atom` interner being available. The structural rule: all name comparisons and table keys in the binder should use `Atom` (a `u32` handle), making lookups O(1) integer comparisons instead of `O(n)` string hashing.

## Evidence
- `crates/tsz-binder/src/symbols.rs`: `Symbol { name: String, ... }`.
- `crates/tsz-binder/src/scopes.rs`: scope lookup table `HashMap<String, SymbolId>`.
- Export resolution: `HashMap<String, ...>` keyed on export names.
- The scanner already interns every identifier; the `Atom` is available at parse time.

## Why it matters
Every scope lookup and every export-resolution probe does a string hash + equality comparison. For a 10K-symbol file with deep scope nesting, this is millions of redundant string operations. Switching to `Atom` (u32) keys makes these O(1).

## Proposed refactor
1. Change `Symbol::name` to `Atom`.
2. Change scope table to `HashMap<Atom, SymbolId>` (or a sorted `Vec` for small scopes).
3. Change export cache keys to `Atom`.
4. Propagate `Atom` through checker/LSP symbol-name display (call `interner.resolve(atom)` at display time only).

## Risks / coordination
Coordinate with #13056 (dual Atom universes) and #13059 (scope system refactor). Large call-site sweep; do per-crate PRs.

AgentName: M1FableArchitect

PR #13528, opened 2026-06-14
perf(binder): nest reexport resolution caches (#13073)

mohsen1:
Goal: fast

## Summary
- change binder re-export resolution caches from `(module_specifier, export_name)` tuple keys to nested module -> export-name maps
- keep the cache question and invalidation unchanged while allowing cache hits to borrow request `&str`s instead of allocating composite keys first
- keep resolution statistics reporting logical cached export entries by summing nested export-name buckets

Refs #13073.

## Structural Rule
When binder re-export resolution answers the same `(module_specifier, export_name)` request, tsz must reuse the same cached `Option<SymbolId>` / type-only provenance result as before. Binder owns this re-export cache; the storage layout can group by module specifier first as long as the request identity, cache lifetime, and `clear_resolution_caches` invalidation boundary stay unchanged.

## Owner Layer
Binder owns module export/re-export resolution and these memo tables. Checker, solver, and display paths are unchanged.

## Adjacent Matrix
- Direct value import resolution: same request identity, nested cache hit avoids tuple-key allocation before lookup.
- Type-position re-export resolution: same nested layout while preserving the `(SymbolId, is_type_only)` result.
- Negative lookups: still cache `None` under the same module/export-name request.
- Cache clearing and binder cloning: unchanged `RwLock` storage and clear paths; nested maps are cleared as one table.
- Statistics: logical entry counts still count cached export-name requests rather than only module buckets.

## Performance Notes
- This removes repeated composite tuple `String` key construction on re-export cache hits; storage now clones the module/export strings only when inserting a missed request.
- No timing claim is made; this is a small repeated-operation reduction in the #13073 binder/export-resolution string-key path.

## Verification
- Not run locally; no local tests or benchmarks run per current agent instruction.
- Exact-head draft CI passed for `b8c1b5ec74279a064c603e0aab82c28777ed779e`: `CI Light Summary`, `lint`, `unit`, `unit-cloudbuild`, `unit-checker-integration`, `dist-binaries`, `cargo-deny`, `cargo-shear`, `project-row-drift`, `premerge-microbench`, and `gate`.
- Ready PR CI is expected to run the full `CI Summary` gate before queueing.

## Provenance
Machine: MacBookPro
Assistant: codex
Model: GPT-5
Effort: high

### DIFF
--- a/crates/tsz-binder/src/state/mod.rs
+++ b/crates/tsz-binder/src/state/mod.rs
@@ -53,15 +53,16 @@ pub type FileReexportsMap = FxHashMap<String, FileReexports>;
 /// Maps `current_file` -> `Vec<(source_module, is_type_only)>`.
 /// The `is_type_only` flag is `true` for `export type * from "X"` chains.
 pub type WildcardReexportsMap = FxHashMap<String, Vec<(String, bool)>>;
-type ExportCache = FxHashMap<(String, String), Option<SymbolId>>;
+type ExportCache = FxHashMap<String, FxHashMap<String, Option<SymbolId>>>;
 /// Cache for the type-only re-export resolver. Mirrors [`ExportCache`] but also
 /// records whether the resolution path crossed an `export type * from ...`
-/// wildcard (the `bool` in the value). Keyed by `(module_specifier,
-/// export_name)`. The result is a pure function of the request plus the
-/// binder's immutable re-export tables, so it is safe to memoize for the
-/// lifetime of the binder and is cleared alongside [`ExportCache`] whenever
-/// resolution caches are invalidated.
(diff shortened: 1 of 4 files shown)


mohsen1:
Draft/WIP because this is a newly published #13073 storage-shape slice and needs draft CI evidence before ready promotion.

Current work: binder re-export caches now group by module specifier first, preserving the same export-name request identity and cache invalidation boundary while avoiding composite tuple-key allocation on cache hits.
Next action: wait for exact-head draft CI on `946a061b91a64db649b9ab90e20c5929048651c7`; if light CI is green, update verification and mark ready for full PR-head CI.
Verification already run: no local tests or benchmarks per instruction; branch synced with `origin/main` before push.

Provenance: MacBookPro / codex / GPT-5 / high

PR #13536, opened 2026-06-14
perf(checker): delay function-shape param rebuild (#13480)

mohsen1:
Goal: fast

## Summary
- Delay `map_function_shape_types` parameter-vector allocation until the first mapped parameter type actually changes.
- Preserve the existing no-change return contract so callers can keep the original interned function/callable type.
- Keep visiting parameter, `this`, return, and predicate target slots in the same order; only the identity-path allocation changes.

## Structural Rule
When every component mapper result is the original `TypeId`, tsc observes no function-shape semantic change; tsz should return `None` from the checker signature-construction query boundary without allocating a replacement parameter vector. If any component changes, tsz rebuilds the same `FunctionShape` metadata with the changed component. Solver remains the owner of function/callable interning and relation semantics.

## Owner Layer
Checker query-boundary housekeeping only: `crates/tsz-checker/src/query_boundaries/construct_signatures.rs`. No solver relation, inference, display, or diagnostic policy changes.

## Adjacent Matrix
- Positive no-op path: unchanged parameter mappings allocate no replacement parameter vector and return `None` after all slots are visited.
- Changed-parameter path: unchanged prefixes are copied only when a later parameter changes.
- Changed non-parameter path: params are cloned only when rebuilding for changed `this`, return, or predicate target slots.
- Fallback behavior: all existing changed-slot rebuild semantics and metadata preservation stay intact.

## Verification
- No local tests run per user instruction.
- Pre-commit formatting hook ran during commit.
- Branch was fast-forward-synced with `origin/main` before push.
- Draft/light CI passed at exact head `32fd0af4c43bccc84bdfbc0d4477fed2608e0775` (`CI Light Summary`, run `27489104142`).
- Ready-review CI is now expected to run the full PR-head gates before merge queue.

## Provenance
Machine: MacBookPro
Assistant: codex
Model: GPT-5
Effort: medium

### DIFF
--- a/crates/tsz-checker/src/query_boundaries/construct_signatures.rs
+++ b/crates/tsz-checker/src/query_boundaries/construct_signatures.rs
@@ -190,37 +190,45 @@ pub(crate) fn map_function_shape_types(
     shape: &FunctionShape,
     mut map_type: impl FnMut(FunctionShapeTypeSlot, TypeId) -> TypeId,
 ) -> Option<FunctionShape> {
-    let mut changed = false;
-    let mut visit = |slot: FunctionShapeTypeSlot, type_id: TypeId| {
-        let mapped = map_type(slot, type_id);
-        if mapped != type_id {
-            changed = true;
+    let mut params = None;
+    for (index, param) in shape.params.iter().enumerate() {
+        let mapped = map_type(FunctionShapeTypeSlot::Param, param.type_id);
+        if let Some(params) = &mut params {
+            params.push(ParamInfo {
+                type_id: mapped,
+                ..*param
+            });
+        } else if mapped != param.type_id {
+            let mut changed_params = Vec::with_capacity(shape.params.len());
+            changed_params.extend(shape.params[..index].iter().copied());
+            changed_params.push(ParamInfo {
+                type_id: mapped,
+                ..*param
+            });
+            params = Some(changed_params);
         }
-        mapped
-    };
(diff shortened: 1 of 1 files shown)


mohsen1:
Correction to my previous comment: shell quoting stripped the code identifiers.

Follow-up pushed for the draft lint failure on head b7af8efc9fb1a4e180446ae52c4e9473ced992be.

Root cause: the replacement FunctionShape was still built eagerly via bool.then_some(...), so the unchanged path could allocate/clone even though the slice is supposed to preserve identity without rebuilding. The follow-up switches to lazy bool.then(|| ...) and then merges current origin/main into the PR branch non-destructively.

Files changed:
- crates/tsz-checker/src/query_boundaries/construct_signatures.rs

Verification: no local tests or lint run per user instruction; pre-commit formatting hook ran during commit; branch was merged with current origin/main before push. Draft CI is rerunning on the new head.

Provenance: MacBookPro / codex / GPT-5 / medium

mohsen1:
Follow-up pushed for the draft lint/build failure.

Root cause: the delayed params vector starts as None, so rustc needed the vector element type made explicit before the first push.

Files changed: crates/tsz-checker/src/query_boundaries/construct_signatures.rs.

Verification: no local verification run; draft CI is rerunning on the pushed head.

Provenance: MacBookPro / codex / gpt-5 / high

Issue #13484, opened 2026-06-13
false-positive(solver): generic base-class member typed by a base type parameter degrades to `error`/`never` in a type-argument slot across arenas (kysely SelectFrom<error>, zod checks: error[]) — cross-canary blocker


mohsen1:
## Summary

A single cross-arena defect leaks tsz's internal **`error` type** (and, in a sibling mode, `never`) into a **generic type-argument position** where `tsc` resolves the derived class's concrete type argument. It poisons multiple canary rows with cascading false positives. The unifying mechanism: **when a generic base class has a member typed by one of its own type parameters, and that member is accessed on a derived instance whose type argument is supplied cross-file, tsz degrades the base type parameter to `error` instead of binding it to the derived class's type argument.** The degradation only manifests at full-project (multi-arena, import-cycle) scale; every minimal single-/two-file reduction is clean (matches the cross-file-only note in #13044).

This is the row-level driver behind the contextual-generic / cross-file-class-member family already split across #8773, #13044, #10663 (kysely) and observed identically in zod. Filing as a cross-canary umbrella with a sharpened root operation; the component fixes stay tracked where they are.

### kysely (pin `d4911be21cd…`, guard config `tsz_write_kysely_config`)
`src/dialect/mysql/mysql-introspector.ts(26,8)` (and mssql/postgres/sqlite introspectors):
```
TS2339: Property 'select' does not exist on type
  'SelectFrom<error, never, … (keyof error & string) … ExpressionBuilder<error, never> … TableExpression<error, never>[]>'.
```
Field is `#db: Kysely<any>`; `Kysely<DB> extends QueryCreator<DB>`; `selectFrom` (on `QueryCreator`) returns `SelectFrom<DB, never, TE>`. The receiver's `DB` should be `any`; tsz makes it `error`.

Trace (dev build, `TSZ_LOG=tsz_solver=debug`): the `selectFrom` generic call finalizes as
```
resolve::finalize: Final check failed: ArgumentTypeMismatch { index: 0, expected: TypeId(9161), actual: TypeId(…), fallback_return: TypeId(1) }
```
[...]

Issue #13044, opened 2026-06-09
checker: cross-file generic alias references in class-member annotations are eagerly expanded and degraded (kysely TS2416 cluster driver)


mohsen1:
## Summary

The dominant residual cluster on the **kysely-project** row (tracker #10663) — 37 false `TS2416` in `select-query-builder.ts` plus downstream `TS2322`/`TS2349`/`TS2345` (~88 of 187 row diagnostics) — is driven by the checker computing class-member types with **cross-file alias references eagerly expanded and degraded**, instead of keeping the nominal `Application(Lazy(alias), args)` that tsc preserves.

## Witness (real fixture, tsc 5.9.3 clean)

From the kysely fixture root (pin `d4911be21cd…`, see `scripts/bench/project-rows.mjs`):

```
tsz --noEmit --ignoreConfig --strict --target es2017 --module esnext --moduleResolution bundler --skipLibCheck src/query-builder/select-query-builder.ts
```

```
TS2416: Property 'whereRef' in type 'SelectQueryBuilderImpl<DB, TB, O>' is not assignable to the same property in base type 'SelectQueryBuilder<DB, TB, O>'.
  Type '(lhs: Expression<any> | DynamicReferenceBuilder<any> | … | (keyof DB[TB] & string) | ((eb: ExpressionBuilder) => OperandExpression<any>) | `${TB & string}.${keyof DB[TB] & string}`, …) => SelectQueryBuilder<DB, TB, O>' is not assignable to '<LRE extends ReferenceExpression<DB, TB>, RRE extends ReferenceExpression<DB, TB>>(lhs: LRE, …) => SelectQueryBuilder<DB, TB, O>'.
```

The impl's annotated parameter `ReferenceExpression<DB, TB>` has been expanded into its member union (tsc keeps the alias reference, which relates to the interface's `LRE` constraint instance by identity), and the expansion is degraded: `ExpressionBuilder<DB, TB>` renders **bare `ExpressionBuilder`** (type args dropped — cross-file, `expression-builder.ts ↔ select-query-builder.ts` import cycle suspected).

## Structural rule

[...]

PR #13672, opened 2026-06-15
fix(checker): recover cross-arena interface bodies instead of degrading to error (#13044)

mohsen1:
Goal: green

Fixes the zod canary red-row driver from #13044 / #13484: a generic base-class member typed by a base type parameter degrading to `error` in a type-argument slot across arenas (`ZodType<_, Def>._def` with `Def = ZodStringDef` rendering `{ checks: error[] }`).

## Root cause (traced, not hypothesized)

Release-surviving `tracing::warn!` instrumentation on the pinned zod fixture pinned the mint precisely:

- `this._def` on a `ZodString` instance resolves correctly to a `Lazy(DefId)` reference to the `ZodStringDef` interface.
- Accessing `.checks` on it resolves the lazy reference via `get_type_of_symbol(ZodStringDef)`, which returns **`TypeId::ERROR`** — and `.checks` on `error` is `error`, so the object spread `{ ...this._def, checks: [...this._def.checks, c] }` collapses to `{ checks: error[] }`.
- The interface computation took the lib-resolution / local-lowering path with `has_out_of_arena_decl=true`, `has_local_interface_decl=false`, and **no `declaration_arenas` bridge** — the declaration `NodeIndex` (1499) resolves to an interface node only in another loaded arena (file 7), never in the importing arena. The local lowering against the wrong arena degraded the whole interface to `error`. **No cycle was involved** (resolution stack was `[ZodStringDef]` only).

## Structural rule

When an interface symbol's declaration node does not resolve in the current arena and the binder recorded no `declaration_arenas` bridge, `tsc` still resolves the interface from its own declaration regardless of the referencing file. tsz must do the same: an interface body is a pure function of its own declaration. Owner: checker cross-arena interface body computation (`compute_type_of_symbol` INTERFACE block).

## Fix

In the INTERFACE block of `compute_type_of_symbol`, when the local lowering produced no object shape (would otherwise be `error`) and the declaration is genuinely out-of-arena, scan the loaded arenas for the one whose `NodeIndex` resolves to this interface declaration, lower it there, and merge its `extends`-bases from that same source arena. Heritage base type symbols are resolved in the declaring file's binder scope (new `resolve_type_symbol_in_arena_scope`) so file-scoped bases (zod `ZodTypeDef`) inherit correctly instead of producing a weak own-members-only shape (which surfaced a transient TS2559).

Anti-hardcoding: the recovery matches structurally on "this index resolves to an interface node in this arena", never on a name or file string. It fires only on the degenerate miss path that would otherwise mint `error` (guarded by `object_shape(interface_type).is_none()`), so interfaces that already lower correctly are untouched. Bounded by the loaded-arena count.

The per-arena heritage-merge body is extracted into a shared `merge_interface_heritage_from_arena` consumed by both the existing `merge_cross_file_heritage` and the new recovery scan, so both observe identical heritage facts.

## Verification

zod guard (pinned ref `93b0b689`, `tsz_write_zod_config`, `RAYON_NUM_THREADS=1`), A/B against this branch's merge base:

- Diagnostics: **51 -> 37** (−14). The 3 `{ checks: error[] }` TS2345 witnesses at `src/types.ts(520)/(762)/(777)` clear, and their downstream TS2345 cascade (20 -> 5) clears. No new diagnostic families; the transient TS2559 introduced by the partial fix is resolved by the heritage merge.
- Performance: **1.92s -> 1.70s user CPU (−11.2%, faster)** — the interface resolves and caches once instead of re-deriving `error` through the property-access fallback on every reference (interleaved 6-run A/B, clean baseline binary).

Adjacent-case matrix (new test module `cross_file_base_type_param_interface_member_tests`, 3 tests, binder names varied to prove structural behavior; PASS on this branch):
- `derived_reads_base_type_param_bound_to_cross_file_interface` — the core witness shape.
- `renamed_derived_reads_base_type_param_bound_to_cross_file_interface` — renamed binders/members.
- `cross_file_interface_recovery_preserves_extends_base_members` — pins the heritage-merge half (weak-base TS2559 guard).

Conformance delta (local, narrow `-p tsz-checker` filter): the 10 pre-existing `cross_file` local failures (jsdoc/jsx-react/lib-name/zod-partial-omit) fail identically on the merge base and on this branch — **zero new failures introduced**. Full conformance runs in CI.

Note: the two-file unit harness (`check_multi_file`) does not reproduce the real arena split (same pre-existing harness gap documented in `cross_file_type_param_decl_identity_tests`), so the included tests are regression guards; the measured flip is the zod canary row above.

## Provenance

Machine: studio
Assistant: claude-code
Model: claude-opus-4-8
Effort: high

### DIFF
--- a/crates/tsz-checker/src/lib.rs
+++ b/crates/tsz-checker/src/lib.rs
@@ -691,6 +691,9 @@ mod contextual_return_wrapper_tests;
 #[path = "../tests/contextual_typing_tests.rs"]
 mod contextual_typing_tests;
 #[cfg(test)]
+#[path = "tests/cross_file_base_type_param_interface_member_tests.rs"]
+mod cross_file_base_type_param_interface_member_tests;
+#[cfg(test)]
 #[path = "../tests/cross_file_class_merge_tests.rs"]
 mod cross_file_class_merge_tests;
 #[cfg(test)]
--- a/crates/tsz-checker/src/state/type_analysis/computed/mod.rs
+++ b/crates/tsz-checker/src/state/type_analysis/computed/mod.rs
@@ -1868,6 +1868,7 @@ impl<'a> CheckerState<'a> {
                 //  - Different NodeIndex (has_out_of_arena_decl): decl not in local arena
                 //  - Same NodeIndex collision (has_cross_file_same_index): decl IS in
                 //    local arena, but declaration_arenas has additional non-local arenas
+                let mut merged_cross_arena_decl = false;
                 if has_out_of_arena_decl || has_cross_file_same_index {
                     for &decl_idx in declarations.iter() {
                         let Some(arenas) =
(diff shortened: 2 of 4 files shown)


mohsen1:
Closing as superseded by #13614 (merged 2026-06-15T08:15Z), which fixed the same zod `{ checks: error[] }` root cause upstream while this was in flight.

**Verification this is a true no-op now (not a guess):** rebased onto current main and A/B'd a clean baseline binary (current main, no change) vs main+this-change on the pinned zod fixture — **byte-identical diagnostic output (30 == 30, `diff` empty)**. The three regression tests added here also PASS on current main *without* this change. So the recovery scan added here never fires anymore.

**Why:** this PR was a *downstream* repair — when an interface degraded to `error` because its declaration node didn't resolve in the importing arena (`has_out_of_arena_decl=true`, no `declaration_arenas` bridge), it scanned loaded arenas to recover the body. #13614 fixes it *upstream* at the delegation guard: it keeps locally-declared interfaces off cross-arena delegation entirely (the re-export-barrel + namespace-import case pointed `resolve_symbol_file_index` at the re-exporter), so `ZodStringDef` never degrades to `error` and `has_out_of_arena_decl` never becomes true for it. #13614's mechanism is cleaner and strictly earlier.

**Preserved trace (for anyone hitting a residual of this family):** the zod witness mint was pinned to `get_type_of_symbol(ZodStringDef)` returning `TypeId::ERROR` with resolution stack `[ZodStringDef]` only (NO cycle), routed through the interface block's lib/local-lowering path with `has_out_of_arena_decl=true` / `has_local_interface_decl=false` and declaration `NodeIndex` 1499 resolving to an interface node only in another loaded arena. That is exactly the delegation mis-routing #13614 addresses. Confirmed distinct from the kysely free-type-var-against-in-progress-class hypothesis (no free vars, no cycle) — see my note on #13044.

No code from this branch is needed. Closing without merge.

Machine: studio
Assistant: claude-code

PR #14012, opened 2026-06-19
fix(checker): render apparent type of receiver in TS7053 (object -> {})

### LINKED ISSUE #14008: checker: TS7053 renders raw receiver type instead of apparent type (object vs {})
## Summary

`TS7053` renders the **raw** receiver type where `tsc` renders its **apparent
type**. For the `object` intrinsic the apparent type is `{}`, so:

- `tsc`:  `... can't be used to index type '{}'.`
- `tsz`:  `... can't be used to index type 'object'.`

Both compilers correctly emit `TS7053` at the same location — only the rendered
receiver type differs. Because the project-compile / conformance line diff
includes the message text, this counts as a `tsz`-only delta.

## Minimal repro (no narrowing involved)

```ts
function h(o: object, k: string) {
  return o[k]; // TS7053
}
```

- `tsc`:  `Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.`
- `tsz`:  `... can't be used to index type 'object'.`

## Structural rule

When element access fails the implicit-any index check, `tsc` displays
`typeToString(getApparentType(objectType))`. The apparent type of the `object`
intrinsic is the empty object type `{}`. `tsz` must render the apparent type of
the receiver in the `TS7053` message, not the raw type.

## Owner layer

Checker diagnostic rendering: `crates/tsz-checker/src/error_reporter/properties/diagnostic_methods_tail.rs:689-694`
(`object_str` via `property_receiver_display_for_node`). Apparent-type machinery
exists at `crates/tsz-solver/src/relations/judge.rs:652` (`apparent_type`).
Caution: `property_receiver_display_for_node` is a large shared path with many
`this`/generic/annotation special cases, and `TS7053` rendering is shared with
conformance — any change must be gated against the full conformance suite (this
is why it is filed rather than hot-patched: ~2 canary deltas vs shared-display
regression risk).

## Canary impact

superstruct: `src/structs/types.ts(377,21)`, `src/utils.ts(9,32)` (the receiver
is `object`, e.g. after a `for-in` over an `isObject`-guarded value). Other
superstruct `TS7053` lines (types.ts 310/314/337/502) already render `{}` in
`tsz` because their receivers are already empty-object-typed — confirming the
divergence is specific to the `object` intrinsic's apparent-type display.

Goal: green

mohsen1:
## Summary

`tsc` renders the **apparent type** of the indexed receiver in the implicit-any
element-access diagnostic (`TS7053`), but `tsz` rendered the raw type. Both
compilers emit `TS7053` at the same location; only the message text diverged:

```ts
function h(o: object, k: string) { return o[k]; } // TS7053
```
- `tsc`:  `... can't be used to index type '{}'.`
- `tsz` (before): `... can't be used to index type 'object'.`

The apparent type widens the `object` intrinsic to the empty object type `{}`
and each primitive to its global wrapper interface (`number` -> `Number`,
`boolean` -> `Boolean`, `symbol` -> `Symbol`, `bigint` -> `BigInt`,
`string` -> `String`). For a type-parameter receiver the apparent type is its
constraint's apparent type (e.g. `T extends number` -> `Number`).

## Structural rule

When element access fails the implicit-any index check, `tsc` displays
`typeToString(getApparentType(receiver))`. The destructuring `TS2538`/`TS2339`
path already applied exactly this widening via a local helper
(`apparent_type_display_for_destructuring`); the element-access `TS7053` path was
written separately and never did. Promote that helper to a shared
`apparent_intrinsic_type_display` (`error_reporter::properties`) and apply it to
the element-access `object_str` as well. Other receivers return `None` and fall
back to the existing formatter, so the behavior change is confined to the six
intrinsics — and since `tsc` always renders the apparent type here, matching it
can only converge tsz toward `tsc`.

## Owner layer

Checker diagnostic rendering. Shared helper:
`crates/tsz-checker/src/error_reporter/properties.rs`. Applied at the TS7053
emission in `error_reporter/properties/diagnostic_methods_tail.rs`; the
destructuring site (`state/variable_checking/destructuring.rs`) now reuses the
same helper (DRY, behavior-preserving). No type semantics changed.

## Adjacent cases

- Receivers `object`/`number`/`boolean`/`symbol`/`bigint`/`string`: render
  apparent type (now match tsc).
- Type-parameter receivers `T extends <intrinsic>`: render the constraint's
  apparent type (match tsc).
- Unconstrained `T` -> `unknown`, `{}` -> `{}`, object/union/etc.: unchanged.
- Destructuring TS2538/TS2339: unchanged (same helper, same mapping).

Goal: green

## Provenance
Machine: m1
Assistant: claude-code
Model: claude-opus-4-8
Effort: max

## Verification

- Minimal repros match `tsc` byte-for-byte (`diff` empty) for all receiver kinds
  (`object`/`number`/`boolean`/`symbol`/`bigint`/`{}`/`unknown`) and type-param
  constraints (`T extends number|object|symbol`, unconstrained `T`, `bigint`).
- superstruct canary: 16 -> 14 tsz-only FPs (`comm -23` vs `tsc` oracle);
  exactly `types.ts(377,21)` + `utils.ts(9,32)` removed, zero new lines.
- kysely / msw canaries: zero new lines (no regression).
- Conformance (prebuilt dist-fast, all 100%): noImplicitAny 37/37,
  indexedAccess 13/13, indexSignature 24/24, destructuring 174/174, keyof 14/14,
  member 261/261, mappedType 55/55, computedProperties 149/149, generics 13/13,
  objectTypes 49/49, indexer 13/13, expressions 377/377, narrowing 29/29,
  controlFlow 94/94.
- `cargo clippy -p tsz-checker` clean; `cargo fmt -p tsz-checker --check` clean.

Resolves #14008.

### DIFF
--- a/crates/tsz-checker/src/error_reporter/apparent_type_display.rs
+++ b/crates/tsz-checker/src/error_reporter/apparent_type_display.rs
@@ -1,0 +1,22 @@
+//! Shared apparent-type display strings for index/destructuring diagnostics.
+
+use tsz_solver::TypeId;
+
+/// Display string for the *apparent* type of a primitive/`object` intrinsic, as
+/// used in implicit-any index and destructuring diagnostics (`TS7053`/`TS2538`/
+/// `TS2339`). `tsc` reports `typeToString(getApparentType(receiver))` for these
+/// failures, so the `object` intrinsic widens to the empty object type `{}` and
+/// each primitive widens to its global wrapper interface (`number` -> `Number`,
+/// etc.). Returns `None` for any other type so callers fall back to the normal
+/// type formatter.
+pub(crate) fn apparent_intrinsic_type_display(type_id: TypeId) -> Option<String> {
+    match type_id {
+        TypeId::OBJECT => Some("{}".to_string()),
+        TypeId::STRING => Some("String".to_string()),
+        TypeId::NUMBER => Some("Number".to_string()),
+        TypeId::BOOLEAN => Some("Boolean".to_string()),
+        TypeId::BIGINT => Some("BigInt".to_string()),
+        TypeId::SYMBOL => Some("Symbol".to_string()),
+        _ => None,
+    }
+}
--- a/crates/tsz-checker/src/error_reporter/mod.rs
+++ b/crates/tsz-checker/src/error_reporter/mod.rs
@@ -9,6 +9,7 @@ pub(crate) enum TypeOnlyKind {
 }
 
 // Submodules
+pub(crate) mod apparent_type_display;
 pub(crate) mod assignability;
 mod assignability_alias_display;
 mod assignability_anchor_helpers;
@@ -42,7 +43,7 @@ mod name_resolution;
 mod noinfer_diagnostic_display;
 mod operator_errors;
 mod primitive_intersection_display;
-pub(crate) mod properties;
+mod properties;
 mod property_receiver_formatting;
 mod recursive_alias_display;
 mod render_failure;
--- a/crates/tsz-checker/src/error_reporter/mod.rs
+++ b/crates/tsz-checker/src/error_reporter/mod.rs
@@ -42,7 +42,7 @@ mod name_resolution;
 mod noinfer_diagnostic_display;
 mod operator_errors;
 mod primitive_intersection_display;
-mod properties;
+pub(crate) mod properties;
 mod property_receiver_formatting;
 mod recursive_alias_display;
 mod render_failure;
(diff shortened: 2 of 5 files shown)


mohsen1:
Superseded by #14015, which implements the same TS7053 apparent-type fix (`object` -> `{}`, primitives -> wrapper interfaces) more thoroughly — solver-level apparent-type machinery (`tsz-solver/src/objects/apparent.rs`) plus dedicated tests — and is currently in the merge queue. Both touch the same `diagnostic_methods_tail.rs` region, so they can't both land. Dequeuing this one to avoid the conflict; will close once #14015 merges. (This PR was the smaller DRY-reuse approach: sharing the existing destructuring `apparent_intrinsic_type_display` helper.)

Issue #14232, opened 2026-06-20
false-positive(solver): TS2322 — conditional with concrete check vs generic extends defers instead of resolving the false branch (ts-essentials)


mohsen1:
## Summary
Mined from **ts-essentials**. tsz leaves a conditional unresolved (→ TS2322) where tsc resolves the false branch:
```ts
function f<T>() {
  type A = [] extends [T, ...T[]] ? "yes" : "no";
  const a: A = "no";   // tsc: A = "no" ; tsz: TS2322 ("no" not assignable to A)
}
export {};
```

## Structural rule
When a conditional's relation fails (check NOT assignable to extends) and the **extends** type is generic but the **check** type is concrete, tsc resolves eagerly: take the false branch iff the relation also fails under the permissive instantiation (all type params → `any`), else defer. tsz defers unconditionally whenever the extends type carries a type parameter, leaving the conditional unresolved so `A` is opaque and `"no"` doesn't match.

## Owner / fix direction
Solver — `crates/tsz-solver/src/.../conditional.rs` `evaluate_conditional` (~987). When the check is concrete and the relation fails, re-test under the permissive (params→any) instantiation; if it still fails, take the false branch instead of deferring.

## Adjacent cases
`[] extends [T, ...T[]]` (the bug); concrete check vs generic extends both true and false; check carrying type params (must still defer); distributive vs non-distributive.

## Verification
Repro above: tsz TS2322, tsc 0. Anti-hardcoding: structural conditional-resolution rule.

Goal: green

PR #14364, opened 2026-06-21
fix(solver): resolve concrete-check conditional with a generic extends type via permissive instantiation (TS2322) (#14232)

mohsen1:
Fixes a green-goal false positive (TS2322). Contained change to the conditional
result-branch decision: it narrows *when* `tsz` defers a conditional, matching
tsc's `getConditionalType` permissive-instantiation gate.

Goal: green

## Structural rule
In `evaluate_conditional`, when the EXTENDS type carries a type parameter
(`extends_has_type_params`), `tsz` unconditionally deferred the conditional —
even when the CHECK type is concrete. So `[] extends [T, ...T[]] ? "yes" : "no"`
stayed deferred, the alias `A` resolved to an opaque/`unknown`-like type, and a
later `const a: A = "no"` produced a false TS2322.

tsc (`getConditionalType`) only keeps such a conditional deferred when either the
check type is *itself* generic, or the false branch is *not* definitive under the
permissive instantiation (every type parameter replaced by `any`,
`getPermissiveInstantiation`). For a concrete check type whose false branch is
definitive — `[]` is never `[any, ...any[]]` — the conditional resolves to the
false branch.

This change adds that gate: defer only when
`extends_has_type_params && (is_generic_conditional_check_type(check) || !permissive_false_branch_is_definitive(check, extends))`,
preserving the existing deferred-wrapper / depth-detection tail unchanged.

Owner: solver — `evaluation/evaluate_rules/conditional.rs`.

## Adjacent cases (test file)
- Witness: `[] extends [T, ...T[]] ? "yes" : "no"` resolves to `"no"`;
  `const a: A = "no"` → no TS2322 (flip).
- Negative control: assigning the *true*-branch literal `"yes"` to `A` (which is
  `"no"`) still emits TS2322 — the fix resolves the branch, it does not
  blanket-suppress assignability.

## Verification
- `cargo nextest run -p tsz-checker -j 2 conditional_concrete_check_generic_extends`
  — flip + negative control green (verified with only `conditional.rs` changed).
- CI gates: lint, unit-checker-integration, conformance-aggregate, emit-aggregate,
  fourslash-*, project-compile-guard, CI Summary.

## Provenance
Machine: Mac.fritz.box
Assistant: claude-code
Model: claude-opus-4-8
Effort: max

### DIFF
--- a/crates/tsz-checker/tests/conformance_issues/types/conditional_concrete_check_generic_extends_2322.rs
+++ b/crates/tsz-checker/tests/conformance_issues/types/conditional_concrete_check_generic_extends_2322.rs
@@ -1,0 +1,52 @@
+//! A conditional type whose CHECK type is concrete (`[]`) but whose EXTENDS type
+//! carries a type parameter (`[T, ...T[]]`) must not be left deferred when the
+//! false branch is definitive under permissive instantiation (T -> any). tsc
+//! resolves `[] extends [T, ...T[]] ? "yes" : "no"` to its false branch (`"no"`);
+//! tsz previously deferred (because the extends type carried `T`), leaving the
+//! alias opaque so a later `"no"` assignment produced a false TS2322. (#14232)
+
+use super::super::core::*;
+
+/// The witness: `A = [] extends [T, ...T[]] ? "yes" : "no"` resolves to `"no"`
+/// (an empty tuple is never a non-empty tuple, even under `T -> any`), so
+/// `const a: A = "no"` type-checks with no TS2322.
+#[test]
+fn concrete_check_generic_extends_resolves_false_branch_no_ts2322() {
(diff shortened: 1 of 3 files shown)


mohsen1:
Closing: the fix flips the #14232 witness but regresses `contextuallyTyped*` conformance — the permissive-instantiation gate in `evaluate_conditional` is too broad (it now resolves a concrete-check conditional's false branch in cases where tsc keeps it deferred for contextual typing). Narrowing it needs the exact regressed fixture, which requires the TypeScript submodule + CI-iteration not available this session. Finding preserved for follow-up: the gate must additionally exclude contextually-typed conditional contexts. Witness: `[] extends [T, ...T[]] ? 'yes' : 'no'` should resolve to `'no'` (false TS2322 today).

PR #14431, opened 2026-06-21
fix(solver): resolve deferred conditional check-type constraint so spread of `P extends Parameters<F>` is variadic (#14217)

### LINKED ISSUE #14217: false-positive(checker): TS2345 — spread of a type param constrained by Parameters<F> not recognized as a rest-spread (true-myth)
## Summary
Mined from **true-myth** (curried apply over `Parameters<F>`). Strict, dependency-free repro — tsz emits TS2345, tsc clean:
```ts
type AnyFunction = (...params: never[]) => unknown;
export function callIt<F extends AnyFunction, P extends Parameters<F>>(fn: F, params: P): unknown {
  return fn(...params);   // tsz: (3,13) TS2345 'any' not assignable to 'never'. tsc: clean.
}
```
`tsc -p` → 0; `tsz -p` → TS2345. The indirection is essential: `params: Parameters<F>` *directly* (no intermediate `P`) is clean in both.

## Structural rule
When a spread argument `f(...p)` has `p` of a type-parameter type `P` whose constraint's **base constraint** is array/tuple-like — including a deferred conditional utility like `Parameters<F>` (= `F extends (...args: infer Q) => any ? Q : never`, whose rest-position `infer Q` carries an implicit array constraint) — tsc treats the spread as landing on a rest parameter and wraps it as a variadic `[...P]` rather than materializing a single representative element. tsz's spread-of-type-parameter gate only recognizes **direct** array/tuple constraints; for a deferred-conditional constraint it materializes the spread into a representative `any`, which then fails the callee's `never[]` rest parameter (false TS2345).

## Owner / fix direction
Checker call path — `crates/tsz-checker/src/checkers/call_checker/candidate_collection.rs` (the spread-of-type-parameter marker-wrapping gate, ~line 679, currently `array_element_type_for_type(constraint) || tuple_elements_for_type(constraint)`). Extend the gate to recognize a constraint whose **base constraint** resolves to array/tuple-like (new `base_constraint_is_array_or_tuple` helper in `query_boundaries/common.rs` → `tsz-solver/src/type_queries/data/content_predicates.rs`), so the deferred `Parameters<F>` case is wrapped as `[...P]`.

## Adjacent cases (verified vs tsc)
- `P extends Parameters<F>`, `P extends [...Parameters<F>]`, and a user conditional `P extends (T extends ... ? unknown[] : never)`.
- Negative control: `P extends string` spread must still error (not array-like).
- Direct `params: Parameters<F>` (no `P`) already clean — confirms the type-parameter indirection is the trigger.

## Verification
Repro above: `tsz` TS2345, `tsc` 0. Anti-hardcoding: structural base-constraint shape check.

Goal: green

mohsen1:
Fixes #14217.

Goal: green

## Structural rule
When a spread argument `f(...p)` has `p` of a type-parameter type `P` whose
constraint is a *deferred* conditional whose **base constraint** is array/tuple-like
— including the lib utility `Parameters<F>` (`F extends (...args: infer P) => any ? P : never`,
whose rest-position `infer P` carries an implicit array shape) — `tsc` treats the
spread as landing on a rest parameter and wraps it as a variadic `[...P]`. `tsz`
only recognized **direct** array/tuple constraints; for a deferred-conditional
constraint it materialized a representative `any`, which then failed the callee's
`never[]` rest parameter (false **TS2345** `'any' is not assignable to 'never'`).

The missing primitive was tsc's `getConstraintFromConditionalType`: for a conditional
whose check type is a type parameter `F` with constraint `C`, substitute `F → C` and
re-instantiate; the now-concrete check type lets the evaluator infer `P` and select a
branch (`Parameters<F>` → `never[]`). The shallow `get_base_constraint_of_type` and the
branch-union `getDefaultConstraintOfConditionalType` both leave the rest-position
`infer P` unresolved, so neither recognized the array base.

## Owner layer
Solver (type semantics). New solver query
`conditional_check_type_substituted_constraint` (construction half of
`getConstraintFromConditionalType`) in `type_queries`, plus
`SubtypeChecker::infer_extraction_conditional_constraint` for the relation. The
substitution is gated on an `infer` in the extends type + a type-parameter check
type, so predicate conditionals (`T extends unknown[] ? true : false`) keep the
existing determinism-guarded distributive path and are not narrowed unsoundly.

The single reduction is routed through four sites that all need
`Parameters<F> → never[]`:
- spread-of-type-parameter gate (`candidate_collection.rs`) — wrap `[...P]`;
- spread iterability check (`iterable_checker.rs`) — reuses the shared gate method;
- conditional subtype relation (`conditionals.rs`, new Strategy 1.25);
- tuple-to-array subtype rule (`tuples.rs`) — `[...Parameters<F>] <: never[]`.

The checker drives resolver-backed evaluation (`evaluate_type_with_env`) because an
imported alias like `Parameters` is not resolvable by the solver's resolver-less
`evaluate_type`; the structural decisions stay in solver query boundaries.

## Adjacent-case matrix (all verified against `tsc` 6.0.2)
- `P extends Parameters<F>` (the repro) — clean.
- `P extends [...Parameters<F>]` — clean (tuple-to-array rest).
- user conditional `P extends (T extends number ? unknown[] : never)` — clean (iterability + relation).
- renamed binders (`G`/`Q`) — clean (structural, not name-based).
- **Negative control** `P extends string` — still errors (TS2345), matching `tsc`.
- predicate conditional `IsArray<T>` assignment — still errors, unchanged.

## Verification
- `cargo test -p tsz-checker --lib -- call_spread_constructor_parameters_tests` (4 new + 1 existing, all green).
- `cargo test -p tsz-solver --lib -- relations::subtype::rules` green.
- CLI repro + adjacent matrix diffed against `tsc` 6.0.2: tsz matches (clean where tsc clean, errors where tsc errors).
- `ts-pattern` project: no new diagnostics introduced by this change.
- The 5 pre-existing solver-lib test failures and 1 checker test failure on this branch's base commit are unchanged (confirmed against `HEAD~1`).
- Anti-hardcoding: structural base-constraint shape check; no identifier/file-name/printer-string predicates; tests vary binder names.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-4-8
Effort: high


mohsen1:
Provenance: MacBookPro / codex / gpt-5 / high

Follow-up root cause: PR-head lint next failed the architecture guard because the new conditional wrapper pushed `crates/tsz-checker/src/query_boundaries/common.rs` above the #8225 quarantine line limit.

Files changed: moved the new conditional substituted-constraint wrapper into focused `query_boundaries::conditional`, registered the module, and updated the call site. The solver still owns the semantic query; checker only calls the boundary and evaluates/probes the result.

Verification: `cargo fmt --check`, `python3 scripts/arch/arch_guard.py`, and `scripts/safe-run.sh cargo clippy -p tsz-checker --tests -- -D warnings` passed locally. New PR-head CI is running at head `0dff428e1a`.

Remaining risk: full PR-head CI still needs to pass before requeueing.

Issue #14345, opened 2026-06-21
[arch] No canonical materialize-once (SCC fixpoint) for recursive cross-arena generic heritage — cycle leaves an opaque deferred Application


mohsen1:
> [!IMPORTANT]
> ## Status update — 2026-07-01
>
> The framing below ("no SCC materialize-once" as the root) has been **partially refuted by this issue's own 07-01 probe chain** — read the last five comments plus [the three-seat direction spec](https://github.com/tsz-org/tsz/issues/14345#issuecomment-4859770811) before starting work:
>
> - **Refuted:** per-arena-copy identity (0/28,011 collapse) · co-inductive/repeated-state guards (blowup is *productive tail-position growth*; nothing repeats in-flight) · Mode A reduction-side index deferral (growth driver is the object-side arg; gate structurally unsound) · OPTION B materialize-at-index as a diagnostic fix (base materializes 1.40M× yet **net-0** — the FP bulk is minted at the inference no-candidate fallback, `normalization.rs` → `unknown`).
> - **Live frontier — three seats:** (1) **inference matching** on deferred pairs (tsc `inferFromTypes` IndexedAccess arm; generalizes #15220's +11) — *measure the inference-vs-relation split of the residual family first; that number picks the bulk seat*; (2) **relation layer** — the pairwise visitor rule already exists (`visitor.rs:996-999`); the gaps are the pre-evaluation seat (both sides eagerly evaluated at `cache.rs` ~1221 before dispatch) and the literal-only constraint fallback; type-param identity may come only from the relation-local alpha frame or decl identity — never name or position; (3) **containment** — recursion-identity (`getRecursionIdentity`/`isDeeplyNestedType` analog) at the *evaluator* keyof/index re-reduce sites; this is the #14346-owned lever with required-row payoff.
> - **Mode B** (registry wall): bounded M-effort priming slice through the merged-but-OFF publish channels; substrate + concrete-URI subfamily only — **not** the fp-ts row lever.
> - Gauges: full flag-ON conformance + soundness oracle (fp-ts is tsc-clean; error-count-down is not a soundness signal); `TSZ_DETERMINISTIC_STORE_ELECTION=1`; one lever per gauge (#15317).
>
> *Status by FableArch · Mac.fritz.box · claude-code · claude-fable-5 · max — original body preserved unchanged below.*

---

> **Severity:** critical · **Impact:** high · **Effort:** XL · **Class:** ROOT SIN
[...]

Issue #14344, opened 2026-06-21
[arch] Identity is modeled at the allocation-order / arena-local level instead of the content-canonical level (the universal cache key is wrong)


mohsen1:
> [!IMPORTANT]
> ## Status update — 2026-07-01
>
> The identity campaign's original mechanism is **gone from main** and one core premise below is refuted; the evidence record below remains valid, but do not resume from its "proposed refactor" as written:
>
> - **Content-election reverted:** #14544/#14546 (`TSZ_CANONICAL_DEFID`) were measured non-functional (`election_links_added=0` structurally; the #13862 witness was already fixed by #13884) and reverted in #14558. The flag is read by no crate; `scripts/bench/canonical-defid-harness` measures a no-op (cleanup tracked in #15317). Activation tracker #14557 recommended closed.
> - **Per-arena-copy premise refuted for the shared store:** `canonical_decl_site_def_id` collapses **0/28,011** defs — the shared `DefinitionStore` already mints one `DefId` per source decl. Cross-arena copy multiplicity is not the convergence lever (07-01 probe, see #14345).
> - **What actually landed:** #15220 `DelegatingHktResolver` inference lever (+11 fp-ts unknown-drop clears, deterministic RAYON 1/4) and #15223/#15234 deterministic `(file_id, span_start, symbol_id)` store election (~51–55 clears). The residual fp-ts family (~720 lines) is minted at the **inference no-candidate fallback**, not at identity resolution — frontier spec in [this #14345 comment](https://github.com/tsz-org/tsz/issues/14345#issuecomment-4859770811).
> - **The 06-27 verdict stands:** the full content-canonical-identity flip (`TSZ_TYPEPARAM_DECL_IDENTITY` default-ON) is walled at every layer; 10 bounded levers refuted; it remains XL with no bounded onramp. The 13 dormant flags this campaign banked now need the #15317 harness/ledger before any further flag lands.
>
> *Status by FableArch · Mac.fritz.box · claude-code · claude-fable-5 · max — original body preserved unchanged below.*

---

> **Severity:** critical · **Impact:** high · **Effort:** XL · **Class:** ROOT SIN
> Part of the architectural root-cause review — see the tracking issue.

## Symptom & blast radius
[...]

PR #14825, opened 2026-06-25
fix(checker): bare expression-statement never-call body infers never, not void (TS2322)

mohsen1:
Goal: green

## Provenance
Machine: Mac
Assistant: claude-code
Model: claude-opus-4-8
Effort: high

## Summary
Structural rule: a function body that is a bare expression-statement calling a never-returning function does not fall through, so its inferred return type is `never` (like `throw` / `return die()`); tsz was defaulting to `void`, producing a spurious TS2322 when the body was assigned to a non-`void` function type.

Witness: immer `proxy.ts` 241/247 (`defineProperty() { die(11) }`).

```ts
declare function die(n: number): never
const a: () => boolean = () => { throw new Error() }  // OK (infers never)
const b: () => boolean = () => { die(1) }             // was TS2322 — now OK (infers never)
const c: () => boolean = () => { return die(1) }      // OK (infers never)
```

Owner: checker return-type inference (`crates/tsz-checker/src/types/utilities/return_type.rs`).

The inferred-return path gates the (expensive) full fall-through analysis behind a cheap pre-check, `body_contains_throw_or_never_call`. That pre-check deliberately ignored expression statements (it was AST-only and never-call detection needs type checking), so the inference short-circuited to `void` (return_type.rs ~line 1070/1083) and never ran `function_body_falls_through`, which via `reachability_checker.rs` `call_expression_terminates_control_flow` would have correctly classified the never-call as terminating.

Fix: the pre-check now consults the existing `call_expression_terminates_control_flow` classifier (the same one the full reachability analysis uses) for expression statements, instead of duplicating never-detection logic. A terminating never-call is recognized, so the body falls through to `function_body_falls_through` → `never`. Anti-hardcoding: structural, reuses the reachability/never classifier; no identifier/file-name string predicates.

Negative controls preserved:
- A body that genuinely falls through (non-never call, e.g. `log(1)`, or empty `{}`) still infers `void` (TS2322 vs a non-`void` type still fires).
- Return-completeness diagnostics (TS2355/TS2366) still fire where a branch falls through (e.g. `if (flag) { die(1) }` with no else).
- A mix where every branch terminates (`if (flag) return 1; die(2)`) emits no return-completeness diagnostic.
- `throw` and `return die()` forms unchanged.

Moderate blast radius (return-inference) — the full return-inference regression delta runs in ready CI.

## Verification
New module `crates/tsz-checker/tests/expr_statement_never_call_return_inference_tests.rs` (registered in `lib.rs`), 9 tests, all PASS:
- `bare_never_call_body_infers_never_no_ts2322` (the repro `b`)
- `throw_body_infers_never_no_ts2322`, `return_never_call_body_infers_never_no_ts2322` (controls `a`/`c`)
- `bare_never_call_body_renamed_binders_no_ts2322` (binder-name independence)
- `fall_through_body_still_infers_void_emits_ts2322`, `empty_body_still_infers_void_emits_ts2322` (negative controls → still `void`)
- `mixed_branch_annotated_body_still_emits_return_completeness` (TS2355/TS2366 still fires)
- `mixed_branch_value_return_and_never_call_no_return_completeness` (no false TS2355/TS2366)
- `annotated_bare_never_call_body_no_return_completeness`

Targeted local sweep: `cargo nextest run -p tsz-checker --lib -E 'test(return_type) + test(reachab) + test(never) + test(falls_through) + test(fall_through) + test(void) + test(terminal_return) + test(implicit_return) + test(ts2355) + test(ts2366) + test(ts7030) + test(control_flow)'` → 498 pass; the only 4 reds (`test_array_helpers_avoid_direct_typekey_interning`, `imported_generator_return_type_is_iterable_in_for_of`, `test_ts2322_async_return_via_return_type_promise_infer`, `yield_star_generator_callback_mismatch_reports_outer_ts2345`) were confirmed PRE-EXISTING on clean origin/main (verified in an isolated worktree), unrelated to this change.

`cargo clippy -p tsz-checker --lib` clean. Full return-inference regression delta runs in ready CI given the moderate blast radius.


mohsen1:
Reverting + closing: the never-call return-inference change perturbs a SHARED evaluation path (materializes the evolving-array receiver type during return inference), regressing controlFlowArrayErrors.ts (loses TS7034/TS7005 implicit-any). High blast radius + witness-blind locally (only CI shows the delta). Protecting the HOLD conformance floor over one green FP; deferring the immer never-call FP to a narrower future approach / the #14344-#14345 materialization work.

Issue #15338, opened 2026-07-01
[hold] main fails 9 tests locally while CI is disabled (3 solver, 3 checker hangs, 3 cli) — attribution + evidence


mohsen1:
## Main is failing 9 tests locally while CI is disabled — attribution + evidence

With PR CI almost fully disabled (every check on current PRs reports `skipping`), full local suite runs against **pristine `origin/main`** (baseline worktree, verified twice — once via the full per-crate suites, once via scoped single-binary reruns) show 9 failures that are landing/have landed unnoticed:

### tsz-solver (3)

1. `relation_cache_config_tests::relation_cache_config_does_not_expose_raw_constructor` — panics: "types.rs must keep RelationCacheConfig impl" (`tests/relation_cache_config_tests.rs:115`). The `RelationCacheConfig` impl apparently moved out of `types.rs` without updating this source-scan.
2. `solver_file_size_ceiling_tests::test_solver_oversized_file_size_ceilings` — `src/evaluation/evaluate_rules/mapped.rs` grew to **1606 lines (frozen baseline: 1504)** without `TSZ_FILE_SIZE_RATCHET_UPDATE=1` + committed baseline diff.
3. `visitors::visitor_predicates::content::tests::free_infer_policy_skips_generic_signature_bodies` — behavioral assert: `contains_free_infer_types(&interner, plain_fn)` now false (`visitor_predicates/content.rs:908`). A semantic predicate changed under the test.

### tsz-checker (3 — hangs, not asserts)

4-6. `union_origin_display_tests::{ts2859_full_fixture_reports_target_union_site, ts2859_message_preserves_top_level_alias_in_target_union, ts2859_constituent_intersection_rule_survives_renamed_aliases}` — all three **TIMEOUT at 60s** (hang, reproducible scoped and in the full suite). A display-path hang is the #14346 "termination is emergent" class showing up on main.

### tsz-cli (3 — possibly environment-sensitive; verify on a second machine)

7. `tsc_compat_tests::tsc_parity_help_all`
8. `tsc_compat_tests::tsc_parity_ts6046_module_resolution`
9. `bin/tsz-server handlers_code_fixes::tests::collect_import_candidates_respects_node_next_package_exports_root_only`

### Notes

[...]

Issue #14351, opened 2026-06-21
[arch] Root-cause review: the core sins blocking tsz green + fast (identity → SCC materialize-once → termination)


mohsen1:
> [!IMPORTANT]
> ## Status update — 2026-07-01 (post-refutation-week consolidation)
>
> Several load-bearing framings of the review below were **measured false** between 06-22 and 07-01. Read [the consolidated direction comment](https://github.com/tsz-org/tsz/issues/14351#issuecomment-4859771262) before picking up work from this body. The short map:
>
> - **#14344 (identity):** content-election built → measured non-functional → **reverted** (#14558); per-arena-copy premise refuted (`canonical_decl_site` collapses 0/28,011). Landed wins took a different route: #15220 (inference lever, +11) and #15223/#15234 (deterministic store election). #14557 (activation) recommended closed.
> - **#14345 (materialize-once):** co-inductive/repeated-state guards gauged inert — the blowup is **productive tail-position growth**; Mode A (index deferral) blocked; OPTION B materialization = net-0 diagnostics (FP bulk is minted at the inference no-candidate fallback). Live frontier = the **three-seat deferred-form algebra** (inference matching / relation pre-eval seat / recursion-identity containment) — spec in [this #14345 comment](https://github.com/tsz-org/tsz/issues/14345#issuecomment-4859770811).
> - **#14346 (termination):** `EvaluationSession` now owns 7 state families; `EvaluationResult` is verdict-carrying. Remaining: relation-boundary session split-brain, lazy.rs fuel double-count, verdict *consumption*, and the evaluator-side recursion-identity containment (the required-row lever) — [directed here](https://github.com/tsz-org/tsz/issues/14346#issuecomment-4859770911).
> - **#14347/#14348:** cache-purity slice train merged; all dual-env writes routed through the deferred authority; collapse blocker is read visibility, not write routing — [directed here](https://github.com/tsz-org/tsz/issues/14348#issuecomment-4859771024).
> - **#14349, #14350:** CLOSED (residuals resolved on main / via #15240).
> - **New:** #15317 — dormant-flag debt (13 default-OFF campaign flags, zero CI exercise, no graduation ledger).
>
[...]

Issue #14348, opened 2026-06-21
[arch] Dual mutable TypeEnvironment synchronized by deferred-write mirroring — the one parity-safe, independently-shippable slice of the boundary-erosion cleanup


mohsen1:
> [!NOTE]
> ## Status update — 2026-07-01
>
> The write-routing half of this issue is essentially **done** on `origin/main @ 1175b9f170`; the collapse blocker has moved. Current state and next steps ([full direction](https://github.com/tsz-org/tsz/issues/14348#issuecomment-4859771024)):
>
> - **Done (#14367/#14577/#15148/#15152/#15214/#15215/#15219/#15261):** every def-body, symbol-type, class-instance/extends, and enum-metadata write routes through the deferred env-write authority; the warn-and-drop paths this body describes are gone. The flow analyzer is provably read-only on `type_environment` (zero `borrow_mut` under `flow/`).
> - **Remaining slice A (mechanical):** close the last write-loss window — `cross_file.rs:1620-1627` still silently skips the child→parent merge-back on borrow failure; dual-write the 5 eval-env-only cache sites (`lazy.rs:406/:618`, `def/resolver.rs:1678`, `assignability_checker.rs:746`, `index_signature_key_helpers.rs:38`).
> - **Slice B:** delete `overlay_missing_from` (`def/resolver.rs:1241`) once A lands — its "list every field" maintenance contract is the worst remaining divergence trap.
> - **The collapse itself is a design decision, not mechanical:** blocker = read visibility under flow's long-lived borrow (`get_type_of_symbol` epilogue publishes mid-check). Two candidate designs — (a) flow reads fall back to `type_env.try_borrow()` on miss, or (b) generation-stamped snapshots at flow-node boundaries — validated against the #13741/#13944/#13951/#13952 witnesses; watch the assignability memo stamps that read both env generations.
>
> *Status by FableArch · Mac.fritz.box · claude-code · claude-fable-5 · max — original body preserved unchanged below.*

---

> **Severity:** medium · **Impact:** medium · **Effort:** M · **Class:** symptom / cleanup (downstream of a root sin)
> Part of the architectural root-cause review — see the tracking issue.

## Symptom & blast radius
[...]

PR #15344, opened 2026-07-02
fix(checker): route the remaining eval-env-only writes through the env authority (#14348)

mohsen1:
Goal: hold

Part of #14348 (tracking: #14351). Slice A of the completion plan from the direction comment: finish the write side so the reconcile overlay can be retired.

## Structural rule

Every `TypeEnvironment` write must reach both the authoritative evaluator env (`type_env`) and the flow-analyzer env (`type_environment`) through the deferred env-write authority — deferring on borrow races, never silently skipping. tsc has one checker with one type table; tsz's dual-env design is safe only while no write path can leave the two views divergent. After the #15148/#15152/#15214/#15215/#15219/#15261 chain, four cache-seed sites still wrote only the evaluator env via raw `try_borrow_mut` (and dropped the write entirely when the env was borrowed):

- `context/resolver.rs` — unresolved-name resolution cache (new `DeferredFlowEnvWrite::InsertUnresolvedResolution`, authority-routed);
- `state/type_environment/lazy.rs` — class-expression provisional-ctor `SymbolRef` seed → `register_symbol_type_in_envs`;
- `state_checking_members/index_signature_key_helpers.rs` — TS2411 display seed → `register_symbol_type_in_envs`;
- `assignability/assignability_checker.rs` — `typeof`-value seed → `register_symbol_type_in_envs`.

`overlay_missing_from` now returns whether it repaired anything, and reconcile emits a `tsz::env_authority` warning when it fires — the measurable deletion-readiness signal for the overlay (whose "list every field" maintenance contract is the last divergence trap of the dual-env design; deletion is the follow-up slice once corpora show the warning silent).

Intentionally NOT routed: the transient `this_type` set/restore sites (flow must not observe transient `this` bindings) and the child→parent cross-file merge-back skip (deferral is structurally impossible there — snapshotting requires the borrow that failed; it keeps its warn).

## Adjacent matrix

- Uncontended write → both envs observe it immediately (new test, first half).
- Contended write (flow env borrowed) → evaluator env immediate, flow env deferred + replayed on flush (new test, second half) — previously the entire write was silently dropped.
- Overlay repair reporting: empty pair → false; eval-only entry → true then false after repair (new test).
- Existing replay/backlog/simultaneous-borrow suites (24 tests) unchanged and green.

## Verification

CI is disabled — full local gate:

- FULL `cargo nextest run -p tsz-checker` and `-p tsz-solver` on this diff vs pristine `origin/main @ 1ea86a5364`: **failure sets byte-identical** (203 unique failures on both sides — all pre-existing main breakage, catalogued in #15338; `comm` diff empty in both directions).
- FULL local conformance vs the checked-in baseline on this head: result posted in a comment below (expected: identical to main's current 12,584/12,585 — the -1 is #15331's, tracked on that PR).
- New tests: `unresolved_resolution_write_reaches_both_envs_and_defers_under_borrow`, `overlay_missing_from_reports_repairs` — the deferral half is red by construction against the old code (the old sites skipped the write under contention; the variant did not exist).
- Focused: `def_mapping`/`env_merge`/`replay`/`reconcile` filters — 24/24 + 2 new pass.
- `cargo clippy --profile ci-lint -p tsz-checker -p tsz-solver --all-targets -- -D warnings` — clean; `cargo fmt --all --check` — clean; `arch_guard.py` — 0 failures.
- Provenance note: the verified tree was recreated deterministically (exact-match scripted transformations, same base SHA) after the original working tree was lost to a worktree cleanup; the full-suite diff ran on the identical original diff, and conformance + focused suites + lint re-ran on this exact head.

## Provenance

Machine: Mac.fritz.box
Assistant: claude-code
Model: claude-fable-5
Effort: max

### DIFF
--- a/crates/tsz-checker/src/assignability/assignability_checker.rs
+++ b/crates/tsz-checker/src/assignability/assignability_checker.rs
@@ -739,10 +739,14 @@ impl<'a> CheckerState<'a> {
                 // TypeEvaluator::visit_type_query can resolve via TypeEnvironment::resolve_ref.
                 // Without this, resolve_ref returns None and the fallback resolve_lazy returns
                 // the INSTANCE type for classes, causing false TS2345 on `typeof ClassName` args.
-                if let Some(value_type) = self.ctx.symbol_types.get(&sym_id)
-                    && let Ok(mut env) = self.ctx.type_env.try_borrow_mut()
-                {
-                    env.insert(tsz_solver::SymbolRef(sym_id.0), value_type);
+                if let Some(value_type) = self.ctx.symbol_types.get(&sym_id) {
+                    // Route through the env-write authority (dual-write + defer
+                    // on borrow race instead of silently skipping; #14348).
+                    self.ctx.register_symbol_type_in_envs(
+                        tsz_solver::SymbolRef(sym_id.0),
+                        value_type,
+                        Vec::new(),
+                    );
                 }
             }
 
--- a/crates/tsz-checker/src/context/def_mapping_tests.rs
+++ b/crates/tsz-checker/src/context/def_mapping_tests.rs
@@ -1133,3 +1133,113 @@ fn resolved_def_mirror_is_deferred_under_borrow_then_envs_reconcile() {
         "envs must agree on every shared DefId -> TypeId after reconciliation"
     );
 }
+
+/// #14348: the unresolved-name resolution cache is authority-routed — it
+/// reaches BOTH environments, and a borrow race defers (never silently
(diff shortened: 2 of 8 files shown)


mohsen1:
Conformance gate on this head: **12,584/12,585 — identical to current main** (the single regression is `generatorTypeCheck63.ts`, owned by #15331, filed there before this PR). This PR introduces zero conformance delta.

Provenance: Agent: FableArch · Machine: Mac.fritz.box · Assistant: claude-code · Model: claude-fable-5 · Effort: max

Issue #16443, opened 2026-08-05
Every TS2741 whose target is a type literal loses the `'x' is declared here.` pointer — single-file too, and the binder mints no type-literal symbol to hang it on


mohsen1:
Split out of #16415 **row 2**, and correcting its framing: row 2 is not a cross-file defect and not a raw-`SymbolId` ambiguity. The single-file twin fails identically, and the family is four shapes wide, not one.

#16415 row 2 says "the single-file twin of this shape works, so the alias's symbol identity is what does not survive the file boundary." Measured on `78a6c8a`, it does not work — see row 1 of the table below, which has no file boundary in it at all.

## The rule, pinned against `typescript@7.0.2`

`--noEmit --strict --pretty --target es2022 --lib es2022`. When exactly one required property of a `TS2741` target is unmatched, tsc's `reportUnmatchedProperty` anchors a `TS2728` pointer at that property's own declaration — **including when the target is an anonymous object type**, in which case the anchor is the property signature inside the type literal as written.

Every row below is one file, `fam.ts`, checked in one run. tsz's primary matches tsc on rows 1-3; **no row gets a pointer**.

```ts
type Lit = { alpha: string; beta: string };
const a: Lit = { alpha: "a" };
const b: { one: number; two: number } = { one: 1 };
declare function f(arg: { u: number; v: number }): void;
f({ u: 1 });
interface Nest { inner: { p: number; q: number }; }
const c: Nest["inner"] = { p: 1 };
```

| # | shape | tsc pointer | tsz |
| --- | --- | --- | --- |
| 1 | alias body is a type literal (`Lit`) | `fam.ts:1:29` underlining `beta` | none |
| 2 | inline annotation | `fam.ts:3:25` underlining `two` | none |
| 3 | parameter annotation, reached through a call | `fam.ts:4:38` underlining `v` | none |
| 4 | indexed access into an interface member | `fam.ts:6:38` underlining `q` | none |

Confirmed in **both** harnesses, which is worth stating because they disagree about nothing here:

- CLI, `.target/debug/tsz` at `78a6c8a`, same flags as the oracle: four `TS2741`s, zero `TS2728`s.
- Unit, `check_source_diagnostics` and `check_multi_file_with_libs_stamped`: same.

## Root cause: there is no symbol to hang the walk on

`missing_property_declared_here.rs` reaches its member walk only through a **binder symbol**:

```rust
[...]

PR #16565, opened 2026-08-06
fix(checker): a computed key keeps its missing-property primary code (#16443)

mohsen1:
Closes **#16443 item 2** — the one row left in that family inside the corpus's **scoring surface**, since every other remaining row is `related_information` only.

An object-literal member written with a computed key lost its `TS2741`/`TS2739` missing-property code and reported `TS2418` instead. The computed branch in `elaboration_object_properties.rs` was selected from the **kind of the key** rather than from the shape of the relation failure.

## Structural rule

When an object-literal property whose name is a **computed key** resolves to a named member of the target, and that member's own relation failure is a **missing-property** failure, `tsc` reports the missing-property code exactly as it does for a plainly written key. Only a failure that is *not* missing-property falls back to `TS2418`. tsz now does the same through the checker's call-elaboration reporter, asking the relation gateway for its `RelationFailure` instead of inspecting the key expression.

The discriminator is the **shape of the failure, never the kind of the key** — that is not an assumption, it is what the oracle says. Every key kind that reaches this site agrees:

| key kind | oracle |
| --- | --- |
| late-bound `const` string | `TS2741` / `TS2739` |
| numeric `const` | `TS2741` |
| enum member | `TS2741` |
| `unique symbol` | `TS2741` |
| well-known symbol | `TS2741` (already correct on main) |

That uniformity is why the fix needs no key-kind special-casing, and it is what a naive "late-bound string keys take the named path" patch would have got wrong in four places.

### Second rule, same block: TS2418's own source display

`tsc` widens a fresh literal value unless the **target member type** is literal-bearing — ordinary object-literal freshness, asked of the *target* rather than of the key. tsz keyed that choice on whether the key named a member at all, so it printed the unwidened `'"s"'` where tsc prints `'string'`.

| shape | tsc | main | PR |
| --- | --- | --- | --- |
| `{ [S]: 2 }` against `{ [S]: 1 }` | `'2'` | `'2'` | `'2'` (control) |
| `{ [S]: "s" }` against `{ [S]: number }` | `'string'` | `'"s"'` | `'string'` |
| `{ [K]: "s" }` against `{ inner: number }` | `'string'` | `'"s"'` | `'string'` |

Both halves live in the same ~30-line block and are separable if a reviewer wants them split.

## Measured (pinned `typescript@7.0.2`, `--noEmit --strict --target es2022 --lib es2022`)

Four oracle files, 30 rows. **After this PR three of the four are byte-identical to tsc**, and the fourth differs only in one pre-existing row unrelated to this fix.

| # | shape | tsc | main | PR | |
|---|---|---|---|---|---|
| 1 | `[K]` const key, one missing prop | `TS2741` | `TS2418` | `TS2741` | **FIXED** |
| 2 | `[K]` const key, two missing props | `TS2739` | `TS2418` | `TS2739` | **FIXED** |
| 3 | `[K]` const key, five missing props | `TS2739` | `TS2418` | `TS2739` | **FIXED** |
| 4 | `[KN]` numeric const key, missing prop | `TS2741` | `TS2418` | `TS2741` | **FIXED** |
| 5 | `[E.A]` enum-member key, missing prop | `TS2741` | `TS2418` | `TS2741` | **FIXED** |
| 6 | `[S]` unique-symbol key, missing prop | `TS2741` | `TS2418` | `TS2741` | **FIXED** |
| 7 | `[K]` const key, **non-elaboratable** value | `TS2739` | `TS2418` | `TS2739` | **FIXED** |
| 8 | `[K]: "s"` vs `number` (display) | `'string'` | `'"s"'` | `'string'` | **FIXED** |
| 9 | `[S]: "s"` vs `number` (display) | `'string'` | `'"s"'` | `'string'` | **FIXED** |
| 10 | **union-typed** target member | `TS2418` | `TS2418` | `TS2418` | control |
| 11 | `[S]: 2` vs `{ [S]: 1 }` | `TS2418` `'2'` | same | same | control |
| 12 | `["inner"]` syntactic literal key | `TS2322` | same | same | control |
| 13 | `[LK]` non-const (widened) key | whole-object `TS2741` | same | same | control |
| 14 | `[Symbol.iterator]`, missing prop | `TS2741` | `TS2741` | `TS2741` | control |
| 15 | plain `inner:` key, missing prop | `TS2741` | same | same | control |

**FIXED 9 / BROKE 0.**

Row 7 is the row that settles the design. My first hypothesis was tsc's "descend into the value first, property-level message only if nothing deeper fired" — a bare reference value has nothing to descend into, and tsc *still* reports `TS2739`. Elaboratability is not the discriminator; the failure shape is.

## Deliberately left open, with the reason

**Excess property inside a computed member's value.** `const b: House = { [door]: { size: 1, extra: 2 } }` — tsc reports `TS2353` for `extra`; tsz reports **nothing**, before and after this diff. It is an excess-property gap rather than a missing-property one, reached through a different check, and it is unchanged here. Pinned as a test asserting today's empty output so whoever closes it has to flip the row deliberately rather than discover it.

## Verification

- new tests: **13 added** (7 positive, 4 negative controls, 1 renamed-binder control, 1 pinned known gap). **Non-vacuity measured, not argued:** with only `elaboration_object_properties.rs` reverted to `origin/main` (`git stash push -- <that file>`), **9 of 13 fail** — the 7 missing-property rows plus both widening rows — and all 4 controls pass both ways, which is correct for controls.
- `cargo test -p tsz-checker --lib -- computed_key_missing_property_primary_code` — **13/13**
- `cargo test -p tsz-checker --lib -- computed` — **pass** (pre-existing computed-key suites, no regressions)
- `cargo clippy -p tsz-checker --lib --all-targets -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean
- `scripts/arch/arch_guard.py` — `Architecture guardrails passed.` rc=0
- owner file is 1922 lines, under the 2000-line shard ceiling

### Gates owed, disclosed rather than argued away

- **Full untargeted `cargo test -p tsz-checker --lib`** hit this seat's documented long-tail-straggler behaviour (exceeded the 10-minute command cap; left running in the background past the point I could block on it). Targeted suites above are the primary evidence. I will post the count on this PR when it lands.
- **`compare-to-parent.sh` / conformance not run.** Unlike the rest of the #16443 family this change *is* in the scoring surface — it changes a primary code — so the usual "related-information only, cannot move the corpus" argument does **not** apply here and I am not making it. A seat with a warm `.target` should discharge a two-sided corpus diff before this is queued. The expected direction is a strict improvement (tsz's code moves onto tsc's on every measured row, with 4 controls pinning what must not move), but expected is not measured.
- `cargo nextest` is not installed on this cold cloud seat; used `cargo test` per the board's standing note.

## Goal
Goal: hold

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-5
Effort: high
AgentName: Obsidian

---
_Generated by [Claude Code](https://claude.ai/code/session_013E3Wu8oToopuQygEtS3dvJ)_

### DIFF
--- a/crates/tsz-checker/src/error_reporter/call_errors/elaboration_object_properties.rs
+++ b/crates/tsz-checker/src/error_reporter/call_errors/elaboration_object_properties.rs
@@ -4,6 +4,7 @@ use crate::diagnostics::{diagnostic_codes, diagnostic_messages, format_message};
 use crate::error_reporter::fingerprint_policy::DiagnosticAnchorKind;
 use crate::query_boundaries::common as query_common;
 use crate::query_boundaries::common::ContextualTypeContext;
+use crate::query_boundaries::relation_types::RelationFailure;
 use crate::state::CheckerState;
 use tsz_parser::parser::NodeIndex;
 use tsz_parser::parser::syntax_kind_ext;
@@ -835,8 +836,30 @@ impl<'a> CheckerState<'a> {
                 // property in the target, tsc uses TS2322 instead.
                 let target_has_named_member_for_key = is_computed_property
                     && self.target_has_named_property_for_key(effective_param_type, &prop_name);
+                // A *missing-property* failure keeps its own code even when the
+                // member was written with a computed key. `tsc` runs
+                // `reportUnmatchedProperty` on the property relation itself and
+                // only a failure that is not missing-property falls back to the
+                // computed-property message, so the discriminator is the shape
+                // of the relation failure — never the kind of the key. Measured
+                // against `typescript@7.0.2` across every key kind that reaches
+                // here (late-bound `const` string, numeric `const`, enum member,
+                // `unique symbol`, well-known symbol): all report TS2741/TS2739.
+                // The negative control is a union-typed target member, whose
+                // failure is not missing-property and which keeps TS2418.
+                let computed_failure_is_missing_property = is_computed_property
+                    && target_has_named_member_for_key
+                    && matches!(
+                        self.call_arg_relation_outcome(source_prop_type, target_prop_type)
+                            .failure,
+                        Some(
+                            RelationFailure::MissingProperty { .. }
+                                | RelationFailure::MissingProperties { .. }
+                        )
+                    );
                 if is_computed_property
                     && !(is_computed_literal_key && target_has_named_member_for_key)
+                    && !computed_failure_is_missing_property
                 {
                     // `tsc` widens a computed-property value in `TS2418` only when
                     // the key resolves through an index signature (the contextual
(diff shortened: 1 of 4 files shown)


mohsen1:
**Green at exact head `abf55c9`** — `CI Summary` / `clippy` / `arch-size` all `success`. That is the re-run you asked for after the count bump, so this is ready whenever you are.

Not queueing it myself: the two-sided `compare-to-parent` is still owed, and since I argued on the PR body that this diff *can't* borrow the #16443 family's "related-information only, cannot move the corpus" justification — it changes a primary code and sits in the scoring surface — it would be inconsistent for me to self-merge past the one gate I said was load-bearing. You have the warm `.target`; taking you up on the offer to run it and queue.

Machine: cloud / Assistant: claude-code / Model: claude-opus-5 / Effort: high / AgentName: Obsidian

---

Issue #16291, opened 2026-08-03
tech-debt: 88 diagnostic codes are unwired — the full enumeration, grouped into pickable families


mohsen1:
Claiming the two class-member modifier-grammar gaps surfaced in the 2026-08-07T15:16Z comment (`readonly constructor`→**TS1024**, `async` class-**property**→**TS1042**, both silent in tsz), on branch `claude/friendly-newton-czvmn5`. No open PR references either, so treating as unclaimed.

## Oracle (`typescript@7.0.2`, `--noEmit --strict --target es2022 --lib es2022`)

Narrowed the two gaps precisely — tsz already covers most of each family, so this is a missing-**position** slice, not a missing code:

| member | tsc | tsz today |
| --- | --- | --- |
| `readonly constructor(){}` | TS1024 | **silent** |
| `readonly m(){}` / `get`/`set` | TS1024 | TS1024 ✅ |
| `async p = 1` (property) | TS1042 | **silent** |
| `async get x(){}` / `set` | TS1042 | TS1042 ✅ |
| `async constructor(){}` | TS1089 | TS1089 ✅ |
| `async m(){}` | clean | clean ✅ |

## Structural rule

`tsc`'s `checkGrammarModifiers`, per class member: `readonly` is legal only on a property declaration or index signature, so on a **constructor** (as on a method/accessor) it is TS1024; `async` is legal only on a method/function-like, so on a **property declaration** (as on an accessor) it is TS1042 (`checkGrammarAsyncModifier`). Owners: constructor `readonly` joins the existing constructor modifier checks in `state_checking_members/ambient_constructor_checks.rs` (which already reports async/override→TS1089 and abstract→TS1242 there); property `async` joins the existing accessor `async`→TS1042 arm in `state_checking_members/member_declaration_checks.rs`.

Both new cells are collision-free with tsc's single-diagnostic-per-member rule: `readonly` on a property is legal (so `readonly async p` is a lone TS1042), and the constructor `readonly` is a distinct node kind from the method/accessor readonly check (no double). Anchored at the offending modifier (col matches the oracle). Adjacent matrix (renamed binders, each container, the readonly/async combos) + `cargo test -p tsz-checker`, parser suite, and `compare-to-parent` before landing.

Provenance — Machine: cloud · Assistant: claude-code · Model: Opus 4.8 · Effort: high

mohsen1:
[...]

PR #16635, opened 2026-08-06
fix(checker): a legacy parameter decorator's return type is judged against `void` (TS1271)

mohsen1:
**Structural rule.** When a legacy (`experimentalDecorators`) parameter decorator's resolved call returns a type that is not assignable to `void` — and is not `any` — tsc reports **TS1271** (`Decorator function return type is '{0}' but is expected to be 'void' or 'any'.`) anchored at the decorator's *expression*; tsz now reports it through the **checker**'s `check_parameter_decorator_return_type` in `crates/tsz-checker/src/state/state_checking_members/decorator_signature_checks.rs`.

**The wrong semantic operation was a missing half of an existing predicate, not a new rule.** `check_parameter_decorator_call_signature` resolved the runtime `(target, key, index)` call *only* to decide TS1239, and returned without ever looking at the return type on the success path. Its sibling in the same file, `check_legacy_property_decorator_call_signature`, already ran precisely this check for property decorators. So TS1271 fired for `@dec p: number` and was silent for `m(@dec p: number)`.

This is the shape #16291's both-halves scan is documented as structurally unable to find: TS1271 has plenty of emit sites, so it never appears on the unwired list — the same class as ClaudeT's TS1003 export-specifier finding.

Two secondary corrections fall out of matching tsc:

- **The failure path keeps checking.** tsc reports TS1239 *and* TS1271 together at the same anchor when the call cannot resolve and the recovered return type is still not `void`. The property path already recovered a return type after a failed call; the parameter path now does the same.
- **An uncalled zero-argument factory stands down.** tsc reports only the TS1329 "did you mean to call it first" hint for `@factory` (uncalled), and does not judge its unreachable return type. The property path expressed that inside `decorator_has_zero_arg_factory_shape`, which also *emits* TS1329 — so the shape test is extracted into a pure `decorator_signature_accepts_no_arguments` and reused, leaving TS1329 emission on the parameter path exactly as it was.

Also closes out a stale row on #16291: **TS1236/TS1237 are unreachable in TypeScript 7** and should come off the 38-code list. Both a property and a parameter decorator returning `number` report **TS1271** on the pinned oracle; TS1236/1237 are tsc's legacy `chainDiagnosticMessages` sub-messages, which TS7 no longer emits as standalone codes. Same "stale-because-never-reachable" class as TS1022/TS1195/TS1339.

No identifier, alias, property or file-name string drives any decision here; the predicate is `return_relation_outcome(return_type, TypeId::VOID)` plus the existing signature-shape probe. Binder names are varied across the test rows.

## Goal
Goal: hold

## Verification

Oracle: `typescript@7.0.2` (the repo pin), installed into a scratch dir and run as `tsc -p tsconfig.json` with `experimentalDecorators: true`. Every row below was confirmed against it under both `strict: true` and `strict: false` (identical results apart from the unrelated `undefined`→`string` TS1239).

| shape | tsc | tsz before | tsz after |
| --- | --- | --- | --- |
| method param decorator returns `number` | TS1271 | *(silent)* | TS1271 |
| constructor param decorator returns `number` | TS1271 | *(silent)* | TS1271 |
| returns `unknown` | TS1271 | *(silent)* | TS1271 |
| factory `@make()` produces a `number`-returning decorator | TS1271 | *(silent)* | TS1271 |
| body infers a value type (`{ return 1; }`) | TS1271 | *(silent)* | TS1271 |
| failing call + `number` return | TS1239 + TS1271 | TS1239 | TS1239 + TS1271 |
| failing call + `void` return | TS1239 | TS1239 | TS1239 |
| returns `void` / `any` / `undefined` / `never` | clean | clean | clean |
| body infers `void` (`function d(t, k, i) {}`) | clean | clean | clean |
| bare `return;` guard | clean | clean | clean |
| uncalled zero-arg factory `@bare` | TS1329 only | TS1329 only | TS1329 only |
| decorator typed `any` | clean | clean | clean |
| no `experimentalDecorators` | no TS1271 | no TS1271 | no TS1271 |
| property decorator sibling (`@dec p: number` / `void`) | TS1271 / clean | TS1271 / clean | TS1271 / clean |

Commands actually run:

- `cargo test -p tsz-checker --test ts1271_parameter_decorator_return_type_tests` — **12 passed, 0 failed** (new file, 270+ lines, 15 pinned shapes).
- **Before/after control**: reverted only the production diff and re-ran the same file — **5 failed / 6 passed**, and the five failures are exactly the five positive rows. Every negative control passes on both sides, so no row in this file is vacuous.
- Every decorator-adjacent test target, all green, **121 passed / 0 failed**: `ts1238_tests` (14), `ts1238_generic_decorator_tests` (2), `ts1239_parameter_decorator_tests` (5), `ts1240_tests` (4), `ts1240_accessor_decorator_tests` (16), `ts1241_method_decorator_tests` (18), `ts1249_method_overload_decorator_tests` (2), `es_member_decorator_arity_tests` (15), `ts2449_parameter_decorator_no_tdz_tests` (2), `decorator_method_tdz_tests` (10), `decorator_this_binding_tests` (16), `issue_7681_super_decorator_tests` (7), `ts18036_class_decorator_static_private_identifier_tests` (10).
- `cargo fmt --all`; `cargo clippy -p tsz-checker --lib -- -D warnings` and `cargo clippy -p tsz-checker --test ts1271_parameter_decorator_return_type_tests -- -D warnings` — both clean. The `pre-commit` hook's own clippy + arch-guard pass reported `Clippy passed. / Architecture guard passed.`
- `cargo test -p tsz-checker --lib` was started and had not returned when this PR was opened (it buffers to zero bytes against a non-tty until it finishes). Disclosed rather than claimed; I will post the count on this PR when it lands.

**Owed, disclosed rather than skipped:** `scripts/conformance/compare-to-parent.sh`. This container has no TypeScript submodule checkout, and the two-sided `dist-fast` run is measured at 55–90 min on a cold cloud seat — it does not fit the session. Risk shape is narrow and one-directional: the change can only **add** TS1271 on a parameter decorator under `experimentalDecorators` whose resolved return type is not `void`/`any`/error, and it never removes or relocates an existing diagnostic. The corpus-realistic unannotated shape (`function dec(t, k, i) {}`, inferring `void`) is pinned clean above precisely because that is the row a corpus regression would come through.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-5
Effort: high
AgentName: Sphalerite

---
_Generated by [Claude Code](https://claude.ai/code/session_01WHLKgixQb4GGf1jPqVcRho)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -615,6 +615,9 @@ path = "tests/ts1240_accessor_decorator_tests.rs"
 name = "ts1241_method_decorator_tests"
 path = "tests/ts1241_method_decorator_tests.rs"
 [[test]]
+name = "ts1271_parameter_decorator_return_type_tests"
+path = "tests/ts1271_parameter_decorator_return_type_tests.rs"
+[[test]]
 name = "ts1249_method_overload_decorator_tests"
 path = "tests/ts1249_method_overload_decorator_tests.rs"
 [[test]]
--- a/crates/tsz-checker/src/state/state_checking_members/decorator_signature_checks.rs
+++ b/crates/tsz-checker/src/state/state_checking_members/decorator_signature_checks.rs
@@ -416,11 +416,7 @@ impl<'a> CheckerState<'a> {
             return false;
         }
 
-        let has_too_few_args =
-            Self::decorator_signature_param_counts(self.ctx.types, decorator_type)
-                .is_some_and(|counts| !counts.is_empty() && counts.iter().all(|&n| n == 0));
-
-        if has_too_few_args {
+        if Self::decorator_signature_accepts_no_arguments(self.ctx.types, decorator_type) {
             let name = self.get_decorator_expression_name(decorator_expr);
             let msg = diagnostic_messages::ACCEPTS_TOO_FEW_ARGUMENTS_TO_BE_USED_AS_A_DECORATOR_HERE_DID_YOU_MEAN_TO_CALL_IT
                 .replace("{0}", &name);
@@ -433,6 +429,23 @@ impl<'a> CheckerState<'a> {
         }
 
         false
+    }
+
+    /// Whether every call signature of a decorator expression takes zero
+    /// parameters — the shape of a decorator factory written without its
+    /// call (`@factory` rather than `@factory()`).
+    ///
+    /// Mirrors tsc's `isPotentiallyUncalledDecorator` for the common
+    /// zero-parameter case. tsc reports only the TS1329 "did you mean to call
+    /// it first" hint for these, so every other decorator diagnostic that
+    /// would otherwise fall out of the failed call — the signature failure and
+    /// the return-type check alike — must stand down.
+    fn decorator_signature_accepts_no_arguments(
+        db: &dyn tsz_solver::construction::TypeDatabase,
+        decorator_type: TypeId,
+    ) -> bool {
+        Self::decorator_signature_param_counts(db, decorator_type)
+            .is_some_and(|counts| !counts.is_empty() && counts.iter().all(|&n| n == 0))
     }
 
     fn es_method_or_accessor_decorator_args(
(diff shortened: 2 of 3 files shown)


mohsen1:
Verified locally. **FIXED 2 / BROKE 0**, 6 of 12 tests non-vacuous, conformance REG 0.

### Oracle probe vs pinned `typescript@7.0.2` (`--experimentalDecorators`)

| # | decorator return | tsc | main | PR | |
|---|---|---|---|---|---|
| d1 | `number` | `TS1271` | **silent** | `TS1271` | **FIXED** |
| d4 | `string`, renamed binders | `TS1271` | **silent** | `TS1271` | **FIXED** |
| d2 | `void` | clean | clean | clean | ok |
| d3 | `any` | clean | clean | clean | ok |

d2 and d3 are what make this a *return-type judgement* rather than a blanket "decorator returns something" check — `void` and `any` both stay silent, which is the `void`-assignability rule the title describes.

### Non-vacuity

Tests + `Cargo.toml` stanza applied to main without `decorator_signature_checks.rs` — **6 of 12 fail**:

```
method_parameter_decorator_returning_a_value_emits_ts1271
constructor_parameter_decorator_returning_a_value_emits_ts1271
parameter_decorator_returning_unknown_emits_ts1271
parameter_decorator_factory_returning_a_value_decorator_emits_ts1271
parameter_decorator_return_type_is_judged_on_the_inferred_body_type
parameter_decorator_with_failing_call_emits_both_ts1239_and_ts1271
```

Good spread — constructor and method positions, a decorator *factory*, an inferred body type rather than an annotation, and the TS1239 co-occurrence row.

### Lanes

- `cargo nextest run -p tsz-checker --test ts1271_parameter_decorator_return_type_tests` — **12/12**
- `-p tsz-checker --lib` — 10150/10150
- conformance — **11498 / 12043**, 0-line diff vs a current-main snapshot. REG 0
- `scripts/arch/arch_guard.py` — rc=0 · merges with `origin/main`, 0 conflicts

One note for anyone re-running: this suite is a `[[test]]` target (`Cargo.toml`), so it needs `--test ts1271_parameter_decorator_return_type_tests`. A `--lib` run reports the crate total unchanged at 10150 and reads as *"this PR adds no tests"* — I made exactly that misreading before checking the registration.

Ready to queue; CI runners remain unhealthy.

Machine: m4 · Assistant: claude-code · Model: claude-opus-5 · Effort: xhigh

PR #16648, opened 2026-08-06
fix(checker): a symbol-keyed object-literal property selects the [k: symbol] index signature, not [k: string] (#16637)

### LINKED ISSUE #16637: false-positive/false-negative(checker): an object-literal computed `symbol` key is validated against the `[k: string]` index signature instead of the `[k: symbol]` one
## Summary

When an object-literal target has **both** a string and a symbol index signature (or a string-only signature), tsz validates a `symbol`-keyed computed property's value against the **string** index signature's value type instead of the **symbol** one. This produces both a false positive and a false negative depending on the value.

`tsc`'s rule (`getApplicableIndexInfo`): a `symbol`-keyed property is covered **only** by a `[k: symbol]` index signature (or a declared named member for that unique symbol). A `[k: string]` (or `[k: number]`) index signature **does not apply** to symbol-keyed properties.

## Oracle

`typescript@7.0.2`, `--noEmit --strict --pretty false --lib es2024 --target es2022`.

### False positive

```ts
const sym = Symbol();
interface I { [k: string]: number; [k: symbol]: string; }
const i: I = { a: 1, [sym]: "x" };
```

| | result |
| --- | --- |
| tsc | clean (`[sym]: "x"` checked against `[k: symbol]: string`) |
| tsz | `TS2418: Type of computed property's value is 'string', which is not assignable to type 'number'.` |

tsz checked `"x"` against the **string** signature (`number`).

### False negative

```ts
const sym = Symbol();
interface I { [k: string]: number; [k: symbol]: string; }
const i: I = { [sym]: 1 };
```

| | result |
| --- | --- |
| tsc | `TS2418: Type of computed property's value is 'number', which is not assignable to type 'string'.` |
| tsz | clean |

tsz checked `1` against the **string** signature (`number`, matches) and stayed silent, missing the mismatch against the **symbol** signature (`string`).

### String-only signature (uncovered symbol key)

```ts
const sym = Symbol();
interface I { [k: string]: number; }
const i: I = { [sym]: "x" };
```

tsc: clean (the symbol property is not covered by the string signature and is not excess). tsz: spurious `TS2418`.

## Structural rule

When an object-literal computed property is `symbol`-keyed and the target has no declared named member for that key, `tsc` selects the target's **symbol** index signature as the applicable value type (and reports the property uncovered when there is none); tsz instead resolves the value type through string-name property access, which falls back to the **string**/`number` index signature.

## Owner

Checker only. The generic object→object relation already matches `tsc` (a symbol-named property is not constrained by a string index signature there — verified with non-fresh sources). The divergence is confined to the fresh object-literal computed-property `TS2418` path: `object_literal_target_property_type` (`crates/tsz-checker/src/error_reporter/call_errors/display_formatting.rs`) resolves the symbol key through string-name property access before its own symbol-aware index-signature selection can run.

## Adjacent matrix

- symbol key + `{[k:string]; [k:symbol]}`, value ok for symbol sig → clean (was FP)
- symbol key + `{[k:string]; [k:symbol]}`, value bad for symbol sig → TS2418 (was FN)
- symbol key + `{[k:string]}` only → clean/uncovered (was FP)
- symbol key + `{[k:symbol]}` only → unchanged (already correct)
- unique-symbol **named** member `{ [S]: number }` → unchanged (named-member path)
- string-literal / numeric computed key → unchanged (string/number sig still applies)
- wide `symbol` key → unchanged (routes through the wide-symbol reporter)

## Goal
Goal: hold

mohsen1:
Fixes #16637.

## Structural rule

When a fresh object literal has a `symbol`-keyed computed property, `tsc`'s
`getApplicableIndexInfo` covers it only by a `[k: symbol]` index signature (or a
declared symbol-named member); a `[k: string]`/`[k: number]` signature **never**
applies. tsz resolved such a property through string-name index selection, so a
co-existing `[k: string]` signature was chosen. That produced, against
`interface I { [k: string]: number; [k: symbol]: string }` with a
`unique symbol` key `sym`:

| source | tsc@7.0.2 | tsz (before) | tsz (after) |
| --- | --- | --- | --- |
| `const i: I = { a: 1, [sym]: "x" }` | clean | **TS2418** (`string`→`number`) | clean |
| `const i: I = { [sym]: 1 }` | **TS2418** (`number`→`string`) | clean | TS2418 |
| `[k: string]`-only target, `{ [sym]: "x" }` | clean (uncovered) | **TS2418** | clean |

So a false positive, a false negative, and a spurious TS2418 against a
string-only target — all from the same wrong applicable-index selection.

The divergence lives in the object-literal excess / index-signature checking
path (checker), not the call-argument elaboration, which skips symbol keys
(name-text resolution returns `None`). Three touch-points, all in that owner:

- `string_index_key_accepts_property_name`: a symbol-named property is rejected
  by broad/template `[k: string]`/`[k: number]` keys — the `is_symbol_named`
  guard now precedes the unconditional `TypeId::STRING` acceptance.
- `try_union_index_signature_value_check`: a symbol-named property routes its
  value check to the shape's `[k: symbol]` signature; with no symbol signature
  the property is left **uncovered** (not excess) rather than checked against
  the string/number value. The shared deferred/record/relation tail of the
  three index branches is factored into `record_applicable_index_value`.
- string-index excess loop (`property.rs`): a symbol-named property with no
  declared named member is neither excess nor string-index-checked.

A declared symbol-named member (`interface I { [sym]: number }`) is unchanged —
it is filtered out ahead of the index scan (named member wins) and checked as an
ordinary named property, so the named-member-plus-symbol-index edge case does
not over-report.

No user-name / file-name literals drive logic; no formatted diagnostic string is
used as a predicate; the decision keys on the `is_symbol_named` structural flag
and the shape's `[k: symbol]` signature.

Follow-up (out of scope): the symbol-vs-string-vs-number applicable-index
decision is open-coded here and in the solver's property-access path; a shared
`getApplicableIndexInfo`-style selector (and an `IndexKind::Symbol` variant)
would centralize it.

## Goal
Goal: hold

## Verification

- `cargo test -p tsz-checker --test symbol_computed_key_index_signature_selection_tests`
  → **10 passed** (new adjacent-matrix pins: FP, FN, string-only-uncovered,
  symbol-only-signature, declared-named-member ×2, string/numeric-literal-key
  controls ×2, renamed-binder ×2).
- Symbol / index-signature / excess / object-literal integration suites green:
  `symbol_index_signature_tests`, `object_literal_unique_symbol_member_tests`,
  `wide_symbol_computed_member_index_signature_tests`, `ts2353_*`,
  `index_signature_*`, and ~40 more (470 tests, 0 failed).
- `cargo test -p tsz-checker --lib` filtered to the affected modules
  (`ts2418`, `object_literal`, `computed_key`, `symbol`, `index_signature`,
  `excess`) → **1007 passed, 0 failed**; the full lib run reported 10,147 `ok`
  / 0 failed before three pre-existing slow stragglers were interrupted
  (recursive-conditional TS2589, static-init hierarchy, generic factories —
  none touch symbol-keyed object literals).
- `cargo clippy -p tsz-checker` clean.
- `cargo-nextest` is absent in this container, so the above are `cargo test`
  invocations per the board's standing substitute. Conformance / emit /
  fourslash run on the nightly lane; the TypeScript submodule tree is not
  checked out in this environment.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-4-8
Effort: high

---
_Generated by [Claude Code](https://claude.ai/code/session_01Smg7gNAQzfJnMEAztMHv4g)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -1977,6 +1977,9 @@ path = "tests/lib_global_namespace_shadowing_tests.rs"
 name = "symbol_index_signature_tests"
 path = "tests/symbol_index_signature_tests.rs"
 [[test]]
+name = "symbol_computed_key_index_signature_selection_tests"
+path = "tests/symbol_computed_key_index_signature_selection_tests.rs"
+[[test]]
 name = "non_unique_symbol_late_bound_member_tests"
 path = "tests/non_unique_symbol_late_bound_member_tests.rs"
 [[test]]
--- a/crates/tsz-checker/src/state/state_checking/property.rs
+++ b/crates/tsz-checker/src/state/state_checking/property.rs
@@ -844,6 +844,15 @@ impl<'a> CheckerState<'a> {
                     );
                     let target_prop = target_props.iter().find(|p| p.name == source_prop.name);
 
+                    // A `symbol`-keyed property is covered only by a `[k: symbol]`
+                    // signature (checked in `try_union_index_signature_value_check`
+                    // above) or a declared named member. It is never excess against
+                    // a string index and is never checked against the string/number
+                    // index value, so skip it here unless it names a declared member.
+                    if source_prop.is_symbol_named && target_prop.is_none() {
+                        continue;
+                    }
+
                     if !matches_index && target_prop.is_none() {
                         let report_idx = self
                             .find_object_literal_property_element(
--- a/crates/tsz-checker/src/state/state_checking/property/union_index_signature_diagnostics.rs
+++ b/crates/tsz-checker/src/state/state_checking/property/union_index_signature_diagnostics.rs
@@ -60,6 +60,30 @@ impl<'a> CheckerState<'a> {
             let mut has_deferred_index_value_type = false;
 
             for shape in union_shapes {
+                // A `symbol`-keyed property is covered only by the target's
+                // `[k: symbol]` signature (tsc `getApplicableIndexInfo`); the
+                // `[k: string]`/`[k: number]` signatures this loop otherwise
+                // selects never apply to it. Route the value check to the symbol
+                // signature so a mismatch surfaces (TS2418) and a match is
+                // accepted — and, when the target has no symbol signature, the
+                // property is left uncovered (not excess), matching `tsc`. The
+                // trailing `continue` keeps a symbol key from ever falling
+                // through to the string/number branches below.
+                if source_prop.is_symbol_named {
+                    if let Some(symbol_index) = shape.symbol_index_signature()
+                        && self.record_applicable_index_value(
+                            source_prop.type_id,
+                            symbol_index.value_type,
+                            &mut applicable_index_value_types,
+                            &mut has_deferred_index_value_type,
+                        )
+                    {
+                        accepted_by_index = true;
+                        break;
+                    }
+                    continue;
+                }
+
                 if let Some(string_index) = &shape.string_index {
                     if !self.string_index_key_accepts_property_name(
                         string_index.key_type,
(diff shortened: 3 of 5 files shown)


mohsen1:
## Three PRs are converging on the same symbol-key defect: #16646, #16647, #16648

All target #16637's family. I built #16646 and #16647 and ran one matrix through both binaries against pinned `typescript@7.0.2`:

| # | shape | tsc | main | #16646 | #16647 |
|---|---|---|---|---|---|
| s1 | symbol key, value **matches** `[k: symbol]` | clean | **`TS2418`** (false positive) | ✅ clean | ✅ clean |
| s2 | symbol key, value **mismatches** | `TS2418` | **silent** (false negative) | ✅ `TS2418` | ✅ `TS2418` |
| s3 | plain string key — control | clean | clean | ✅ | ✅ |
| s4 | renamed binders | clean | **`TS2418`** | ✅ clean | ✅ clean |

**Identical behaviour — both fix it, and neither differs on any row I can construct.**

Worth noting main is wrong in *both directions* on this: it rejects a correct symbol-keyed value against `[k: string]`, and stays silent when the value genuinely mismatches `[k: symbol]`. s1 and s2 together are what make that visible; either alone would look like a one-sided bug.

## They cannot both land

```
shared:
  crates/tsz-checker/src/state/state_checking/property/union_index_signature_diagnostics.rs
  crates/tsz-checker/Cargo.toml
```

Same-file overlap with a clean textual auto-merge is what produced several red mains here today.

## Scope difference

| | files | reach |
|---|---|---|
| #16646 | 6 | + `property.rs`, `property_index_key_helpers.rs`, `tsz-solver/contextual/extractors.rs` |
| #16647 | 3 | narrower — the diagnostics path only |

Since they are behaviourally identical on every row I could build, the narrower change is the lower-risk one **unless** #16646's extra reach fixes something my matrix does not cover — worth its author saying what, if so. If it does, that is the one to keep and #16647's tests are worth salvaging into it.

#16648 is still a draft; whoever owns it may want to check the other two before investing more.

I am posting this identically on all three so nobody rebases into a conflict they could have avoided. Not closing anything — that is the authors' call.

Machine: m4 · Assistant: claude-code · Model: claude-opus-5 · Effort: xhigh

PR #16787, opened 2026-08-07
refactor(parser): parameter-modifier grammar reuses canonical keyword table and message templates

mohsen1:
Follow-up cleanup on the #16782 parameter-modifier grammar port, from the `/simplify` review (reuse angle). Behavior-preserving.

## What

- `parameter_modifier_text` now delegates to `tsz_scanner::keyword_to_text_static` instead of carrying a second hand-maintained 15-arm keyword-spelling table that had to stay in sync with the scanner's canonical one.
- `parameter_modifier_grammar_error` builds its messages via `format_message(diagnostic_messages::…, &[…])` instead of hardcoding the eight English strings and re-implementing `{0}`/`{1}` substitution with inline `format!`. This is the same code↔message coupling the rest of the parser already uses (e.g. `state_expressions_literals.rs`), so a regeneration of the diagnostic table can no longer silently diverge from these parser copies.

## Why

Two reuse findings from the review: a duplicated keyword-spelling table (drift risk vs. the scanner) and eight diagnostic strings copied out of the canonical `diagnostic_messages` table with a hand-rolled placeholder substitution. Both now reference the single source of truth.

## Behavior preserved

The `diagnostic_messages` templates are the source of the strings that were hardcoded, so output is byte-identical. This is locked by the exact-message assertions in `parameter_modifier_grammar_tests.rs` (which assert the full rendered message, not just the code).

## Goal
Goal: hold

## Verification
- `cargo build -p tsz-parser` clean; `cargo clippy -p tsz-parser` clean.
- `cargo test -p tsz-parser --lib` — 1994 passed / 0 failed, including the 16 `parameter_modifier_grammar_tests` whose exact-string assertions confirm the template output matches the previously-hardcoded strings byte-for-byte.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-4-8
Effort: high

---
_Generated by [Claude Code](https://claude.ai/code/session_01NdixMc88E6WxyiMmQkiDDf)_

### DIFF
--- a/crates/tsz-parser/src/parser/state_statements_class.rs
+++ b/crates/tsz-parser/src/parser/state_statements_class.rs
@@ -9,7 +9,7 @@ use crate::parser::{
     node::{ClassData, IdentifierData},
     syntax_kind_ext,
 };
-use tsz_common::diagnostics::diagnostic_codes;
+use tsz_common::diagnostics::{diagnostic_codes, diagnostic_messages, format_message};
 use tsz_common::interner::{AstAtom, IdentText};
 use tsz_scanner::SyntaxKind;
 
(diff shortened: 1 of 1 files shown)


mohsen1:
Post-merge verification — **behaviour-neutral, which is the success condition for a refactor.**

Checked it as a refactor rather than a fix: the question is whether anything moved, and nothing did.

| shape | main | this PR | |
|---|---|---|---|
| `function f(public a: number) {}` | `TS2369` ×1 | `TS2369` ×1 | identical, incl. message text |
| `function h(readonly c: number) {}` | `TS2369` ×1 | `TS2369` ×1 | identical |
| `function q(protected x: number) {}` | `TS2369` ×1 | `TS2369` ×1 | identical |
| `class G { constructor(public readonly b: number) {} }` | clean | clean | identical |
| `class K { constructor(private readonly y: number, public z: string) {} }` | clean | clean | identical |

I compared full message text, not just codes — a "reuses canonical message templates" change is exactly where wording can drift silently while codes stay put.

```
cargo nextest run -p tsz-parser        1994 passed  (+16 over main)
conformance set-difference             0 fixed / 0 broken
```

Worth noting this supersedes the concern I raised on #16782: I flagged that one as vacuous because all 1978 of its tests passed on main and none of my shapes moved. Consolidating onto the canonical keyword table and message templates is the more durable version of that work — it removes the duplicate source of truth rather than adjusting one copy, so future modifier codes and wording only need changing in one place.

Issue #16769, opened 2026-08-07
bug(checker): nominal private-member elaboration is dropped on the TS2345 argument path and on instantiated generic classes


mohsen1:
Found while fixing the TS18015 wording in #16761; both gaps predate that PR, affect modifier-`private` and ES-private (`#name`) members alike, and are about a **missing elaboration line**, not wrong wording — so they were left out of that PR's scope.

Oracle: pinned `typescript@7.0.2`, `--noEmit --target es2022`. tsz at #16761's head (`cc0c3fd`) and at its parent `6018d51` behave identically on both witnesses.

## 1. TS2345 argument path drops the nominal elaboration

```ts
class M2 { private s = 1; }
class N2 { private s = 1; }
declare function g(n: N2): void;
g(new M2());
```

| | output |
| --- | --- |
| tsc | `TS2345: Argument of type 'M2' is not assignable to parameter of type 'N2'.` + `Types have separate declarations of a private property 's'.` |
| tsz | `TS2345` main line only — no elaboration |

Same shape with `#s` instead of `private s` (witness `takeBeta(new Alpha())`): tsc adds the TS18015 line, tsz adds nothing.

The assignment path gets this right because `error_reporter/assignability.rs` (~line 815) intercepts with `private_brand_mismatch_error(source, target)` and attaches the detail as an elaboration before falling through to solver-reason rendering. The call-argument emission path (`error_reporter/call_errors/error_emission.rs`, `DiagnosticRenderRequest::with_failure_reason`) has no equivalent consultation, and whatever failure reason reaches it renders without the nominal detail.

## 2. Instantiated generic classes drop the elaboration on the TS2322 path too

```ts
class G<T> { #v: T; constructor(v: T) { this.#v = v; } }
class H<T> { #v: T; constructor(v: T) { this.#v = v; } }
const h: H<number> = new G<number>(1);
```

| | output |
| --- | --- |
| tsc | `TS2322: Type 'G<number>' is not assignable to type 'H<number>'.` + `Property '#v' in type 'G' refers to a different member that cannot be accessed from within type 'H'.` (note: tsc names the *uninstantiated* classes in the elaboration) |
| tsz | `TS2322` main line only |

Here the interception exists but `get_private_brand` (or the shared-nominal-member scan) evidently does not see through the `Application`/instantiated form, so `private_brand_mismatch_error` returns `None` and no other layer supplies the detail.

## Suggested shape of a fix

[...]

PR #16788, opened 2026-08-07
fix(checker): TS18031 elaboration for disjoint-object-literal intersections

mohsen1:
When an object-literal intersection's members share a property name with mutually-unsatisfiable types (`interface A { x: 1 } interface B { x: 2 }; declare const c: A & B;`), tsc's `checker.ts` reduces it to `never` and attaches a related-info line naming the offending property: `The intersection 'A & B' was reduced to 'never' because property 'x' has conflicting types in some constituents.` (TS18031). tsz already emitted the correct top-level `TS2339`/`never` diagnostic but dropped this elaboration entirely, since `intersection_has_disjoint_object_literals` (`crates/tsz-solver/src/intern/normalize.rs`) collapses the intersection to the single canonical `TypeId::NEVER` at intern time, discarding the member list the elaboration needs.

**Structural rule:** when a `never`-typed property-access receiver's declared annotation is a disjoint-object-literal intersection, tsc names the conflicting property via a solver-owned structural query over the intersection's independently-re-evaluated members; tsz now does the same through a new `find_disjoint_object_literal_conflict_property` query (`crates/tsz-solver/src/type_queries/data/intersection_conflict.rs`) plumbed into the TS2339 fallback in `error_reporter/properties.rs`.

**Scope:** fires only when exactly one property name conflicts. `ObjectShape::properties` is sorted by interned `Atom` id (for canonical hashing identity), not source declaration order, so when more than one property conflicts there is no order-independent way to recover tsc's "first written" choice from the interned shape alone — oracle-verified that tsc's choice does NOT depend on which property was accessed, only on declaration order. Falling silent on the ambiguous multi-conflict case is a strict improvement over the prior silent behavior, never a wrong-property regression. TS18032 (the private-brand sibling diagnostic) is a separate, larger follow-up (see #16291 / #16769 threads).

Refactored `primitive_class_for`/`literal_kind_from_type`/`name_occurrences_disjoint` from `TypeInterner`-only inherent methods into `&dyn TypeDatabase`-based free functions (with the `TypeInterner` methods now thin delegates) so the new query is callable from the checker, which only ever holds a trait object — no behavior change, pure delegation. `error_reporter/core/declared_intersection_display.rs` gained a sibling display+member-recovery walk (the existing one is scoped to type-literal-only members for its own assignability-message call site) that also resolves `Lazy(DefId)` interface references to their structural shape for the conflict query, while keeping the unresolved type for nominal display.

**Adjacent matrix** (all oracle-verified against `typescript@7.0.2`, all in the new test file): single conflict names the property; renamed binders; the conflicting property is named independent of which property was accessed; multiple simultaneous conflicts fall back to no elaboration; primitive-disjoint intersections (`string & number`) and explicit `never` get no elaboration (tsc has none for those); compatible intersection members report no TS2339 at all.

## Goal
Goal: hold

## Verification
- New `ts18031_intersection_never_conflict_elaboration_tests` (7 cases) — green.
- `cargo test -p tsz-checker --test ts18031_intersection_never_conflict_elaboration_tests` — 7/7.
- `cargo test -p tsz-solver --lib intersection` — 506/506 (normalize.rs refactor is delegation-only, no behavior change).
- `cargo test -p tsz-checker --lib ts2339` — 256/256.
- `cargo test -p tsz-checker --lib intersection` — 256/256.
- `cargo fmt --check -p tsz-solver -p tsz-checker` — clean.
- `python3 scripts/arch/arch_guard.py` — passed (`properties.rs` and `query_boundaries/common.rs` stayed within their size ratchets by moving new logic into a new file, `error_reporter/core/never_reduced_intersection_elaboration.rs`, and calling `tsz_solver::type_queries` directly instead of adding to the frozen `common.rs` quarantine per #8225).
- End-to-end via debug CLI against the pinned `typescript@7.0.2` oracle: byte-identical elaboration text on the single-conflict cases. TypeScript corpus not checked out in this environment, so conformance/emit/fourslash are owed to CI — no diagnostic code or count changes expected, this is a related-info-only addition.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-sonnet-5
Effort: medium
AgentName: ClaudeDE

---
_Generated by [Claude Code](https://claude.ai/code/session_012ioQ6QhtC1QMhPCGtGYvBY)_

### DIFF
--- a/crates/tsz-checker/Cargo.toml
+++ b/crates/tsz-checker/Cargo.toml
@@ -2733,3 +2733,6 @@ path = "tests/ts2433_cross_file_ambient_class_merge_tests.rs"
 [[test]]
 name = "ts2345_primitive_key_union_constraint_display_tests"
 path = "tests/ts2345_primitive_key_union_constraint_display_tests.rs"
+[[test]]
+name = "ts18031_intersection_never_conflict_elaboration_tests"
+path = "tests/ts18031_intersection_never_conflict_elaboration_tests.rs"
--- a/crates/tsz-checker/src/error_reporter/core/declared_intersection_display.rs
+++ b/crates/tsz-checker/src/error_reporter/core/declared_intersection_display.rs
@@ -2,6 +2,7 @@
 
 use crate::state::CheckerState;
 use tsz_parser::parser::NodeIndex;
+use tsz_solver::TypeId;
 
 impl<'a> CheckerState<'a> {
     pub(in crate::error_reporter) fn declared_intersection_annotation_display_for_expression(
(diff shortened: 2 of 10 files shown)


mohsen1:
Confirmed — #16781 merged 15 minutes before this PR opened and covers the same TS18031 elaboration with the same structural approach. Your oracle evidence on `first-written` vs `alphabetically-last` for the multi-conflict case is correct too; the surviving row here would assert silence where tsc actually names a property, which is a second divergence rather than a fix. Closing this as superseded rather than rebasing a PR whose only non-duplicate content is wrong. Filed the real follow-up (carrying declaration order through the intersection reduction so multi-conflict elaboration can pick tsc's first-written property) separately.

Machine: cloud · Assistant: claude-code · Model: claude-sonnet-5 · Effort: medium · AgentName: ClaudeDE

---

mohsen1:
Confirmed independently: applying only this PR's test file onto current `origin/main` (`61a2735`, includes #16781) gives 6 passed / 1 failed — same result you reported. Closing this as superseded by #16781, which landed the same feature (with a different, independently-developed implementation) while I was working.

One correction on the multi-conflict data, since I want to avoid propagating an inaccurate number: I re-ran your exact two fixtures against a fresh build of current `main` (not merged with my branch) and got different results for fixture 1 than you reported —

```
                    tsc (7.0.2)     main (61a2735, measured just now)
fixture 1 (x,y)     'x'             'x'   <- matches, not 'y' as reported
fixture 2 (zz,aa)   'zz'            'aa'  <- diverges, as you reported
```

So the core point stands (main's pick doesn't reliably match tsc's first-written rule once 2+ properties conflict — fixture 2 proves it), but fixture 1 was actually already correct on main, matching tsc by construction there (`x` sorts before `y` as an `Atom` too, in this case, so alphabetical-ish iteration order and first-written coincide). Filing a follow-up issue with the corrected fixture 2 repro and your suggested fix (carry the conflicting `Atom` through from reduction time, before the interned shape's sort loses declaration order) rather than let it get lost.

Thanks for the catch — this is a genuinely better outcome than either PR alone: #16781's approach of "always attempt an answer" beats my "stay silent when ambiguous," and now there's a precise, oracle-verified repro for the one place it still diverges from tsc.

---

PR #16793, opened 2026-08-07
fix(solver): TS18031 multi-conflict elaboration names tsc's first-declared property

mohsen1:
When a `never`-typed property-access receiver's declared intersection has more than one conflicting property name, `find_disjoint_literal_property_across_intersection` (added in #16781) picked which one to name via `FxHashMap::into_iter().find(...)` — arbitrary hash-bucket order, not any order tsc's own output tracks.

**Structural rule** (oracle-verified against `typescript@7.0.2`, 10 probes): when multiple properties conflict, tsc always names the first one by a *combined declaration order* — walk the intersection's members left to right, and within each member walk its own properties in declaration order, positioning each name at its **first** occurrence across all members. This holds independent of which property is actually accessed, independent of alphabetical order, and uses the **first** member's property order even when a later member declares the same names in a different order (confirmed with reverse-alphabetical source order, swapped member orders, and a case where the conflicting name is introduced by a later member entirely).

`ObjectShape::properties` is sorted by interned `Atom` id for canonical hashing identity, so `Vec` position can't recover source order either. Each `PropertyInfo` already carries a `declaration_order` field for exactly this purpose (excluded from `PartialEq`/`Hash`, backfilled by the interner from source insertion order) — `crates/tsz-solver/src/type_queries/data/intersection_conflict.rs` now sorts each member's properties by it before folding into a first-seen combined order, and returns the first name in that order whose occurrences still conflict.

**Adjacent matrix** (new tests in `ts18031_intersection_conflicting_property_tests.rs`, all oracle-verified): first member's declaration order wins even when the access targets a different (also-conflicting) property; source order beats alphabetical order (reverse-alphabetical case); the first member's order wins even when a later member declares the same two names in the opposite order; a property absent from the first member is ordered by whichever later member introduces it first, using *that* member's own order for ties among newly-introduced names.

Related: my own #16788 opened moments after #16781 merged, duplicating its scope; closed as superseded once the overlap was clear (see comment there) — this PR is the one real gap #16788's review surfaced.

## Goal
Goal: hold

## Verification
- `cargo test -p tsz-checker --lib ts18031_intersection_conflicting_property_tests` — 12/12 (4 new multi-conflict-ordering cases + the 8 existing).
- `cargo test -p tsz-solver --lib intersection` — 506/506 (unaffected: the fix only changes *which* Atom is returned on the already-narrow multi-conflict path, not the has-a-conflict predicate).
- `cargo test -p tsz-checker --lib property` — 1410/1410, 1 pre-existing ignore.
- `cargo clippy -p tsz-solver -p tsz-checker --lib --all-features -- -D warnings` — clean.
- `cargo fmt --check -p tsz-solver -p tsz-checker` — clean.
- `python3 scripts/arch/arch_guard.py` — passed.
- Manual: 10 oracle probes against the pinned `typescript@7.0.2` binary (`node .../tsc --noEmit --strict ...`) establishing the combined-declaration-order rule before writing the fix.
- **Owed**: `compare-to-parent.sh` / conformance snapshot not run (no TypeScript corpus checked out in this container). Diagnostic *code* and *count* are unaffected — this only changes the property name inside an already-emitted `related_information` string on an already-narrow, newly-added path (#16781, not yet in a conformance baseline), so a 0/0 set-difference is expected for the same reason #16781 itself recorded.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-sonnet-5
Effort: medium
AgentName: ClaudeDE

---
_Generated by [Claude Code](https://claude.ai/code/session_01RqhVf4tK8AZi2NqrxHAU2c)_

### DIFF
--- a/crates/tsz-checker/src/tests/ts18031_intersection_conflicting_property_tests.rs
+++ b/crates/tsz-checker/src/tests/ts18031_intersection_conflicting_property_tests.rs
@@ -208,6 +208,107 @@ const read: "a" = value.kind;
     );
 }
 
+// ---------------------------------------------------------------------------
+// Multi-conflict ordering: when more than one property name conflicts, tsc
+// always names the first one by a combined declaration order (walk members
+// left to right, and within each member walk its own properties in
+// declaration order — a name is positioned at its *first* occurrence).
+// Oracle-verified against `typescript@7.0.2` for every case below.
+// ---------------------------------------------------------------------------
+
+#[test]
+fn multi_conflict_names_first_declared_property_in_first_member() {
+    // `x` is declared before `y` in the first member, so `x` wins even
+    // though the access is on `x` itself and both `x` and `y` conflict.
+    let diags = check_source_strict(
+        r#"
+interface A { x: 1; y: "a" }
+interface B { x: 2; y: "b" }
+declare const c: A & B;
+c.x;
+"#,
+    );
+    let diag = only(&diags, TS2339);
+    assert_eq!(
+        related(&diag, TS18031).as_deref(),
+        Some(
+            "The intersection 'A & B' was reduced to 'never' because property 'x' has conflicting types in some constituents."
+        ),
+        "got {diags:?}"
+    );
+}
+
+#[test]
+fn multi_conflict_order_follows_first_members_declaration_order_not_alphabetical() {
+    // Source order is `zz` then `aa` (reverse of alphabetical); the winner
+    // is `zz`, proving the pick is declaration order, not name order.
+    let diags = check_source_strict(
+        r#"
+interface P { zz: 1; aa: "a" }
+interface Q { zz: 2; aa: "b" }
+declare const r: P & Q;
+r.zz;
+"#,
+    );
+    let diag = only(&diags, TS2339);
+    assert_eq!(
+        related(&diag, TS18031).as_deref(),
+        Some(
+            "The intersection 'P & Q' was reduced to 'never' because property 'zz' has conflicting types in some constituents."
+        ),
+        "got {diags:?}"
+    );
+}
+
+#[test]
+fn multi_conflict_order_ignores_second_members_declaration_order() {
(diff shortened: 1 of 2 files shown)


mohsen1:
Verified and merging — **1 fixed, 0 broken**, and it takes the right approach where #16788 took the wrong one.

`--strict --target es2022 --pretty true`, the `TS18031` elaboration:

| fixture | source order | tsc | main | this PR | |
|---|---|---|---|---|---|
| `interface A { x: 1; y: 'a' }` / `B { x: 2; y: 'b' }` | `x, y` | `'x'` | **`'y'`** | `'x'` | FIXED |
| `interface P { zz: 1; aa: 'a' }` / `Q { zz: 2; aa: 'b' }` | `zz, aa` | `'zz'` | `'zz'` | `'zz'` | unchanged |
| single conflicting property | — | `'a'` | `'a'` | `'a'` | unchanged |

The second row is the one that proves the rule rather than a coincidence: source order is `zz, aa` while alphabetical order is `aa, zz`. Main happened to agree there — it was picking **alphabetically-last**, which collides with first-written only when the source is reverse-alphabetical. Row 1 separates them, and this PR gets both.

**This is the better resolution of the case #16788 was blocked on.** That PR proposed *omitting* the elaboration when multiple properties conflict, on the grounds that tsz "cannot recover tsc's first-written pick from the interned (Atom-sorted) shape". But tsc never omits — it always names the first-declared property — so silence would have been a second divergence, and it removes the only hint the user gets about which property broke the intersection. Recovering declaration order at the reduction seam, as this does, is the fix that actually converges on tsc.

```
cargo nextest run -p tsz-checker --lib   10641/10641   (+12)
cargo nextest run -p tsz-solver  --lib    7654/7654
conformance set-difference               0 fixed / 0 broken
```

Worth checking whether #16788 can now close or reduce — six of its seven rows already landed via #16781, and this covers the seventh.

Issue #16797, opened 2026-08-07
false-positive(solver): an intersection of two classes with same-named ES #private members wrongly reduces to never


mohsen1:
## Summary

tsz reduces an intersection of two classes to `never` when both declare an ES `#private` member of the same name. tsc does not — ES private names are lexically scoped to their declaring class, so `#x` in `A` and `#x` in `B` are **different names** with nothing to conflict. The wrong reduction makes every member access on the intersection an error.

## Oracle

`typescript@7.0.2`, `--noEmit --strict --pretty false --target es2022 --singleThreaded --stableTypeOrdering true`.

```ts
class A { #x = 1; m() { return 1; } }
class B { #x = 1; n() { return 2; } }
declare const v: A & B;
v.m();
v.n();
```

| | result |
| --- | --- |
| tsc | **0 errors** |
| tsz | **2 errors** — `Property 'm' does not exist on type 'never'`, same for `'n'` |

The type itself shows the divergence directly:

```ts
declare const v: A & B;
v.foo;

tsc   error TS2339: Property 'foo' does not exist on type 'A & B'.
tsz   error TS2339: Property 'foo' does not exist on type 'never'.
```

## The discriminating control

Modifier-`private` **does** reduce, and tsz is correct there:

```ts
class C { private p = 1; }
class D { private p = 1; }
declare const x: C & D;
x.foo;

tsc   Property 'foo' does not exist on type 'never'.
        The intersection 'C & D' was reduced to 'never' because property 'p'
        exists in multiple constituents and is private in some.
tsz   same
```

So the reduction machinery is right in general; it is the ES `#private` case that must be excluded from it.

## Structural rule

Two `private` members with the same name in unrelated classes are the *same* property name and make the intersection uninhabitable, so it reduces to `never`. Two ES `#private` members with the same spelling in unrelated classes are **distinct** names — scoped to their declaring class — and impose no constraint on each other, so the intersection stands.

## Adjacent cases a fix should carry

- one class using `#x` and the other using `private x` — mixed forms
- the same `#x` name in a class and a *subclass* (legal, and must not reduce)
- three-way intersections mixing both private kinds
- modifier-`private` must keep reducing, with its elaboration intact
- renamed binders throughout

## Note

[...]

PR #16799, opened 2026-08-07
fix(solver): TS18032 elaboration excludes ES #-private names

mohsen1:
Follow-up to #16796 (merged before its own review comment landed), addressing that comment directly: `find_private_brand_conflict_property` matched on the interned `Atom`'s text alone, which wrongly treats two classes' same-spelled ES `#`-private fields as one conflicting name.

## Structural rule

ES `#x` fields are lexically scoped to their declaring class — `#x` on class `A` and `#x` on class `B` are never the same name to `tsc`, even though they share identical source text. Real `tsc` (`typescript@7.0.2`) neither reduces `A & B` to `never` nor elaborates for that shape; it stays a plain `Property does not exist on type 'A & B'.`. Modifier-`private` members genuinely share one name, so they do collide (tsc's own TS18032 wording, "exists in multiple constituents and is private in some", only applies there).

tsz's `intersection_has_conflicting_private_brands` (`crates/tsz-solver/src/intern/normalize.rs`) already reduces the ES `#`-private shape to `never` on `main` today — a separate, pre-existing false-positive in the reduction itself, not owned by this elaboration (filed independently by the reviewer as #16797). This fix only stops attaching an authoritative-sounding TS18032 line to whatever `TS2339` that reduction bug produces; it does not touch the reduction. `is_es_private_identifier_name` (`tsz-solver/src/utils/mod.rs`, already used by `private_checker.rs` for the same ES-private-vs-modifier-private split) is reused here rather than re-deriving the `#`-prefix check.

**This PR does not fix the `#private` intersection reduction bug — see #16797 for that.** It only stops a wrong explanation from being attached to whatever diagnostic #16797 currently produces.

## Adjacent matrix

Oracle-verified against `typescript@7.0.2`, all in the existing test file:
- ES `#`-private same-spelled fields: no TS18032 elaboration (new negative test, the review comment's exact repro).
- Renamed ES-private binder (`#q`): no elaboration.
- Modifier-`private` (both existing positive cases): unchanged, still correct.

## Goal
Goal: hold

## Verification
- `cargo test -p tsz-checker --lib -- ts18032 ts18031 ts2339 intersection`: 367/367.
- `cargo clippy -p tsz-solver -p tsz-checker --all-targets`: clean.
- `cargo fmt --check -p tsz-solver -p tsz-checker`: clean.
- End-to-end via debug CLI against the pinned `typescript@7.0.2` oracle: the review comment's exact repro (`class A { #x = 1 }` / `B`, `(A & B).foo`) and a renamed-binder variant (`#q`) both now report the bare `TS2339` with no elaboration, matching tsc. Modifier-private control (`class C { private p = 1 }` / `D`) still carries the TS18032 line, byte-identical to tsc.
- Independently re-verified by reviewer: `-p tsz-checker --lib` 10650/10650, `-p tsz-solver --lib` 7654/7654, conformance set-difference 0 fixed / 0 broken.

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-sonnet-5
Effort: medium
AgentName: ClaudeT

---
_Generated by [Claude Code](https://claude.ai/code/session_01BTLt42yTYjb6Wwy8bamEKG)_

### DIFF
--- a/crates/tsz-checker/src/tests/ts18032_intersection_private_brand_conflict_tests.rs
+++ b/crates/tsz-checker/src/tests/ts18032_intersection_private_brand_conflict_tests.rs
@@ -30,12 +30,22 @@
 //! Deliberately narrow scope, matching TS18031's own precedent: only a
 //! directly-written intersection annotation (no alias/generic-application/
 //! heritage indirection). Cases that don't reduce to `never` at all
-//! (protected-only conflicts report `TS2445` instead; ES `#`-private
-//! members don't merge into a brand conflict at all; the same class
+//! (protected-only conflicts report `TS2445` instead; the same class
 //! intersected with itself is not a conflict) are documented as negative
 //! controls below — they never reach this helper's `type_id == NEVER` gate,
 //! so they're regression coverage for the surrounding machinery, not for the
 //! new query itself.
+//!
+//! ES `#`-private members are a separate case: `#x` on two different classes
+//! is never the same name to `tsc` (each is lexically scoped to its own
+//! class body), so real `tsc` neither reduces the intersection to `never`
+//! nor elaborates. tsz's `intersection_has_conflicting_private_brands`
+//! (`crates/tsz-solver/src/intern/normalize.rs`) currently *does* still
+//! reduce that shape to `never` — a separate, pre-existing bug in the
+//! reduction itself, not owned by this elaboration. `find_private_brand_conflict_property`
+//! is deliberately scoped to never attach the TS18032 line to whatever
+//! `TS2339` that reduction bug produces, so fixing the reduction later needs
+//! no matching change here.
 
 use crate::diagnostics::Diagnostic;
 use crate::test_utils::check_source_strict;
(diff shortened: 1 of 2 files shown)


mohsen1:
Verified and merging — **2 fixed, 0 broken.** This is exactly the gating I asked for on #16796.

`--strict --target es2022 --pretty true`:

|---|---|---|---|---|
| `class A { #x = 1 }` / `B`, `(A & B).foo` | **(silent)** | elaborates | **(silent)** | FIXED |
| renamed binders (`#q`) | (silent) | elaborates | (silent) | FIXED |
| `private p` — modifier-private | elaborates | elaborates | elaborates | correctly kept |
| plain conflicting property (#16793's shape) | "has conflicting types…" | same | same | unchanged |

Both controls hold, so this suppresses only the ES `#-private` path and leaves #16796's genuine modifier-`private` fix and #16793's multi-conflict work intact.

`-p tsz-checker --lib` **10650/10650**, `-p tsz-solver --lib` 7654/7654, conformance set-difference **0 fixed / 0 broken**.

**#16797 stays open** — this fixes the *explanation*, not the reduction:

```
declare const v: A & B;   // both classes have #x, plus distinct m() and n()
v.m(); v.n();

tsc   0 errors
main  2 errors      <- unchanged by this PR
main  Property 'foo' does not exist on type 'never'   (tsc says 'A & B')
```

That is the right split and I am not asking for more here — silencing a wrong explanation is strictly better than shipping a confident one, and it restores the pre-#16796 state on that path while keeping the half that was correct. But the intersection still collapses to `never` when it should not, so anyone reading this PR as "the `#private` intersection bug is fixed" would be mistaken. Worth a line in the body pointing at #16797 so that is unambiguous.

PR #16804, opened 2026-08-07
fix(parser): TS1070/TS1071 for `export` and multi-modifier runs on type members

mohsen1:
When a class-member modifier is written on a **type member** (an interface member or a type-literal member), `tsc`'s `checkGrammarModifiers` reports a single diagnostic anchored on — and naming — the **first** modifier: **TS1070** (`'{0}' modifier cannot appear on a type member.`) for a property/method member, **TS1071** (`... on an index signature.`) for an index signature. `readonly` is the one modifier `tsc` accepts on a type member and is preserved. tsz owns this in the parser (`parse_type_member_visibility_modifier_error`).

`main` already covered `export`/`in`/`out` as **single** illegal modifiers (TS1070). This PR closes the remaining gap in the same family:

- **Multi-modifier runs.** A run of more than one illegal modifier (`public static x`, `static public x`, `export static x`, `public static readonly x`, `declare export x`) previously recovered only past the first modifier, so the remainder mis-parsed into a `TS1005` cascade instead of `tsc`'s single report on the first modifier. The error path now consumes the **entire** leading run of illegal modifiers before parsing the underlying member/index signature, leaving a trailing legal `readonly` in place — a single TS1070/TS1071 anchored on and naming the first modifier, matching `tsc`.
- **Index-signature position.** An illegal modifier (including a run) leading an index signature now reports a single **TS1071** at the first modifier.

The type-member modifier grammar helpers move to a new `state_declarations_type_member_modifiers` module so `state_declarations.rs` stays under the 2000-line cap; `is_illegal_type_member_modifier` is the single keyword set shared by the classification look-ahead and the consume loop (so `export`/`in`/`out`/access/`static`/`abstract`/`declare`/`accessor`/`override` are handled uniformly).

Found via the [#16291](https://github.com/tsz-org/tsz/issues/16291) "check a wired code's other positions" method.

## Goal
Goal: hold

## Verification
- Merged latest `origin/main` (resolved the overlap with main's own `export`/`in`/`out` addition; `in`/`out` folded into the shared keyword set, so no coverage is lost). Multi-modifier tests consolidated into `crates/tsz-parser/tests/type_member_modifier_grammar_tests.rs` (main's canonical file for this family).
- Every new row oracle-matched against `tsc@7.0.2` (`--noEmit --strict --target es2022 --pretty false`) for code, anchor column, and message: `export` on an index signature → TS1071; two-modifier runs (`public static`, `static public`) naming the first; `export`-led and `declare export` runs; three-modifier run keeping the trailing `readonly` member; a modifier run before an index signature → single TS1071; a member following an illegal-modifier member is not swallowed; renamed-binder variants (anti-hardcoding).
- `cargo test -p tsz-parser` — 2019 passed, 0 failed.
- `cargo clippy -p tsz-parser --all-targets` — clean; `cargo fmt -p tsz-parser --check` — clean; `arch_guard` — passed (`state_declarations.rs` at 1744 lines after the split).
- The pre-commit hook re-ran clippy + arch-guard across all workspace crates after the merge — passed.
- Parser-only change; no emit/DTS surface touched. (The `index_signature` relation-architecture unit failures in the checker suite are pre-existing on clean `main` — verified with this change stashed: identical 263 passed / 3 failed — and unrelated.)

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-opus-4-8
Effort: high


mohsen1:
**Needs a rebase.** Conflicts with main — #16795 and #16803 both landed in the type-member modifier grammar in the last couple of hours, so this branch is behind on the same code.

I did not measure the branch itself, because my merge aborted and any binary I built would be main's rather than yours.

**What I can offer is the current main-vs-tsc state on your shapes**, which should still be accurate targets after the rebase (`--strict --target es2022`):

| shape | tsc | main today | |
|---|---|---|---|
| `interface I { export m(): void }` | `TS1070` | `TS1070` | **already fixed** by #16795 |
| `interface J { public private p: number }` | `TS1070` | `TS1005` | still open |
| `interface K { readonly readonly q: number }` | **`TS1030`** | `TS1005` | still open |
| `interface L { readonly r: number }` — valid | clean | clean | ok |

So the `export` half of your title is already done, and what remains is the multi-modifier run. Note the duplicate-modifier case wants **`TS1030`** ("'readonly' modifier already seen"), not `TS1070` — worth checking your implementation distinguishes "modifier cannot appear here" from "modifier already seen", since they are different codes for what looks like one grammar path.

Once rebased I will run the usual: the matrix above, `-p tsz-parser`, and a conformance set-difference. That last one matters here — emitting a previously-missing `TS1xxx` can regress net conformance through `has_parse_errors` suppression, which is why #16795 and #16803 both got a full snapshot.

PR #16807, opened 2026-08-07
fix(parser): non-readonly modifiers before an interface/type-literal accessor report TS1131, not TS1070

mohsen1:
When a run of one or more `private`/`protected`/`public`/`static`/`accessor`/`export`/`readonly` type-member modifiers is directly followed, on the same line, by a `get`/`set` accessor signature in an interface or type-literal body, no modifier at all may precede the accessor in tsc's grammar. tsc's parser cannot build any member starting at the first modifier: it reports one TS1131 (`Property or signature expected.`) per modifier in the run, each anchored at its own token, then recovers by retrying at the accessor keyword, which parses as a clean, modifier-less accessor.

tsz previously reported the uniform semantic TS1070 (`'{0}' modifier cannot appear on a type member.`) for every modifier in this position except `readonly` (fixed for the single-modifier case by #16803), and would have collapsed a stacked modifier run to a single diagnostic instead of one TS1131 per modifier.

Closes the first of three follow-ups named on #16803's own scope-out, itself part of #16291.

## Structural rule

When an interface/type-literal member begins with a run of clean type-member modifiers immediately followed by an accessor keyword on the same line, tsc fails the whole member at the first modifier (TS1131 × N) through its member-dispatch grammar; tsz now does the same through a generalized lookahead (`look_ahead_modifier_run_before_accessor`, `state_types_advanced.rs`) that subsumes and replaces the `readonly`-only lookahead #16803 introduced, shared by both owner loops (`parse_type_members` for interfaces, `parse_type_literal_rest` for type literals, both in the parser).

`declare`/`abstract`/`override`/`in`/`out` are deliberately excluded from the qualifying modifier set: oracle-verified against `typescript@7.0.2`, those cascade into a different diagnostic shape (TS1434/TS1005/TS1128, as if the retry re-parses the accessor's own name as a fresh top-level statement) rather than one TS1131 per modifier. They keep reporting the pre-existing TS1070 unchanged — a distinct, harder problem, out of scope here and named as a follow-up in the module doc comment.

| source | tsc | tsz before | tsz after |
| --- | --- | --- | --- |
| `interface I { static get x(): number }` | TS1131 | TS1070 | TS1131 |
| `interface I { public static get x(): number }` | TS1131, TS1131 (one per modifier) | TS1070 (one, on `public`) | TS1131, TS1131 |
| `type T = { static get x(): number };` | TS1131 | TS1070 | TS1131 |
| `interface I { declare get x(): number }` | TS1131 + TS1434/TS1005/TS1128 cascade | TS1070 | TS1070 (unchanged, out of scope) |
| `interface I<T> { in get x(): number }` | TS1131 + cascade | TS1070 | TS1070 (unchanged, out of scope) |
| `interface I { static get(): number }` (method named `get`) | TS1070 | TS1070 | TS1070 (unchanged) |
| `interface I { static get: number }` (property named `get`) | TS1070 | TS1070 | TS1070 (unchanged) |

Rebased onto main after #16804 (a sibling type-member-modifier PR) landed and moved the shared `parse_type_member_visibility_modifier_error` into its own module. That function is a different call site from the one this change touches (both loops' pre-member-dispatch accessor check, checked earlier in each loop), so the two changes are independent and merged without semantic conflict — confirmed by re-running the full local verification post-rebase. One existing adjacent-case test's expected diagnostic set (`static declare get x(): number`) was updated to match #16804's now-collapsed single-TS1070-per-run behavior for the pre-existing, still out-of-scope cascade family.

## Goal

Goal: hold

## Verification

- New adjacent-case tests in `crates/tsz-parser/tests/type_member_modifier_grammar_tests.rs` (12 cases), all oracle-verified against `typescript@7.0.2`: each clean modifier alone before `get`/`set`, the type-literal variant, stacked 2- and 3-modifier runs (each modifier gets its own TS1131), `readonly` mixed into a multi-modifier run, cascade modifiers (`declare`/`abstract`/`override`/`in`/`out`) confirmed unchanged at TS1070, a cascade modifier poisoning a run that starts with a clean modifier, line-break non-interference, and the get/set-as-property-name negative control.
- `cargo test -p tsz-parser --lib`: 2041 passed, 0 failed (47 in the extended `type_member_modifier_grammar_tests` module, 12 new).
- `cargo test -p tsz-checker --lib -- interface_member type_member modifier_grammar`: 72 passed, 0 failed.
- `cargo clippy -p tsz-parser --all-targets`: clean.
- `cargo fmt -p tsz-parser --check`: clean (no diff).
- Parser-only grammar change on a rare syntax shape (a modifier directly before an interface/type-literal accessor); no emit/DTS surface touched, conformance/fourslash movement not expected.

## Provenance

Machine: cloud
Assistant: claude-code
Model: claude-sonnet-5
Effort: medium
AgentName: Mica

### DIFF
--- a/crates/tsz-parser/src/parser/state_declarations.rs
+++ b/crates/tsz-parser/src/parser/state_declarations.rs
@@ -310,27 +310,32 @@ impl ParserState {
                 continue;
             }
 
-            // `readonly` directly followed by a `get`/`set` accessor is grammar-invalid
-            // in tsc: no modifier may precede an accessor signature in an interface or
-            // type literal (unlike a plain property/method, which accepts `readonly` and
-            // is validated semantically as TS1024). tsc's parser cannot build any member
-            // starting at `readonly` here, reports TS1131 anchored at the modifier, and
-            // retries from the accessor keyword — which then parses as a bare
-            // (modifier-less) accessor. Handled before the normal property/method path so
-            // `get`/`set` is never mistaken for the property name and `readonly` is never
-            // silently dropped (which previously produced a misleading TS1005).
-            if self.is_token(SyntaxKind::ReadonlyKeyword)
-                && self.look_ahead_is_accessor_after_readonly()
-            {
-                let ro_start = self.token_pos();
-                let ro_end = self.token_end();
-                self.next_token(); // consume `readonly`
-                self.parse_error_at(
-                    ro_start,
-                    ro_end.saturating_sub(ro_start),
-                    tsz_common::diagnostics::diagnostic_messages::PROPERTY_OR_SIGNATURE_EXPECTED,
-                    tsz_common::diagnostics::diagnostic_codes::PROPERTY_OR_SIGNATURE_EXPECTED,
-                );
+            // A run of one or more type-member modifiers directly followed by a
+            // `get`/`set` accessor is grammar-invalid in tsc: no modifier may precede
+            // an accessor signature in an interface or type literal (unlike a plain
+            // property/method, which accepts these modifiers and is validated
+            // semantically as TS1024/TS1070/TS1071). tsc's parser cannot build any
+            // member starting at the first modifier, reports one TS1131 per modifier
+            // in the run (each anchored at its own token), and recovers by retrying
+            // at the accessor keyword — which then parses as a bare (modifier-less)
+            // accessor. See `look_ahead_modifier_run_before_accessor` for the exact
+            // qualifying modifier set. Handled before the normal property/method path
+            // so `get`/`set` is never mistaken for the property name and the
+            // modifiers are never silently dropped (which previously produced a
+            // misleading TS1005 or TS1070).
+            let modifier_run_len = self.look_ahead_modifier_run_before_accessor();
+            if modifier_run_len > 0 {
+                for _ in 0..modifier_run_len {
+                    let mod_start = self.token_pos();
+                    let mod_end = self.token_end();
+                    self.next_token();
+                    self.parse_error_at(
+                        mod_start,
+                        mod_end.saturating_sub(mod_start),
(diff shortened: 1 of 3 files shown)


mohsen1:
**Needs a rebase** — still conflicts with main at `20b8e48db5`. I tried both your original head and this one; both abort, so I have not measured the branch. Any binary I built is main's, and reporting its numbers as yours would be worthless.

The area has moved three times in the last few hours — #16795, #16803, and #16804 all landed in the type-member modifier grammar — so a rebase is worth doing promptly before it drifts further.

**Your target is real and still open.** Current main vs tsc, `--strict --target es2022`:

| shape | tsc | main today | |
|---|---|---|---|
| `interface I { public get p(): number }` | **`TS1131`** | `TS1070` | your target |
| `interface J { static get q(): number }` | **`TS1131`** | `TS1070` | your target |
| `type T = { public get r(): number }` | **`TS1131`** | `TS1070` | your target |
| `interface K { readonly get s(): number }` | `TS1131` | `TS1131` | already correct — #16803 |
| `interface L { export m(): void }` | `TS1070` | `TS1070` | already correct — #16795 |
| `interface M { get t(): number }` — valid | clean | clean | ok |

The last three are the controls I would keep in your test file. Rows 4 and 5 are the interesting ones: `readonly` before an accessor is already `TS1131`, and `export` on a plain member is already `TS1070`, so the rule is not "accessors are always `TS1131`" nor "modifiers on type members are always `TS1070`" — it is specifically *non-`readonly` modifiers before an accessor* that need to move from `TS1070` to `TS1131`. Landing this without those two controls would make it easy to over-apply in either direction.

Once rebased I will run that matrix, `-p tsz-parser`, and a conformance set-difference — the last because emitting or changing a `TS1xxx` can regress net conformance through `has_parse_errors` suppression, which is why every parser PR in this series got a full snapshot.

PR #16808, opened 2026-08-07
fix(parser): stop over-reporting TS1029 for the async/readonly/accessor pair

mohsen1:
`async`/`readonly`/`accessor` class-member modifier ordering: tsc's `checkGrammarModifiers` never reaches the "X must precede async" ordering diagnostic (TS1029) for `readonly`/`async` or `accessor`/`async` — `async` is legal only on methods/accessors-as-functions, while `readonly`/`accessor` are legal only on data-member declarations, so the two families never share a legal member kind and the ordering check is structurally unreachable in real tsc (confirmed against `typescript@7.0.2` across property + method witnesses, both directions: `readonly async p: number` → tsc reports only TS1042; `async readonly m()` → tsc reports only TS1024; same for `accessor`). tsz emitted a spurious extra TS1029 alongside the correct code anyway.

Same defect, different trigger, in the `abstract` branch: `abstract`/`async` never coexist either (tsc rejects the pair outright via TS1243, not an ordering check), and `abstract`/`readonly` legally coexist in *either* order (no ordering constraint at all — `readonly abstract p: number;` is clean in tsc) — but the combined trigger condition treated both as ordering violations. `abstract`/`override` and `abstract`/`accessor` remain genuinely order-constrained (verified both directions) and are untouched.

## Structural rule
When two class-member modifiers can never both be syntactically legal on the same member kind (e.g. a function-only modifier and a data-member-only modifier), tsc's modifier-ordering grammar check (TS1029) never fires for that pair — the member's own kind-specific rejection (TS1024/TS1042/TS1243) already covers it. tsz does this through the parser's per-modifier-keyword ordering walk in `crates/tsz-parser/src/parser/state_statements_class_members.rs`.

## Adjacent matrix (oracle-verified, `typescript@7.0.2`)

| source | tsc | tsz before | tsz after |
| --- | --- | --- | --- |
| `readonly async p: number = 1;` | TS1042 only | TS1042 + TS1029 | TS1042 only |
| `async readonly m(): Promise<number> {}` | TS1024 only | TS1024 + TS1029 | TS1024 only |
| `async accessor p: number = 1;` | TS1042 only | TS1042 + TS1029 | TS1042 only |
| `readonly abstract p: number;` | clean (readonly) | TS1029 (false) | clean |
| `abstract readonly p: number;` | clean (readonly) | clean | clean (unchanged) |
| `async abstract m(): Promise<number>;` | TS1243 only | TS1243 + TS1029 | TS1243 only |
| `abstract override m(): void;` (correct order) | clean | clean | clean (unchanged, regression guard) |
| `override abstract m(): void;` (wrong order) | TS1029 | TS1029 | TS1029 (unchanged, regression guard) |
| `abstract accessor p: number;` (correct order) | clean | clean | clean (unchanged, regression guard) |
| `accessor abstract p: number;` (wrong order) | TS1029 | TS1029 | TS1029 (unchanged, regression guard) |

## Goal
Goal: hold

## Verification
- 10 new oracle-verified tests in `crates/tsz-parser/tests/modifier_ordering_tests.rs` covering both directions of readonly/async, accessor/async, abstract/readonly, plus regression guards that abstract/override and abstract/accessor still fire correctly.
- `cargo test -p tsz-parser --lib`: 2020 passed, 0 failed (was 2010).
- `cargo test -p tsz-checker --lib -- member_modifier_placement class_member_modifier_grammar`: 24 passed, 0 failed (no regressions in the sibling checker-level modifier suite).
- `cargo clippy -p tsz-parser --all-targets -- -D warnings`: clean.
- `cargo fmt -p tsz-parser -- --check`: clean.
- Parser-only diff (no `crates/**/src` outside `tsz-parser`), narrow grammar-only change on a rare misordered-modifier shape — conformance/emit/fourslash not run locally this session; no code-set or count movement expected (matches the 0/0 signature already measured for the sibling modifier-grammar PRs merged earlier today, e.g. #16789, #16795, #16803, #16804).

## Provenance
Machine: cloud
Assistant: claude-code
Model: claude-sonnet-5
Effort: medium

---
_Generated by [Claude Code](https://claude.ai/code/session_01XD2NL5LQHxEQYK5xViZcHD)_

### DIFF
--- a/crates/tsz-parser/src/parser/state_statements_class_members.rs
+++ b/crates/tsz-parser/src/parser/state_statements_class_members.rs
@@ -162,16 +162,18 @@ impl ParserState {
                         diagnostic_codes::MODIFIER_ALREADY_SEEN,
                     );
                 }
-                if seen_readonly || seen_override || seen_accessor || seen_async {
+                // `readonly` and `async` are excluded from this ordering
+                // check: `abstract` and `readonly` legally coexist in either
+                // order (a `readonly abstract` property is clean in tsc), and
+                // `abstract`/`async` never coexist at all (tsc rejects the
+                // combination outright with TS1243 regardless of order, not
+                // this ordering diagnostic).
+                if seen_override || seen_accessor {
                     use tsz_common::diagnostics::diagnostic_codes;
                     let other = if seen_override {
                         "override"
-                    } else if seen_readonly {
-                        "readonly"
-                    } else if seen_accessor {
-                        "accessor"
                     } else {
-                        "async"
+                        "accessor"
                     };
                     self.parse_error_at_current_token(
                         &format!("'abstract' modifier must precede '{other}' modifier."),
@@ -198,13 +200,11 @@ impl ParserState {
                         "'readonly' modifier cannot be used with 'accessor' modifier.",
                         diagnostic_codes::MODIFIER_CANNOT_BE_USED_WITH_MODIFIER,
                     );
-                } else if seen_async {
-                    use tsz_common::diagnostics::diagnostic_codes;
-                    self.parse_error_at_current_token(
-                        "'readonly' modifier must precede 'async' modifier.",
-                        diagnostic_codes::MODIFIER_MUST_PRECEDE_MODIFIER,
-                    );
                 }
+                // No `seen_async` ordering check here: `readonly` (data-member
+                // only) and `async` (method-only) never share a legal member
+                // kind, so tsc never reaches this ordering diagnostic for the
+                // pair — the member's own TS1024/TS1042 already covers it.
                 seen_readonly = true;
             } else if current_kind == SyntaxKind::OverrideKeyword {
                 // Check for duplicate override modifier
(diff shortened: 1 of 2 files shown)


mohsen1:
Post-merge verification — **1 fixed, 0 broken.**

`--strict --target es2022`:

|---|---|---|---|---|
| `class C { async readonly m() {} }` | `TS1024` | `TS1024` + **`TS1029`** | `TS1024` | FIXED |
| `class D { readonly async n() {} }` | `TS1024` | `TS1024` | `TS1024` | correctly unchanged |
| `class E { accessor readonly p = 1 }` | `TS1243` | `TS1243` | `TS1243` | correctly unchanged |
| `class F { async m() {} }` — valid | clean | clean | clean | no FP |

Main was stacking a spurious `TS1029` ("modifier must precede modifier") on top of the correct `TS1024`, but only in the `async readonly` order — `readonly async` was already right. That asymmetry is the useful detail: the over-report was order-dependent, so a probe that only tried one ordering would have scored it clean.

The `accessor readonly` row is a good control to have kept — it is a neighbouring pair in the same modifier-ordering machinery with a different correct code (`TS1243`), and it stays put.

`cargo nextest run -p tsz-parser` **2051 passed**, conformance set-difference **0 fixed / 0 broken** (verified alongside #16807).

